PR #1, opened 2020-06-08
Changes to move into the  ansible organization

matburt:
Some of these changes are transitional, such as the removal of travis
and the changes to some maintainer info. We'll bring these back in
once we decide where they'll land.


README.md
```diff
@@ -186,4 +184,4 @@ If everything works, commit the updated version, then tag a new repository relea
 
 ## Author
 
-This operator was built in 2019 by [Jeff Geerling](https://www.jeffgeerling.com), author of [Ansible for DevOps](https://www.ansiblefordevops.com) and [Ansible for Kubernetes](https://www.ansibleforkubernetes.com).
+This operator was originally built in 2019 by [Jeff Geerling](https://www.jeffgeerling.com), author of [Ansible for DevOps](https://www.ansiblefordevops.com) and [Ansible for Kubernetes](https://www.ansibleforkubernetes.com) and is now maintained by the Ansible Team
```
geerlingguy, line 187: You can drop the book links here. For now, at least, I'm happy with just a link / recognition, and this can fall off at some point.

geerlingguy:
@ryanpetrello - Thanks for the note! Will both images still be maintained (though identical), or should we then update both images to `ansible/awx` once 12.0.0 is out?

ryanpetrello:
Hey @geerlingguy,

Nope, both images will be converged into one as of 12.0.0 (the task and web container will just use the same image and run different services).

PR #10, opened 2020-07-19
Use k8s-exec instead of shell

### LINKED ISSUE #3: Switch to using k8s_exec module instead of shell module + kubectl
Filing this issue to keep tabs on [this issue](https://github.com/geerlingguy/tower-operator/issues/8). The use of `shell` when a module exists makes me sad.

tylerauerbeck:
This updates the `tower` role to use the `k8s_exec` module instead of shelling out to use the `kubectl` command. Had to explicitly add the kubernetes collection to the role so that it would pick things up and know how to find the appropriate module, which I believe is a known issue.

resolves #3

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,8 +1,5 @@
 FROM quay.io/operator-framework/ansible-operator:v0.17.0
 
-# Install kubectl.
-COPY --from=lachlanevenson/k8s-kubectl:v1.18.3 /usr/local/bin/kubectl /usr/local/bin/kubectl
-
 # Install Ansible requirements.
 COPY requirements.yml ${HOME}/requirements.yml
 RUN ansible-galaxy collection install -r ${HOME}/requirements.yml \
--- a/roles/tower/meta/main.yml
+++ b/roles/tower/meta/main.yml
@@ -26,3 +26,6 @@ galaxy_info:
     - deployment
 
 dependencies: []
+
+collections:
+  - community.kubernetes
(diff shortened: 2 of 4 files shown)


tylerauerbeck:
I think this is just failing because of `ansible-lint`. I've tried a few things to get this fixed, but I'm not quite sure how to go about this. I also came across this comment from @geerlingguy (https://github.com/ansible/ansible-lint/issues/538#issuecomment-588348782) that mentions a similar issue.

Any input here would be appreciated. Or if this is a larger issue I guess diabling ansible-lint may be the solution for now?

tylerauerbeck:
@matburt @geerlingguy Any thoughts here? I see I need to rebase as the refactor PR went in, but wasn't sure about the current failing test.

geerlingguy:
It looks like it needs a reroll at this point.

Regarding the fail:

```
Couldn't parse task at /home/travis/build/ansible/awx-operator/roles/tower/tasks/initialize.yml:2 (couldn't resolve module/action 'k8s_exec'. This often indicates a misspelling, missing collection, or incorrect module path.
```

This is likely because Molecule/Ansible aren't picking up the collection correctly (or maybe it's not installed at all? I don't see it being added to any installation steps or anything in CI).

matburt:
Yeah this needs to be rebased against the head of devel before we can look at merging it.

shanemcd:
Closing in favor of https://github.com/ansible/awx-operator/pull/39

PR #17, opened 2020-07-28
Remove redis and memcached unused variables and their exposed ports

schen1:
We should remove the unused variables for Redis and memcached.

Also, their respective containers are exposing some ports that are not used because Redis and memcached are communicating via sockets.

### DIFF
--- a/roles/awx/templates/tower.yaml.j2
+++ b/roles/awx/templates/tower.yaml.j2
@@ -37,16 +37,12 @@ spec:
           - '/var/run/memcached/memcached.sock'
           - '-a'
           - '0666'
-        ports:
-        - containerPort: 1121
         volumeMounts:
           - name: {{ meta.name }}-memcached-socket
             mountPath: "/var/run/memcached"
       - image: '{{ tower_redis_image }}'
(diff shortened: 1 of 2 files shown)


matburt:
I agree, can you rebase this PR so we can clear the merge conflicts?

schen1:
I did a rebase from upstream following https://medium.com/@topspinj/how-to-git-rebase-into-a-forked-repo-c9f05e821c8a

I hope it's fine. Otherwise, please let me know how I should do it.

Thanks

matburt:
I'm sorry about this @schen1 but it looks like this yielded a merge commit which is specifically what I don't want to see.

Go back to where you were before this and just try:

`git fetch --all` to pull the upstream changes into a copy of your devel branch and then
`git rebase origin/devel` or whatever you call the ansible remote.

Alternatively, generate a patch... regen your branch from the HEAD of devel and then reapply it.

schen1:
Alright, thanks for the suggestions. I checked and the branches had diverged so it created a bit of a mess.

I regenerated the patch from devel so it should be fine.

PR #23, opened 2020-09-10
Update README.md

nomyownnet:
Changed example of secret for external database for helping copypasters.

Otherwise you'll get error "'dict object' has no attribute 'host'" during installation.


matburt (changes requested):
Can you also update some other locations where this is enumerated such as:
https://github.com/ansible/awx-operator/blob/devel/deploy/crds/awx_v1beta1_crd.yaml#L48

matburt (approved):

PR #39, opened 2020-10-11
Use k8s-exec instead of shell

shanemcd:


shanemcd:
Resolves https://github.com/ansible/awx-operator/issues/3

PR #41, opened 2020-10-12
Add AWX_ANSIBLE_COLLECTIONS_PATHS to configmap

### LINKED ISSUE #36: Ansible Collections path still not set ALL Inventory plugins broken
Version: 0.5.0
Issue: 
The `AWX_ANSIBLE_COLLECTIONS_PATHS = '/var/lib/awx/vendor/awx_ansible_collections'` is not being set still in the lastest version of the operator. Without this it is not possible to use the VMware inventory plugin. 

FIX:
Ensure that the latest version that is built picks up the changes to the https://github.com/ansible/awx-operator/blob/devel/roles/awx/templates/tower_config.yaml.j2 template

shanemcd:
Fixes https://github.com/ansible/awx-operator/issues/36

### DIFF
--- a/roles/awx/templates/tower_config.yaml.j2
+++ b/roles/awx/templates/tower_config.yaml.j2
@@ -27,6 +27,7 @@ data:
     ADMINS = ()
     STATIC_ROOT = '/var/lib/awx/public/static'
     PROJECTS_ROOT = '/var/lib/awx/projects'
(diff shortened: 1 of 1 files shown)


shanemcd:
Wait... this is already here: https://github.com/ansible/awx-operator/blob/c34d11be69f015b03e2aa35b66bac4711ad9eddc/roles/awx/templates/tower_config.yaml.j2#L40

PR #64, opened 2021-01-20
Provide clusterserviceversion

Spredzy:


shanemcd:
@Spredzy Can you take a look at the test failures?

PR #80, opened 2021-02-05
More flexible database

Spredzy:


roles/awx/defaults/main.yml
```diff
@@ -68,15 +68,42 @@ tower_web_extra_volume_mounts: ''
...
+# the operator will look for <awx-instance-name>-postgres-configuration secret
+tower_postgres_configuration_secret: ''
+
+# In an external_database=true scenario, those are the variables used to
+# connect to the database - Not recommended. A secret should be used.
+#
+# In an external_database=false scenario, those are the variables used to
+# configure the database
+tower_postgres_user: awx
+tower_postgres_pass: ''
```
shanemcd, line 106: Having these variables live in this file makes it seem like they can be provided directly by the user via a CR, which they can't / shouldn't be. They should only ever be looked up via a Secret.

Spredzy: Well, this is actually the whole point of this PR. It is to be more flexible to the way a user can provide those data.

1. Provide a secret (name it the way you want) specify it as `tower_postgres_configuration_secret`
2. Provide connection data (user, pass, database, ...) with one variable per value (not recommended) but it allows an easy onboarding (think UI at the OLM level)
3. Default to the Operator to read `<instance-name>-postgres-configuration`

1 and 3 are the recommended option. Number 2 is not, but for PoC perspective it allows a really easy way in.

README.md
```diff
@@ -107,41 +108,111 @@ If you are attempting to do this on an OpenShift cluster, you will need to grant
...
+### Database Setup
 
-When the Operator installs the AWX services and generates a Postgres deployment it will lay down a config file to enable AWX to connect to that service. To use an external database you just need to create a `Secret` that the AWX deployment will use instead and then set a property in the CR:
+The AWX Operator supports two different scenarios with regard to database setup.
 
-    ---
-    spec:
-      ...
-      external_database: true
+One can either let the Operator manages a PostgreSQL deployment within the same namespace the AWX deployment will take place. Or, one can connect to an already - externally managed - PostgreSQL instance. The decision is made based on the boolean value of the `external_database` ansible variable.
```
shanemcd, line 115:
```suggestion
One can either let the Operator manage a PostgreSQL deployment within the same namespace as AWX, or one can connect to an externally managed PostgreSQL instance. The decision is made based on the boolean value of the `external_database` Ansible variable.
````

File: README.md [hunk 3]
```diff
@@ -107,41 +108,111 @@ If you are attempting to do this on an OpenShift cluster, you will need to grant
...
+There are three ways one can specify those information. Those way are listed in order of priority in which they are treated.
 
-If you need to use a specific storage class for Postgres' storage, specify `tower_postgres_storage_class` in your AWX spec:
+* Providing a specifically crafted secret
 
-    ---
-    spec:
-      ...
-      tower_postgres_storage_class: fast-ssd
+One can create a specially formated secret within the namespace the AWX instance will be deployed into. This secret needs then to be specified as `tower_postgres_configuration_secret` variable in the spec requirements.
```
shanemcd, line 125:
```suggestion
One can create a specially formated `Secret` within the namespace the AWX instance will be deployed into. This `Secret` needs then to be specified as `tower_postgres_configuration_secret` variable under the `spec` field in your AWX Custom Resource.
```

README.md
```diff
@@ -107,41 +108,111 @@ If you are attempting to do this on an OpenShift cluster, you will need to grant
...
+| tower_postgres_port     | Port to connect to PostgreSQL     | 5432          |
 
-If it's not specified, Postgres will store it's data on a volume using the default storage class for your cluster.
+The following is an example of an AWX spec section that relie on those variables.
+
+```yaml
+spec:
+  external_database: true
+  tower_postgres_user: awx
+  tower_postgres_pass: awx
```
shanemcd, line 168: ⚠️ This is not secure. We should _never_ instruct the user to provide sensitive information in anything except a `Secret`.

roles/awx/defaults/main.yml
shanemcd: Well, sorry, this isn't going to work 😅

shanemcd: https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#should-i-use-a-configmap-or-a-custom-resource

>Note: Use a secret for sensitive data, which is similar to a configMap but more secure.

Spredzy:
Will revisit and reopen.

PR #81, opened 2021-02-05
tower_admin_user: default to admin

Spredzy:


shanemcd (approved):

roles/awx/defaults/main.yml
```diff
@@ -15,7 +15,7 @@ tower_ingress_tls_secret: ''
 
 tower_hostname: example-awx.test
 
-tower_admin_user: test
+tower_admin_user: admin
 tower_admin_email: test@example.com
 tower_admin_password: changeme
```
shanemcd, line 20: Either as a part of this PR or some followup, we should delete this as it will never be allowed to come from the CR.

Spredzy: Will remove in a follow-up PR where I will introduce the secret based pattern for it.

PR #82, opened 2021-02-05
deployment_type: Introduce the new variable

Spredzy:


shanemcd (approved):
WFM pending test results.

PR #83, opened 2021-02-05
PostgreSQL: Remove unused/wrongly used variable

Spredzy:


shanemcd (approved):
LGTM but per our discussion earlier, we will ultimately need to move this info to a Secret.

Spredzy:
Correct, this will be done in a subsequent PR

PR #133, opened 2021-03-19
Backup role for awx-operator

rooftopcellist:


roles/backup/README.md
```diff
@@ -0,0 +1,84 @@
...
+```yaml
+---
+apiVersion: awx.ansible.com/v1beta1
+kind: Backup
+metadata:
+  name: awx
+  namespace: my-namespace
+```
+
+Finally, use `kubectl` to create the awx instance in your cluster:
```
Zokormazo: To create the backup CR ? the awx instance is already there :)

deploy/crds/awx.ansible.com_v1beta1_backup_cr.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: awx.ansible.com/v1beta1
+kind: Backup
+metadata:
+  name: example-awx
+  namespace: example-awx
+spec:
+  tower_backup_pvc: ''
+  tower_backup_size: ''
+  tower_postgres_storage_class: ''
```
Zokormazo: `_postgres_` or `_backup_` ?

roles/backup/README.md
```diff
@@ -0,0 +1,84 @@
...
+```
+tower_backup_pvc: 'awx-backup-volume-claim'
+```
+
+> If no pvc or storage class is provided, the cluster's default storage class will be used to create the pvc.
+
+This role will automatically create a pvc using a Storage Class if provided:
+
+```
+tower_postgres_storage_class: 'standard'
```
Zokormazo: postgres again :)

deploy/crds/awx.ansible.com_v1beta1_backup_cr.yaml
rooftopcellist: Our installer role also has this variable, so I thought it might be better to call it `tower_postgres_storage_class` for consistency across the operator.  But I could also see the argument for consistency in the role.  

It is probably better to have these vars be distinct (and consistently named within the role), I'll change this to `tower_backup_storage_class`

roles/backup/README.md
rooftopcellist: The idea here is that the AWX CR object has already been made and AWX is already deployed when a backup is performed.  So to perform a backup, the user will create a new backup CRO (using the same awx.ansible.com/v1beta1 API), which will then kick off the backup role.  


------------------------
Oh, I see now, I mentioned the awx object rather than the backup object there.  thanks!

deploy/crds/awx.ansible.com_v1beta1_awxbackup_cr.yaml
```diff
@@ -0,0 +1,12 @@
+---
```
rooftopcellist: AFAIK, this file is unneeded.  If people think it is useful to keep as a starting point for users to create their own CR yaml file from, I am happy to keep it.  Otherwise, I'll give it the :axe:

roles/backup/tasks/secrets.yml
```diff
@@ -0,0 +1,124 @@
...
+    pod: "{{ meta.name }}-db-management"
+    command: >-
+      bash -c "echo '{{ secret_key_template }}' > {{ _backup_dir }}/secret_key_secret.yml"
+
+- name: Get admin_password
+  k8s_info:
+    kind: Secret
+    namespace: '{{ meta.namespace }}'
+    name: '{{ tower_admin_password_secret }}'
+  register: _admin_password
```
chambridge, line 19: I think this will only work in the case where the admin password was generated for the user as its using the default which assumes naming conventions: https://github.com/ansible/awx-operator/pull/133/files#diff-f9dcb95da85ffdf5274f2bf0f6f1dfea4073036a82af584a51b4e981af2a0e60R15

Instead you could gather the admin password from the status object of the AWX CR. This would ensure you are getting the value from either the supplied or generated admin password secret: https://github.com/ansible/awx-operator/blob/devel/ansible/templates/crd.yml.j2#L270

roles/backup/tasks/postgres.yml
```diff
@@ -0,0 +1,89 @@
...
+- name: Get the postgres pod information
+  k8s_info:
+    kind: Pod
+    namespace: '{{ meta.namespace }}'
+    label_selectors:
+      - "app={{ tower_name }}-{{ deployment_type }}-postgres"
+  register: postgres_pod
+  until: "postgres_pod['resources'][0]['status']['phase'] == 'Running'"
+  delay: 5
+  retries: 60
```
chambridge: This really only allows backup if the user makes use of the **managed** Postgres Statefulset. If the user supplies their own container-based database it would either need to be documented that they apply the above label or you could provide them an override option.

In the pulp-operator a label selector can be provided in the CRD. 
https://github.com/pulp/pulp-operator/blob/main/deploy/crds/pulpproject_v1beta1_pulp_crd.yaml#L100-L102
https://github.com/pulp/pulp-operator/blob/main/roles/postgres/tasks/migrate_data.yml#L11-L25

Just a suggestion, it may not be what you are looking to support.

roles/backup/tasks/secrets.yml
chambridge: I take it back I guess you allow the override via CR input here: https://github.com/ansible/awx-operator/pull/133/files#diff-ea6ce699660ccb70f61f22b9377eeec75ac9b63126ec970e7c9e90d441be415bR44-R46

My only question is wouldn't it be cleaner if you had these secrets (https://github.com/ansible/awx-operator/pull/133/files#diff-ea6ce699660ccb70f61f22b9377eeec75ac9b63126ec970e7c9e90d441be415bR41-R52) updated in the status of the AWX CR and then just needed the CR name to be able to pull them out and back them up?

roles/backup/tasks/secrets.yml
```diff
@@ -0,0 +1,124 @@
...
+  register: _postgres_configuration
+
+- name: Set postgres configuration
+  set_fact:
+    database_password: "{{ _postgres_configuration['resources'][0]['data']['password'] | b64decode }}"
+    database_username: "{{ _postgres_configuration['resources'][0]['data']['username'] | b64decode }}"
+    database_name: "{{ _postgres_configuration['resources'][0]['data']['database'] | b64decode }}"
+    database_port: "{{ _postgres_configuration['resources'][0]['data']['port'] | b64decode }}"
+    database_host: "{{ _postgres_configuration['resources'][0]['data']['host'] | b64decode }}"
+    database_type: "{{ _postgres_configuration['resources'][0]['data']['type'] | b64decode }}"
```
chambridge: If a customer supplies the db secret I'm not sure they are required to provide `type`. Might make sense to have a default type of `unmanaged` and set this value if it is defined in a separate item.

ansible/templates/awx-operator.yaml.j2
```diff
@@ -10,3 +10,9 @@
 {% include 'service_account.yml.j2' %}
 
 {% include 'operator.yml.j2' %}
+
+{% include 'crd.yml.j2' %}
```
shanemcd: This looks like a botched rebase. If you look at devel, this has been moved to the top of this file. The other CRDs should also be moved up before the other resources.

molecule/default/prepare.yml
```diff
@@ -11,10 +11,18 @@
...
-    - name: Create Custom Resource Definition
+    - name: Create AWX Custom Resource Definition
       k8s:
         definition: "{{ lookup('file', '/'.join([deploy_dir, 'crds/awx_v1beta1_crd.yaml'])) }}"
 
+    - name: Create AWXBackup Custom Resource Definition
+      k8s:
+        definition: "{{ lookup('file', '/'.join([deploy_dir, 'crds/awxbackup_v1beta1_crd.yaml'])) }}"
+
+    - name: Create AWXBackup Custom Resource Definition
```
shanemcd:
```suggestion
    - name: Create AWXRestore Custom Resource Definition
```

ansible/templates/awx-operator.yaml.j2
rooftopcellist: I had done that intentionally thinking that the order mattered.  turns out the backup and restore CRD's need to exist prior to deploying the operator, so the order doesn't matter anyways where.  I'll put it back so it matches devel.

roles/backup/tasks/init.yml
```diff
@@ -0,0 +1,61 @@
+---
+
+- name: Delete any existing management pod
+  community.kubernetes.k8s:
```
shanemcd:
```suggestion
  k8s:
```

Please remove any FQCNs.

ansible/templates/awx-operator.yaml.j2
rooftopcellist: Oh, I see now, there are duplicates

roles/backup/tasks/init.yml
```diff
@@ -0,0 +1,61 @@
...
+  set_fact:
+    _default_backup_pvc: "{{ meta.name }}-backup-claim"
+
+# by default, it will re-use the old pvc if already created (unless pvc is provided)
+- name: Set PVC to use for backup
+  set_fact:
+    backup_pvc: "{{ tower_backup_pvc | default(_default_backup_pvc, true) }}"
+
+- name: Create PVC for backup
+  community.kubernetes.k8s:
```
shanemcd:
```suggestion
  k8s:
```

roles/backup/tasks/init.yml
```diff
@@ -0,0 +1,61 @@
...
+    kind: PersistentVolumeClaim
+    template: "backup_pvc.yml.j2"
+  when:
+    - tower_backup_pvc == '' or tower_backup_pvc is not defined
+
+- name: Create management pod from templated deployment config
+  community.kubernetes.k8s:
```
shanemcd:
```suggestion
  k8s:
```

roles/backup/tasks/postgres.yml
```diff
@@ -0,0 +1,89 @@
...
+- name: Determine the timestamp for the backup once for all nodes
+  set_fact:
+    now: '{{ lookup("pipe", "date +%F-%T") }}'
+
+- name: Set backup directory name
+  set_fact:
+    backup_dir: "/backups/tower-openshift-backup-{{ now }}"
+
+- name: Create directory for backup
+  community.kubernetes.k8s_exec:
```
shanemcd:
```suggestion
  k8s_exec:
```

roles/backup/tasks/postgres.yml
```diff
@@ -0,0 +1,89 @@
...
+    namespace: "{{ tower_backup_pvc_namespace }}"
+    pod: "{{ meta.name }}-db-management"
+    command: >-
+      mkdir -p {{ backup_dir }}
+
+- name: Precreate file for database dump
+  community.kubernetes.k8s_exec:
```
shanemcd:
```suggestion
  k8s_exec:
```

roles/backup/tasks/postgres.yml
```diff
@@ -0,0 +1,89 @@
...
+    namespace: "{{ tower_backup_pvc_namespace }}"
+    pod: "{{ meta.name }}-db-management"
+    command: >-
+      touch {{ backup_dir }}/tower.db
+
+- name: Set permissions on file for database dump
+  community.kubernetes.k8s_exec:
```
shanemcd:
```suggestion
  k8s_exec:
```

roles/backup/tasks/postgres.yml
```diff
@@ -0,0 +1,89 @@
...
+  set_fact:
+    pgdump: >-
+      pg_dump --clean --create
+      -h {{ awx_postgres_host }}
+      -U {{ awx_postgres_user }}
+      -d {{ awx_postgres_database }}
+      -p {{ awx_postgres_port }}
+
+- name: Write pg_dump to backup on PVC
+  community.kubernetes.k8s_exec:
```
shanemcd:
```suggestion
  k8s_exec:
```

roles/backup/tasks/secrets.yml
```diff
@@ -0,0 +1,124 @@
...
+    src: secret_key_secret.yml.j2
+    dest: "_secrets/secret_key_secret.yml"
+    mode: '0700'
+
+- name: Set secret key template
+  set_fact:
+    secret_key_template: "{{ lookup('file', '_secrets/secret_key_secret.yml') }}"
+
+- name: Write secret_key to pvc
+  community.kubernetes.k8s_exec:
```
shanemcd:
```suggestion
  k8s_exec:
```

roles/backup/tasks/secrets.yml
```diff
@@ -0,0 +1,124 @@
...
+    src: admin_password_secret.yml.j2
+    dest: "_secrets/admin_password_secret.yml"
+    mode: '0700'
+
+- name: Set admin_password template
+  set_fact:
+    admin_password_template: "{{ lookup('file', '_secrets/admin_password_secret.yml') }}"
+
+- name: Write secret_key to pvc
+  community.kubernetes.k8s_exec:
```
shanemcd:
```suggestion
  k8s_exec:
```

roles/backup/tasks/secrets.yml
```diff
@@ -0,0 +1,124 @@
...
+    src: broadcast_websocket_secret.yml.j2
+    dest: "_secrets/broadcast_websocket_secret.yml"
+    mode: '0700'
+
+- name: Set broadcast_websocket template
+  set_fact:
+    broadcast_websocket_template: "{{ lookup('file', '_secrets/broadcast_websocket_secret.yml') }}"
+
+- name: Write broadcast_websocket definition to pvc
+  community.kubernetes.k8s_exec:
```
shanemcd:
```suggestion
  k8s_exec:
```

roles/backup/tasks/secrets.yml
```diff
@@ -0,0 +1,124 @@
...
+    src: postgres_secret.yml.j2
+    dest: "_secrets/postgres_secret.yml"
+    mode: '0700'
+
+- name: Set postgres configuration
+  set_fact:
+    postgres_secret_template: "{{ lookup('file', '_secrets/postgres_secret.yml') }}"
+
+- name: Write postgres configuration to pvc
+  community.kubernetes.k8s_exec:
```
shanemcd:
```suggestion
  k8s_exec:
```

roles/backup/tasks/cleanup.yml
```diff
@@ -0,0 +1,15 @@
+---
+
+# After copying secret files to the PVC, delete the local tmp copies
+- name: Clean up _secrets directory
+  ansible.builtin.file:
+    path: "_secrets"
+    state: absent
+
+- name: Delete any existing management pod
+  community.kubernetes.k8s:
```
shanemcd:
```suggestion
  k8s:
```

roles/backup/tasks/error_handling.yml
```diff
@@ -0,0 +1,16 @@
...
+  set_fact:
+    api_version: '{{ hostvars["localhost"]["inventory_file"].split("/")[4:6] | join("/")  }}'
+    kind: '{{ hostvars["localhost"]["inventory_file"].split("/")[6]  }}'
+
+- name: Determine the timestamp
+  set_fact:
+    now: '{{ lookup("pipe", "date +%FT%TZ") }}'
+
+- name: Emit ocp event with error
+  community.kubernetes.k8s:
```
shanemcd:
```suggestion
  k8s:
```

roles/backup/tasks/awx-cro.yml
```diff
@@ -0,0 +1,38 @@
...
+    src: awx_object.yml.j2
+    dest: "_secrets/awx_object.yml"
+    mode: '0600'
+
+- name: Set AWX object template file as var
+  set_fact:
+    awx_object_template: "{{ lookup('file', '_secrets/awx_object.yml') }}"
+
+- name: Write awx object to pvc
+  community.kubernetes.k8s_exec:
```
shanemcd:
```suggestion
  k8s_exec:
```

roles/restore/tasks/apply_secret.yml
```diff
@@ -0,0 +1,24 @@
+---
+
+- name: Get secret definition from pvc
+  community.kubernetes.k8s_exec:
```
shanemcd:
```suggestion
  k8s_exec:
```

roles/restore/tasks/cleanup.yml
```diff
@@ -0,0 +1,14 @@
+---
+
+- name: Clean up _secrets directory
+  ansible.builtin.file:
+    path: "_definitions"
+    state: absent
+
+- name: Delete any existing management pod
+  community.kubernetes.k8s:
```
shanemcd:
```suggestion
  k8s:
```

roles/restore/tasks/error_handling.yml
```diff
@@ -0,0 +1,16 @@
...
+  set_fact:
+    api_version: '{{ hostvars["localhost"]["inventory_file"].split("/")[4:6] | join("/")  }}'
+    kind: '{{ hostvars["localhost"]["inventory_file"].split("/")[6]  }}'
+
+- name: Determine the timestamp
+  set_fact:
+    now: '{{ lookup("pipe", "date +%FT%TZ") }}'
+
+- name: Emit ocp event with error
+  community.kubernetes.k8s:
```
shanemcd:
```suggestion
  k8s:
```

roles/restore/tasks/init.yml
```diff
@@ -0,0 +1,85 @@
...
+
+    - name: Fail early if pvc is defined but does not exist
+      fail:
+        msg: "{{ error_msg }}"
+  when:
+    - tower_backup_pvc != ''
+    - provided_pvc.resources | length == 0
+
+- name: Delete any existing management pod
+  community.kubernetes.k8s:
```
shanemcd:
```suggestion
  k8s:
```

roles/restore/tasks/init.yml
```diff
@@ -0,0 +1,85 @@
...
+    name: "{{ meta.name }}-db-management"
+    kind: Pod
+    namespace: "{{ tower_backup_pvc_namespace }}"
+    state: absent
+    force: true
+    wait: true
+
+- name: Create management pod from templated deployment config
+  community.kubernetes.k8s:
```
shanemcd:
```suggestion
  k8s:
```

roles/restore/tasks/init.yml
```diff
@@ -0,0 +1,85 @@
...
+    name: "{{ meta.name }}-db-management"
+    kind: Deployment
+    state: present
+    template: "management-pod.yml.j2"
+    wait: true
+
+- name: Check to make sure backup directory exists on PVC
+  community.kubernetes.k8s_exec:
```
shanemcd:
```suggestion
  k8s_exec:
```

roles/restore/tasks/init_awx.yml
```diff
@@ -0,0 +1,32 @@
+---
+
+- name: Get AWX object definition from pvc
+  community.kubernetes.k8s_exec:
```
shanemcd:
```suggestion
  k8s_exec:
```

roles/restore/tasks/postgres.yml
```diff
@@ -0,0 +1,80 @@
...
+
+- name: Set pg_restore command
+  set_fact:
+    psql_restore: >-
+      psql -U {{ awx_postgres_user }}
+      -d template1
+      -p {{ awx_postgres_port }}
+
+- name: Restore database dump to the new postgresql container
+  community.kubernetes.k8s_exec:
```
shanemcd:
```suggestion
  k8s_exec:
```

shanemcd (changes requested):

roles/backup/tasks/awx-cro.yml
```diff
@@ -0,0 +1,38 @@
...
+    awx_api_version: "{{ _awx['apiVersion'] }}"
+
+- name: Set user specified spec
+  set_fact:
+    awx_spec: "{{ _awx['spec'] }}"
+
+- name: Template AWX object definition
+  template:
+    src: awx_object.yml.j2
+    dest: "_secrets/awx_object.yml"
```
shanemcd: What's going to happen when there are more than 1 reconciliation loops running at once?

roles/backup/tasks/cleanup.yml
```diff
@@ -0,0 +1,15 @@
+---
+
+# After copying secret files to the PVC, delete the local tmp copies
+- name: Clean up _secrets directory
+  ansible.builtin.file:
```
shanemcd:
```suggestion
  file:
```

roles/restore/tasks/init_awx.yml
```diff
@@ -0,0 +1,32 @@
...
+    namespace: "{{ tower_backup_pvc_namespace }}"
+    pod: "{{ meta.name }}-db-management"
+    command: >-
+      bash -c "cat '{{ tower_backup_dir }}/awx_object.yml'"
+  register: awx_object
+
+- name: Write temp AWX definition template file
```
shanemcd: I don't think this step is necessary. You could just read the YAML into a variable and pass that directly to the `definition` parameter of the `k8s` module.

roles/restore/tasks/init_awx.yml
```diff
@@ -0,0 +1,32 @@
...
+      {{ awx_object.stdout }}
+    mode: '0600'
+
+- name: Deploy AWX
+  k8s:
+    state: "{{ state | default('present') }}"
+    namespace: "{{ namespace | default('default') }}"
+    apply: yes
+    wait: yes
+    template: "_definitions/awx_object.yml.j2"
```
shanemcd: Following along with my comment above, this could become something like:

```
definition: "{{ awx_object.stdout | from_yaml }}"
```

roles/backup/tasks/awx-cro.yml
shanemcd: Do we need to create a temp directory instead of writing this out to a hardcoded path?

roles/restore/tasks/init_awx.yml
rooftopcellist: That's what I was originally doing, but the backed up YAML files are actually templates.  From what I read and from trying it, it looks like you can only pass a file path to `template:`, not a string.

roles/restore/tasks/init_awx.yml
rooftopcellist: When I tried this before, I used `definition: "{{ awx_object.stdout }}"`, maybe I just needed the `from_yaml`, I'll give that a try.

roles/backup/tasks/awx-cro.yml
rooftopcellist: good point, I can add a random hash at the end of the `_secrets` directory name, same for `_definitions` for the restore role.  

-------------------
The tempfile module is a better solution.

roles/restore/tasks/init_awx.yml
rooftopcellist: @shanemcd passing yaml to a `definition:` field like that works with `from_yaml`, but it won't template the `name` and `namespace` fields.  So we'll still need to template to a file.

Zokormazo:
I deployed a tower called `mytower` on `backup` namespace.

Then apply this:
```yaml
---
kind: AWXBackup
metadata:
  name: awxbackup
  namespace: backup
spec:
  tower_name: mytower
```

My backup pvc ended on default:
```
oc get pvc --all-namespaces
NAMESPACE   NAME                          STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
backup      postgres-mytower-postgres-0   Bound    pvc-b3708b04-24b0-4ab6-a2ef-5f6284e4dc78   8Gi        RWO            standard       61m
default     awxbackup-backup-claim        Bound    pvc-16f63172-8982-4af9-a12f-6161b324730c   5Gi        RWO            standard       17m
```

Using operator built from this branch installed on `awx-qe` project
```
oc get deployments --all-namespaces | grep awx-operator
awx-qe                                           awx-operator                             1/1     1            1           20h
```

The default namespace for the pvc should be the awx instance's namespace imo, not default

rooftopcellist:
@Zokormazo I have updated the backup role to create the backup pvc on the namespace the AWX deploying being backed up is on.  Note, this must be specified by the user via the `tower_backup_pvc_namespace` variable now at restore time.  

Commit: https://github.com/ansible/awx-operator/pull/133/commits/c0410f2ac2ba11a389c27906a2eb2461a7ccfe14

There were also some issues with the postgres backup logic not using the postgres-configuration secret status, I updated that.

roles/backup/tasks/secrets.yml
rooftopcellist: Thanks, I set the default to `unmanaged` here: https://github.com/ansible/awx-operator/pull/133/files#diff-a29bce4ae2beeb7039f9e579e788bf0cdcf975514ce5e7674f5e2b3ba0984e58R50

roles/backup/tasks/secrets.yml
rooftopcellist: Thanks for the suggestion @chambridge, I used your idea of updating the AWX CR with the secret names as statuses.  I added that and use those status names during the backup role now: 
* [Set Status](https://github.com/ansible/awx-operator/pull/133/files#diff-da6ce4fb98bdf57987f09e9b524e00f6600a03d966e5121c7c87c103351abec4R43-R50)
* [Use status to retrieve secret values](https://github.com/ansible/awx-operator/pull/133/files#diff-a29bce4ae2beeb7039f9e579e788bf0cdcf975514ce5e7674f5e2b3ba0984e58R3-R8)

roles/backup/tasks/postgres.yml
rooftopcellist: That makes sense.  I've added this to the data migration, backup, and restore tasks which look for a postgres pod. 

Commit: https://github.com/ansible/awx-operator/pull/133/commits/091c11eba91243a9180d8401f694fcaefac80d59

rooftopcellist:
@Zokormazo Note, the restore-db-management container will still come up in the same namespace the pvc is in, I don't see a way around this.  

Also, deleting the awxrestore object after it has deployed a new awx no longer garbage-collects the new awx instance (ownerRef is now null).

Zokormazo:
> @Zokormazo Note, the restore-db-management container will still come up in the same namespace the pvc is in, I don't see a way around this.
> 
ack
> Also, deleting the awxrestore object after it has deployed a new awx no longer garbage-collects the new awx instance (ownerRef is now null).

Great. btw, could the restore object be autodestroyed? I mean, once we have an restored awx instance, do we need awxrestore CR for something? Would be great if it autodestroys itself (without destroying the backup or the awx instance) if we don't need it anymore

Zokormazo:
I made two backups with these specs:

```yaml
---
kind: AWXBackup
metadata:
  name: awxbackup1
  namespace: tower-qe-1619166985
spec:
  tower_name: tower-1619166988
```

```yaml
---
kind: AWXBackup
metadata:
  name: awxbackup2
  namespace: tower-qe-1619166985
spec:
  tower_name: tower-1619166988
```

As an operator, I would expect just a PVC with two backups, but I have two different pvcs:

```
NAME                                   STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
awxbackup-backup-claim                 Bound    pvc-7f464485-e83b-45de-bd31-3eff4b0db0a2   5Gi        RWO            standard       9m38s
awxbackup2-backup-claim                Bound    pvc-15e78bb4-03b7-43c5-a760-35913bac873f   5Gi        RWO            standard       4m45s
```

Would be possible to use always the same claim? Additionaly, could the default name of the pvc be related to the tower instance we are backing up? Something like `tower-161966988-backups-claim` ?

Ok, I have two backups, so I continue with the restoring process following the instructions on this PR..., so I should apply something similar to this:

```yaml
---
kind: AWXRestore
metadata:
  name: restore1
  namespace: tower-qe-1619166985
spec:
  tower_name: mytower
  tower_backup_pvc: awxbackup2-backup-claim
  tower_backup_pvc_namespace: tower-qe-1619166985
  tower_backup_dir: ??
```

How can I know the tower_backup_dir? :D Ok, I can take it from `oc describe awxbackup awxbackup2`: `/backups/tower-openshift-backup-2021-04-23-09:08:22` .

IMO, here are some UX issues. This spec is great if you just have a backup file from another cluster/on-prem and you want to restore it on ocp. I create a pvc, upload the backup on a known directory on it, apply this spec file and restore it.

But let think on an another scenario: Someones has already running an awx instance deployed w operator, with some backups on it, and it brokes her installation. She wants to restore her last backup. So, the namespace has an awx CR, a postgres PVC, and some AWXBackup CRs.

Would be great is she could:
1. oc delete awx myawx
2. oc delete pvc postgres-myawx

This would teardown the awx instance and the managed db, leaving some pvcs and somes awxbackup CRs there:

```
oc get awxbackup
NAME         AGE
awxbackup-2021-04-23    47m
awxbackup-2021-04-22   1day
```

Would be great if she could restore it applying an spec file that specifies an AWXBackup CR:

```yaml
---
kind: AWXRestore
metadata:
  name: restore1
  namespace: tower-qe-1619166985
spec:
  tower_name: mytower
  tower_backup_namespace: tower-qe-1619166985
  tower_backup_cr: awxbackup-2021-04-23
```

The pvc and directory name are inside the CR, the operator does not need to deal with them if she can just pass that to the spec file.

`tower_backup_namespace` could be optional, defaulting to `metadata.namespace` and making it mandatory just for cross-project restores, so redundant in this use case

Zokormazo:
The restore process works properly btw. But I found another weird thing.

```
oc get secrets
mytower-admin-password           Opaque                                1      6m5s
mytower-app-credentials          Opaque                                3      5m35s
mytower-broadcast-websocket      Opaque                                1      6m5s
mytower-dockercfg-w6d22          kubernetes.io/dockercfg               1      5m34s
mytower-postgres-configuration   Opaque                                6      6m5s
mytower-secret-key               Opaque                                1      6m5s
mytower-token-nlv5k              kubernetes.io/service-account-token   4      5m34s
mytower-token-pdqdp              kubernetes.io/service-account-token   4      5m34s
```
Ok, this were autogenerated during the restore process, but `mytower-admin-password` is a random new one that is not applied on the restored instance, which is ok imo, the restores admin should keep the password it has on the old instance.

But the existence of the secret is confusing. Would be possible to not generate it during restore workflows?

Zokormazo:
Oh, more fun along the way :D

I forgot to note the old password, so I rshed inside the awx pod to set a new password:

```
sh-4.4$ awx-manage update_password --username admin --password admin
Traceback (most recent call last):
  File "/usr/bin/awx-manage", line 11, in <module>
    load_entry_point('awx==4.0.0', 'console_scripts', 'awx-manage')()
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/__init__.py", line 164, in manage
    execute_from_command_line(sys.argv)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/management/commands/update_password.py", line 36, in handle
    res = cp.update_password(options['username'], options['password'])
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/management/commands/update_password.py", line 13, in update_password
    u = User.objects.get(username=username)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/db/models/query.py", line 410, in get
    raise self.model.MultipleObjectsReturned(
django.contrib.auth.models.MultipleObjectsReturned: get() returned more than one User -- it returned 2!
```

Nice, we have two admins now, the new and the old ones :D

Zokormazo:
Those secrets are now owned by the restore CR. I've lost `mytower-admin-password`, `mytower-broadcast-websocket`, `mytower-postgres-configuration` and `mytower-secret-key` removing the awxrestore object

rooftopcellist:
@Zokormazo Thank you for all of the great feedback!  


#### PVC Naming Scheme
I have changed the naming scheme for the backup pvc's to use the `tower_name`, so it will only create one pvc for backups for that AWX now.  

```
NAME                      STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
awx-backup-claim          Bound    pvc-fa04401f-31e4-4f76-be43-5937589424c9   5Gi        RWO            standard       7m6s
postgres-awx-postgres-0   Bound    pvc-9046a75b-7193-4279-af40-97b88c7bf30e   8Gi        RWO            standard       61m
```

$ cat backup-awx.yml 

```
---
kind: AWXBackup
metadata:
  name: awxbackup1
  namespace: chadams-backup
spec:
  tower_name: awx
  # tower_backup_pvc: dne-foobar
  # tower_postgres_configuration_secret: awx-postgres-configuration
```

```
$ cat backup-awx2.yml 
---
kind: AWXBackup
metadata:
  name: awxbackup2
  namespace: chadams-backup
spec:
  tower_name: awx
```

#### Simplify vars for Restore spec (tower_backup_cr)

You can now do a restore by either:
1. Providing the `tower_backup` name (awxbackup includes necessary metadata) and `tower_backup_pvc_namespace`
2. OR, providing the `tower_backup_pvc`, `tower_backup_dir`, and `tower_backup_pvc_namespace`

#### AWXRestore created secrets get garbage collected

I added a step in `cleanup.yml` to set the ownerReference's to null to prevent this from happening.  

> Note: if the owner ref's are removed before the secrets are used to deploy AWX, migrations never completed.

rooftopcellist:
@Zokormazo I am not able to reproduce the multiple admin users issue with the current PR.  That is really weird.  Let me know if you are still seeing that after my changes.  

For the admin-password secret that doesn't match the actual admin password, it looks like that is actually not an issue with the backup & restore logic, but instead with the data migration logic.  

```
# Restored AWX
$ oc -n my-namespace get secrets mytower-admin-password -o jsonpath={.data.password} | base64 --decode
Bffhv3PHeTscKqVzwDSGYNud2DVGcrLd

# AWX that was backed up
$ oc -n chadams-backup get secrets awx-admin-password -o jsonpath={.data.password} | base64 --decode
Bffhv3PHeTscKqVzwDSGYNud2DVGcrLd
```

The admin_password in the db doesn't match either of these, it is the one from the original AWX deployment.  
 The `psql` restore task in the `migrate_data.yml` logic adds an admin user to the database, and when `secret_key_configuration.yml` attempts to create the admin user from the secret, it fails because a user by the name `admin` already exists.  

The admin password is not known at restore time in this case, but we could selectively not create the `<tower_name>-admin_password` secret in the case where the `admin` user already exists.

rooftopcellist:
Re: destroying the awxrestore object after the restore logic has completed.  AFAIK, doing this would require creating another `awxcleanup` custom resource object which would watch for awxrestore objects with the "completed" status.  Since users won't be restoring their deployments too often, I think it is better to leave behind the awxrestore object, it can serve as an audit trail, showing that this deployment was restored from a specific backup.  The user can always delete it manually and the secrets and new awx deployment will remain.  

I'm happy to talk about this more, that's just the conclusion @shanemcd and I came to after chatting about it.

rooftopcellist:
I have changed the default behavior for the admin password.  @Zokormazo Even when a super user is already created, it's password will be updated to reflect the value in the provided/generated `tower_name-admin-password` secret.

Zokormazo:
Something weird is going on with the restore now.

I ended with 6 admins and 2 admin roles.

My steps:

mytower.yaml:
```yaml
---
kind: AWX
metadata:
  name: tower
  namespace: tower-backup
spec:
  deployment_type: awx
  tower_admin_user: admin
  tower_admin_email: test@example.com
  tower_ingress_type: Route
  tower_hostname: tower
  tower_replicas: 3
  tower_web_resource_requirements:
    requests:
      cpu: 500m
      memory: 1Gi
    limits:
      cpu: 500m
      memory: 1Gi
  tower_task_resource_requirements:
    requests:
      cpu: 500m
      memory: 1Gi
    limits:
      cpu: 500m
      memory: 1Gi
  tower_image_pull_policy: Always
```

```
oc apply -n tower-backup -f my-tower.yaml
```

backup-awx.yaml:
```yaml
---
kind: AWXBackup
metadata:
  name: awxbackup1
  namespace: tower-backup
spec:
  deployment_name: tower
```


```
oc apply -n tower-backup -f backup-awx.yaml
oc delete -n tower-backup awx --all
oc delete -n tower-backup secret {tower-admin-password,tower-broadcast-websocket,tower-postgres-configuration,tower-secret-key}
oc delete -n tower-backup pvc postgres-tower-postgres-0
```


awx-restore.yaml:
```yaml
---
kind: AWXRestore
metadata:
  name: restore
  namespace: tower-backup
spec:
  deployment_name: newtower
  backup: awxbackup1
  backup_pvc_namespace: tower-backup
```

```
oc apply -n tower-backup -f awxrestore.yaml
```

After waiting to finish the reconciliation-loop I went to login. The password on the secret does not work, so went to rsh to see what's going on like the other day:

```
sh-4.4$ awx-manage update_password --user admin --password admin
Traceback (most recent call last):
  File "/usr/bin/awx-manage", line 11, in <module>
    load_entry_point('awx==4.0.0', 'console_scripts', 'awx-manage')()
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/__init__.py", line 164, in manage
    execute_from_command_line(sys.argv)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/management/commands/update_password.py", line 36, in handle
    res = cp.update_password(options['username'], options['password'])
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/management/commands/update_password.py", line 13, in update_password
    u = User.objects.get(username=username)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/db/models/query.py", line 410, in get
    raise self.model.MultipleObjectsReturned(
django.contrib.auth.models.MultipleObjectsReturned: get() returned more than one User -- it returned 6!
```

Well, let create a new admin to restore this:
```
sh-4.4$ awx-manage createsuperuser
Username (leave blank to use 'awx'): julen
Email address: julen@landa.eus
Password: 
Password (again): 
Traceback (most recent call last):
  File "/usr/bin/awx-manage", line 11, in <module>
    load_entry_point('awx==4.0.0', 'console_scripts', 'awx-manage')()
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/__init__.py", line 164, in manage
    execute_from_command_line(sys.argv)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/contrib/auth/management/commands/createsuperuser.py", line 61, in execute
    return super().execute(*args, **options)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/contrib/auth/management/commands/createsuperuser.py", line 156, in handle
    self.UserModel._default_manager.db_manager(database).create_superuser(**user_data)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/contrib/auth/models.py", line 162, in create_superuser
    return self._create_user(username, email, password, **extra_fields)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/contrib/auth/models.py", line 145, in _create_user
    user.save(using=self._db)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/contrib/auth/base_user.py", line 66, in save
    super().save(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/db/models/base.py", line 743, in save
    self.save_base(using=using, force_insert=force_insert,
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/db/models/base.py", line 791, in save_base
    post_save.send(
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/dispatch/dispatcher.py", line 173, in send
    return [
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/dispatch/dispatcher.py", line 174, in <listcomp>
    (receiver, receiver(signal=self, sender=sender, **named))
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/signals.py", line 133, in sync_superuser_status_to_rbac
    Role.singleton(ROLE_SINGLETON_SYSTEM_ADMINISTRATOR).members.add(instance)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/models/rbac.py", line 435, in singleton
    role, _ = Role.objects.get_or_create(singleton_name=name, role_field=name)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/db/models/query.py", line 538, in get_or_create
    return self.get(**kwargs), False
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/db/models/query.py", line 410, in get
    raise self.model.MultipleObjectsReturned(
awx.main.models.rbac.MultipleObjectsReturned: get() returned more than one Role -- it returned 2!
```

Zokormazo:
Ok, just noticed that I was testing w/ a custom awx container. Reran the above test without overriding the container and it works properly.

Zokormazo (approved):

rooftopcellist:
I just did another pass through this PR and it is working as expected.  

1. Built & Deployed the awx-operator

```
# Build & Push Operator
operator-sdk build quay.io/chadams/awx-operator:new-crd
docker push quay.io/chadams/awx-operator:new-crd

# Deploy Operator
ansible-playbook ansible/deploy-operator.yml \
    -e pull_policy=Always \
    -e operator_image=quay.io/chadams/awx-operator \
    -e operator_version=new-crd
```

2. Deployed AWX using the operator

```
# Deploy AWX
ansible-playbook ansible/instantiate-awx-deployment.yml -e namespace=chadams-backup -e tower_ingress_type=Route -e development_mode=False -vvv
```

```
$ oc get awx awx -o yaml

...
spec:
  api_version: awx.ansible.com/v1beta1
  deployment_type: awx
  development_mode: false
  kind: AWX
  tower_admin_email: test@example.com
  tower_admin_user: test
  tower_create_preload_data: true
  tower_garbage_collect_secrets: false
  tower_image_pull_policy: IfNotPresent
  tower_ingress_type: Route
  tower_loadbalancer_port: 80
  tower_loadbalancer_protocol: http
  tower_projects_persistence: false
  tower_projects_storage_access_mode: ReadWriteMany
  tower_projects_storage_size: 8Gi
  tower_replicas: 1
  tower_route_tls_termination_mechanism: Edge
  tower_task_privileged: false
status:
  conditions:
  - lastTransitionTime: "2021-05-03T17:54:28Z"
    reason: Successful
    status: "True"
    type: Running
  towerAdminPasswordSecret: awx-admin-password
  towerAdminUser: test
  towerBroadcastWebsocketSecret: awx-broadcast-websocket
  towerImage: quay.io/ansible/awx
  towerMigratedFromSecret: awx-old-postgres-configuration
  towerPostgresConfigurationSecret: awx-postgres-configuration
  towerSecretKeySecret: awx-secret-key
  towerURL: https://awx-chadams-backup.apps.tower.ocp4.testing.ansible.com
  towerVersion: 19.0.0
```

3. Made a user on the resulting AWX deployment

image


4. Created AWXBackup object

```
# Create AWX Backup CRO
# oc create -f backup-awx.yml
```

```
$ oc get awxbackup awxbackup1 -o yaml
...
spec:
  deployment_name: awx
status:
  backupClaim: awx-backup-claim
  backupDirectory: /backups/tower-openshift-backup-2021-05-03-18:03:02
  conditions:
  - ansibleResult:
      changed: 0
      completion: 2021-05-03T18:04:24.858054
      failures: 0
      ok: 3
      skipped: 7
    lastTransitionTime: "2021-05-03T18:02:25Z"
    message: Awaiting next reconciliation
    reason: Successful
    status: "True"
    type: Running
```


5. Created AWXRestore object & namespace (optional)

```
# Create AWX Restore CRO
oc new-project my-namespace
oc create -f restore-awx.yml
```

```
$ oc -n my-namespace get awxrestore restore1 -o yaml
...
spec:
  backup: awxbackup1
  backup_pvc_namespace: chadams-backup
  deployment_name: mytower
status:
  conditions:
  - ansibleResult:
      changed: 0
      completion: 2021-05-03T18:08:01.210994
      failures: 0
      ok: 3
      skipped: 7
    lastTransitionTime: "2021-05-03T18:05:11Z"
    message: Awaiting next reconciliation
    reason: Successful
    status: "True"
    type: Running
  towerRestoreComplete: true
```

6. Logged in to https://mytower-my-namespace.apps.tower.ocp4.testing.ansible.com/#/users to verify that the new user was added.

shanemcd (approved):

PR #137, opened 2021-03-22
Check the default old PostgreSQL configuration

Zokormazo:
Signed-off-by: Julen Landa Alustiza <jlanda@redhat.com>

### DIFF
--- a/roles/installer/tasks/database_configuration.yml
+++ b/roles/installer/tasks/database_configuration.yml
@@ -14,13 +14,25 @@
     name: '{{ meta.name }}-postgres-configuration'
   register: _default_pg_config_resources
 
-- name: Check for old PostgreSQL configuration secret
+- name: Check for specified old PostgreSQL configuration secret
   k8s_info:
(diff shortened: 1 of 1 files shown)


roles/installer/tasks/database_configuration.yml
```diff
@@ -14,14 +14,26 @@
...
-  register: old_pg_config
+  register: _custom_old_pg_config_resources
   when: tower_old_postgres_configuration_secret | length
 
+- name: Check for default old PostgreSQL configuration
+  k8s_info:
+    kind: Secret
+    namespace: '{{ meta.namespace }}'
+    name: '{{ meta.name }}-old-postgres-configuration'
+  register: _default_old_pg_config_resources
```
rooftopcellist, line 30: This makes sense.  In case the user doesn't explicitly specify the name of the old postgres configuration secret, if it is named as recommended in the docs, it will still be picked up by this task.

rooftopcellist (approved):

PR #153, opened 2021-03-26
Handling HTTP loadbalancer

### LINKED ISSUE #151: Handling HTTP loadbalancer
Hello,

At job, we plan to deploy awx using awx-operator. Everything works great except one thing : we want to deploy awx using an HTTP loadbalancer. I tried to find in the documentation the way to do that but it's look like it's not implemented.

I use this YML code to describe the AWX CR : 
```---
kind: AWX
metadata:
  name: awx
  namespace: ${K8S_AWX_NAMESPACE}
spec:
  tower_hostname: ${INTERNAL_AWX_HOST}
  tower_admin_user: admin
  tower_admin_email: XXXX 
  tower_admin_password_secret: ${AWX_ADMIN_PASSWORD}
  tower_broadcast_websocket_secret: awx-broadcast-websocket
  tower_postgres_configuration_secret: awx-postgres-configuration
  tower_ingress_type: LoadBalancer
  tower_loadbalancer_annotations: "networking.gke.io/load-balancer-type: Internal"
  tower_loadbalancer_protocol: http
```

This configuration led me to this error in the awx-operator pods : 
```{
  "ansible_loop_var": "item",
  "changed": false,
  "error": 422,
  "item":
  "tower_service",
  "msg": "Failed to apply object: b'{
    "kind":"Status",
    "apiVersion":"v1",
    "metadata":{},
    "status":"Failure",
    "message":"Service "awx-service" is invalid: spec.ports: Required value",
    "reason":"Invalid",
    "details":{
      "name":"awx-service",
      "kind":"Service",
      "causes":
        [
          {
          "reason":"FieldValueRequired",
          "message":"Required value",
          "field":"spec.ports"
          }
        ]
      }
  }
}
```

Looking in "awx-operator/roles/installer/templates/tower_service.yaml.j2" I found that the loadbalancer with HTTP case is not implemented.

Do I missed something ? Is there a way to do it using the default variable ?

If it's not possible, I will make a PR to develop this feature :)

billett83:
fixes #151

Amended the logic for the service template around the load balancer to include a missing default HTTP setup. I have also added the tower_loadbalancer_port reference into the logic which allows users to change the port if required.

The readme has also be updated to include extra information around the capabilities of the Load Balancer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,9 @@ The following variables are customizable to specify the TLS termination procedur
 | tower_loadbalancer_protocol    | Protocol to use for Loadbalancer ingress | http          |
 | tower_loadbalancer_port        | Port used for Loadbalancer ingress       | 80            |
 
+When setting up a Load Balancer for HTTPS you will be required to set the `tower_loadbalancer_port` to move the port away from `80`.
+
+The HTTPS Load Balancer also uses SSL termination at the Load Balancer level and will offload traffic to AWX over HTTP.
 
 ### Database Configuration
 
--- a/roles/installer/templates/tower_service.yaml.j2
+++ b/roles/installer/templates/tower_service.yaml.j2
@@ -25,10 +25,15 @@ spec:
       name: https
 {% endif %}
 {% if tower_ingress_type | lower == 'loadbalancer' and tower_loadbalancer_protocol | lower == 'https' %}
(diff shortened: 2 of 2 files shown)


shanemcd (approved):
Thanks 😄

PR #165, opened 2021-04-01
Add support for nodeSelector and tolerations

athak:
This PR adds support for adding `nodeSelector` and `tolerations` to the AWX pods.

### DIFF
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -38,6 +38,22 @@ tower_route_host: ''
 
 tower_hostname: '{{ deployment_type }}.example.com'
 
+# Add a nodeSelector for the AWX pods. It must match a node's labels for the pod
+# to be scheduled on that node. Specify as literal block. E.g.:
+# tower_node_selector: |
+#   disktype: ssd
+#   kubernetes.io/arch: amd64
+#   kubernetes.io/os: linux
+tower_node_selector: ''
+
+# Add node tolerations for the AWX pods. Specify as literal block. E.g.:
+# tower_tolerations: |
+#   - key: "dedicated"
+#     operator: "Equal"
+#     value: "AWX"
+#     effect: "NoSchedule"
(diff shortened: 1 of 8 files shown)


roles/installer/defaults/main.yml
```diff
@@ -38,6 +38,22 @@ tower_route_host: ''
...
+tower_node_selector: ''
+
+# Add node tolerations for the AWX pods. Specify as literal block. E.g.:
+# tower_tolerations: |
+#   - key: "dedicated"
+#     operator: "Equal"
+#     value: "AWX"
+#     effect: "NoSchedule"
+tower_tolerations: ''
+
```
tchellomello, line 56: @athak would you be willing to add this section also on the docs at https://github.com/ansible/awx-operator#advanced-configuration? I think that will be important for our users to have a reference point on how to configure it without looking at the code.

Spredzy:
@athak looks great. Thanks for the PR. In addition to @tchellomello 's comment, could you update:

  * `deploy/olm-catalog/awx-operator/manifests/awx.ansible.com_awxs_crd.yaml`
  * `deploy/olm-catalog/awx-operator/manifests/awx-operator.clusterserviceversion.yaml` (Mark them as both `advanced` and `hiddden` - just like: https://github.com/ansible/awx-operator/blob/devel/deploy/olm-catalog/awx-operator/manifests/awx-operator.clusterserviceversion.yaml#L294-L298)
  
 Then looks good to me for inclusion

roles/installer/defaults/main.yml
athak: Added in 2a3f700.

athak:
@Spredzy addressed in 2a3f700.

tchellomello (approved):
LGTM! Thanks @athak

shanemcd:
Going to close and reopen to trigger the new GitHub Actions check.

PR #166, opened 2021-04-03
fix wrong conf file for SUPERVISOR_WEB_CONFIG_PATH in kubernetes deployment file

ilijamt:
When a reload is issued it looks in the wrong configuration file, this is only referenced in the task container should it also be referenced in the web container maybe? 

AWX: https://github.com/ansible/awx/blob/9a61c4368784c421e47551e4175020ced116f56e/awx/main/utils/reload.py#L20

Operator: https://github.com/ansible/awx-operator/blob/210ac2c41955f96a4fa3f748b54879483d1bae62/roles/installer/templates/tower_deployment.yaml.j2#L152-L153

```
2021-04-03 11:17:54,377 DEBUG    [-] awx.main.utils.reload Issuing command to restart services, args=['supervisorctl', '-c', '/supervisor.conf', 'restart', 'tower-processes:awx-rsyslogd']
2021-04-03 11:17:54,513 ERROR    [-] awx.main.utils.reload supervisorctl restart awx-rsyslogd errored with exit code `2`, stdout:
```

### DIFF
--- a/roles/installer/templates/tower_deployment.yaml.j2
+++ b/roles/installer/templates/tower_deployment.yaml.j2
@@ -150,7 +150,7 @@ spec:
 {% endif %}
           env:
             - name: SUPERVISOR_WEB_CONFIG_PATH
-              value: "/supervisor.conf"
+              value: "/etc/supervisord_task.conf"
             - name: AWX_SKIP_MIGRATIONS
               value: "1"
             - name: MY_POD_UID
--- a/roles/installer/templates/tower_deployment.yaml.j2
+++ b/roles/installer/templates/tower_deployment.yaml.j2
@@ -150,7 +150,7 @@ spec:
 {% endif %}
           env:
             - name: SUPERVISOR_WEB_CONFIG_PATH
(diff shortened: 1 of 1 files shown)


roles/installer/templates/tower_deployment.yaml.j2
```diff
@@ -150,7 +150,7 @@ spec:
 {% endif %}
           env:
             - name: SUPERVISOR_WEB_CONFIG_PATH
-              value: "/supervisor.conf"
+              value: "/etc/supervisord_task.conf"
```
shanemcd:
```suggestion
              value: "/etc/supervisord.conf"
```

shanemcd (changes requested):
I saw this too, just haven't had time to open up a PR. Made a suggestion with the correct value.

shanemcd (approved):
Looks like I had the ability to directly commit the suggestion. Will merge after checks pass.

ilijamt:
Shouldn't it be supervisord_task.conf ? as the container inside it's running supervisor with the supervisord_task file

PR #184, opened 2021-04-07
add tower_image_pull_secret as property of AWX resource

### LINKED ISSUE #183: tower_image_pull_secret is no property of AWX CRD
The AWX CRD is missing the tower_image_pull_secret property, so it's not possible to pass a pull secret to the awx deployment.
This prevents using custom awx images from private docker registries.

no-12:
fixes #183 
Tested with manually applying a awx resource with the new property "tower_image_pull_secret" set.
The secret of Type kubernetes.io/dockerconfigjson was correctly listed in the resulting deployment.
And pulling an AWX image from a private Registry was possible.

### DIFF
--- a/ansible/templates/crd.yml.j2
+++ b/ansible/templates/crd.yml.j2
@@ -134,6 +134,9 @@ spec:
                     - never
                     - IfNotPresent
                     - ifnotpresent
+                tower_image_pull_secret:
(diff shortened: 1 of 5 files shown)


no-12:
fixes #183

shanemcd (approved):

Issue #175, opened 2021-04-05
Provide way for additional configmaps to be mounted on the control plane


shanemcd:
There have been several cases where folks want to be able to bind-mount random stuff into the control plane:

- https://github.com/ansible/awx/issues/9807
- https://github.com/ansible/awx/issues/9753

We should come up with a generalized solution for these kinds of use cases.

billett83:
Additional use case - Setting up realms in /etc/krb5.conf

shanemcd:
Looks like this is already a thing! https://github.com/ansible/awx/issues/9807#issuecomment-819014975

Issue #150, opened 2021-03-26
SSL Termination at Load Balancer


billett83:
I'm running into an issue whereby I'm terminating SSL at the Load balancer level. This is causing 400 Bad Request errors during CallBacks from newly provisioned servers.

Reading through the Ansible documentation (https://docs.ansible.com/ansible-tower/3.8.2/html/installandreference/proxy-support.html#ir-proxy-support) It advises the following:

```
If using SSL termination at the load balancer and forwarding traffic to a different port on the tower node (443 -> 80), set the following values in the /etc/tower/conf.d/custom.py file accordingly:

USE_X_FORWARDED_PORT = True
USE_X_FORWARDED_HOST = True
```

Is there a current method to get these parameters into the running POD during the deployment process?

Spredzy:
`USE_X_FORWARDED_PORT = True` is set here https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/tower_config.yaml.j2#L85 - but no, there is no mechanism to specify it at that point.

In order to provide more flexibility, and not have to support each and every possible config from an operator perspective, I think we should provide a way for the user to provide a settings configmap, where user would be able to add customization one needs. (If something overwrite default config in user ConfigMap, then it should take precedence)

What would you think of such an approach @billett83 ?

billett83:
Thank you @Spredzy . I did come across this one setting whilst trawling through the POD yesterday.

Yes, this approach sounds feasible

Spredzy:
@tchellomello something you are interested working on ? If not let me know andd I'll assign myself.

tchellomello:
@Spredzy yes, I'll work on it

tchellomello:
@Spredzy, @gamuniz and I spoke today to provide a `custom.py` to override the configmap and allow users to modify it. Originally his suggestion was for `insigths*` but it can be extended for any option pretty much. Since he already started working on this PR, @gamuniz++ will submit the PR

shanemcd:
@Spredzy @tchellomello @gamuniz Opened up https://github.com/ansible/awx-operator/issues/175 to capture the last part of this thread.

tchellomello:
Closing this as it was documented on PR https://github.com/ansible/awx-operator/pull/214

PR #185, opened 2021-04-07
Configure custom settings

gamuniz:
Added ConfigMap option to allow for overriding default settings via custom.py


gamuniz:
potentially fixes https://github.com/ansible/awx-operator/issues/175 https://github.com/ansible/awx-operator/issues/150

roles/installer/tasks/tower_custom_settings.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: Retrieve ConfigMap
+  community.kubernetes.k8s_info:
```
shanemcd, line 3:
```suggestion
  k8s_info:
```

The FQCN shouldn't be needed here. I recently went through and removed all of these occurrences. We're going to be moving to `kubernetes.core` in the near future.

README.md
```diff
@@ -375,6 +376,35 @@ To create the secret, you can use the command below:
...
+To create a ConfigMap, you can use the example yaml file as a reference:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: awx-custom-settings
+  namespace: default
+data:
+  custom.py: |
```
shanemcd, line 396: I'm wondering if we can make this a bit more flexible by making this key the full path the file should be bind-mounted to. So in your case here it would be:

```suggestion
  '/var/lib/awx/custom.py': |
```

If we did this I think it would fully satisfy the ask in https://github.com/ansible/awx-operator/issues/175

shanemcd: Stating the obvious: If we did this we would need to rework the logic around how this is consumed inside of the Operator. LMK if you have any questions about what I mean.

README.md
```diff
@@ -375,6 +376,35 @@ To create the secret, you can use the command below:
 ```sh
 # kubectl create secret generic <resourcename>-ldap-ca-cert --from-file=ldap-ca.crt=<PATH/TO/YOUR/CA/PEM/FILE>
 ```
+### Tower Custom Settings
```
shanemcd, line 379:
```suggestion
### Custom Settings
```

README.md
```diff
@@ -26,6 +26,7 @@ An [Ansible AWX](https://github.com/ansible/awx) operator for Kubernetes built w
          * [Containers Resource Requirements](#containers-resource-requirements)
          * [LDAP Certificate Authority](#ldap-certificate-authority)
          * [Persisting Projects Directory](#persisting-projects-directory)
+         * [Custom Setting Configuration](#tower-custom-settings)
```
shanemcd, line 29:
```suggestion
         * [Custom Setting Configuration](#custom-settings)
```

shanemcd: A little on the nitpicky side, but we should be careful about places we say "Tower" 

The variable names are fine (Not really - we need to go through and remove all references to `tower_` at some point soon, but I'd prefer to follow the existing pattern to make find/replace easier)

roles/installer/templates/tower_deployment.yaml.j2
```diff
@@ -155,6 +173,12 @@ spec:
             - name: awx-devel
               mountPath: "/awx_devel"
 {% endif %}
+{% if discovered_tower_custom_settings %}
+            - name: "{{meta.name }}-custom-settings"
```
shanemcd, line 177:
```suggestion
            - name: "{{ meta.name }}-custom-settings"
```

roles/installer/tasks/tower_custom_settings.yml
```diff
@@ -0,0 +1,11 @@
...
+    kind: ConfigMap
+    namespace: '{{ meta.namespace }}'
+    name: '{{ tower_custom_settings }}'
+  register: custom_settings
+- name: Check for custom settings in configmap
```
shanemcd, line 8:
```suggestion

- name: Check for custom settings in configmap
```

roles/installer/tasks/main.yml
```diff
@@ -11,6 +11,10 @@
   when:
     - ldap_cacert_secret != ''
 
+- name: Load Custom Setting Configmap
+  include_tasks: tower_custom_settings.yml
+  when: tower_custom_settings != ''
```
shanemcd, line 16: We have a pattern that we follow for other secrets where we look for the default name, and allow the user to supply a custom name if they want to. May want to do that here.

shanemcd:
We may have tried to reimplement an existing feature here 😓 https://github.com/ansible/awx/issues/9807#issuecomment-819014975

PR #198, opened 2021-04-09
Added support to override PosgreSQL sslmode

### LINKED ISSUE #188: Cannot change Postgres SSL mode from operator
It's currently not possible to change `pg_sslmode` from the operator spec, which is causing trouble when trying to connect to an external Azure managed Postgres instance

tchellomello:
cc: @01100010011001010110010101110000 @gamuniz

Fixes: #188

This PR allows the user to select the PostgreSQL sslmode via `spec`.

### DIFF
--- a/roles/installer/tasks/database_configuration.yml
+++ b/roles/installer/tasks/database_configuration.yml
@@ -78,6 +78,7 @@
     awx_postgres_database: "{{ pg_config['resources'][0]['data']['database'] | b64decode }}"
     awx_postgres_port: "{{ pg_config['resources'][0]['data']['port'] | b64decode }}"
     awx_postgres_host: "{{ pg_config['resources'][0]['data']['host'] | b64decode }}"
+    awx_postgres_sslmode: "{{ pg_config['resources'][0]['data']['sslmode'] |  default('prefer'|b64encode) | b64decode }}"
 
 - name: Look up details for this deployment
   k8s_info:
--- a/README.md
+++ b/README.md
@@ -207,8 +207,11 @@ stringData:
   database: <desired database name>
   username: <username to connect as>
   password: <password to connect with>
+  sslmode: prefer
 type: Opaque
 ```
+
+**Note**: The variable `sslmode` is valid for `external` databases only. The allowed values are: `prefer`, `disable`, `allow`, `require`, `verify-ca`, `verify-full`.
 
 #### Migrating data from an old AWX instance
 
--- a/roles/installer/templates/credentials.py.j2
+++ b/roles/installer/templates/credentials.py.j2
@@ -7,7 +7,7 @@ DATABASES = {
         'PASSWORD': "{{ awx_postgres_pass | quote }}",
         'HOST': '{{ awx_postgres_host }}',
         'PORT': "{{ awx_postgres_port }}",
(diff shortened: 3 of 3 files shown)


Spredzy (changes requested):
Great work and awesome PR. However, philosophically, I think I would rather prefer to have this value being specified in the PostgreSQL configuration secret provided by the user (when it applies) and add an extra line here[1]

```
awx_postgres_sslmode: "{{ pg_config['resources'][0]['data']['sslmode'] | b64decode }}"
```
and the following line [2] updated to:

```
        'OPTIONS': { 'sslmode': '{{ awx_postgres_sslmode|default("prefer") }}',
```

To me, this path feels more like what I would expect. (ie the secret configuration to contain everything configuration wise)

[1] https://github.com/ansible/awx-operator/blob/devel/roles/installer/tasks/database_configuration.yml#L74-L80
[2] https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/credentials.py.j2#L10

tchellomello:
> Great work and awesome PR. However, philosophically, I think I would rather prefer to have this value being specified in the PostgreSQL configuration secret provided by the user (when it applies) and add an extra line here[1]
> 
> ```
> awx_postgres_sslmode: "{{ pg_config['resources'][0]['data']['sslmode'] | b64decode }}"
> ```
> 
> and the following line [2] updated to:
> 
> ```
>         'OPTIONS': { 'sslmode': '{{ awx_postgres_sslmode|default("prefer") }}',
> ```
> 
> To me, this path feels more like what I would expect. (ie the secret configuration to contain everything configuration wise)
> 
> [1] https://github.com/ansible/awx-operator/blob/devel/roles/installer/tasks/database_configuration.yml#L74-L80
> [2] https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/credentials.py.j2#L10

Yes, @Spredzy that makes sense. We can expose the PostgreSQL settings via UI when we provide the same option for the `managed` DB. As you mentioned, after analyzing your suggestion, it does make more sense directly to the `secret` as the `ssl` for now is only supported on `external` database, therefore, it can be a little confusing or generate the wrong expectation when deploying `managed` databases.

Spredzy (approved):
Thank you. Small nit inside

roles/installer/tasks/database_configuration.yml
```diff
@@ -78,6 +78,7 @@
     awx_postgres_database: "{{ pg_config['resources'][0]['data']['database'] | b64decode }}"
     awx_postgres_port: "{{ pg_config['resources'][0]['data']['port'] | b64decode }}"
     awx_postgres_host: "{{ pg_config['resources'][0]['data']['host'] | b64decode }}"
+    awx_postgres_sslmode: "{{ pg_config['resources'][0]['data']['sslmode'] | b64decode }}"
```
Spredzy: @tchellomello nit: If the entry does not exist. Would that fail saying keyError `sslmode` does not exist ?

tchellomello: Yes it would. So I've updated the PR to ensure this scenario. Tested with and without the `sslmode`  in the secret and worked as expected.

shanemcd (approved):

tchellomello:
Thanks guys.. going to merge it!

PR #207, opened 2021-04-13
adding service account image pull secret

techBeck03:
This PR adds an image pull secret to the service account associated with a given AWX deployment.  This helps get around the daily pull limits of docker registries such as DockerHub by allowing a user to use their registered accounts.  The current `tower_image_pull_secret` only applies to awx images and does not apply to images such as `postgres` or `redis`

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,34 +7,37 @@ An [Ansible AWX](https://github.com/ansible/awx) operator for Kubernetes built w
 # Table of Contents
 
 <!--ts-->
-* [AWX Operator](#awx-operator)
-* [Table of Contents](#table-of-contents)
-   * [Purpose](#purpose)
-   * [Usage](#usage)
-      * [Basic Install](#basic-install)
-      * [Admin user account configuration](#admin-user-account-configuration)
-      * [Network and TLS Configuration](#network-and-tls-configuration)
-         * [Ingress Type](#ingress-type)
-         * [TLS Termination](#tls-termination)
-      * [Database Configuration](#database-configuration)
-         * [External PostgreSQL Service](#external-postgresql-service)
-         * [Migrating data from an old AWX instance](#migrating-data-from-an-old-awx-instance)
-         * [Managed PostgreSQL Service](#managed-postgresql-service)
-      * [Advanced Configuration](#advanced-configuration)
-         * [Deploying a specific version of AWX](#deploying-a-specific-version-of-awx)
-         * [Privileged Tasks](#privileged-tasks)
-         * [Containers Resource Requirements](#containers-resource-requirements)
-         * [LDAP Certificate Authority](#ldap-certificate-authority)
-         * [Persisting Projects Directory](#persisting-projects-directory)
-   * [Development](#development)
-      * [Testing](#testing)
-         * [Testing in Docker](#testing-in-docker)
-         * [Testing in Minikube](#testing-in-minikube)
-      * [Generating a bundle](#generating-a-bundle)
-   * [Release Process](#release-process)
-      * [Build a new release](#build-a-new-release)
-      * [Build a new version of the operator yaml file](#build-a-new-version-of-the-operator-yaml-file)
-   * [Author](#author)
+- [AWX Operator](#awx-operator)
+- [Table of Contents](#table-of-contents)
+  - [Purpose](#purpose)
+  - [Usage](#usage)
+    - [Basic Install](#basic-install)
+    - [Admin user account configuration](#admin-user-account-configuration)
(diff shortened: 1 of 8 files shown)


tchellomello:
Thanks for your PR @techBeck03, however, there is already a `imagePullSecrets` configured at the deployment level https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/tower_deployment.yaml.j2#L31-L34 instead of the `serviceAccoutName`. 

I believe the reason why we are doing it at the deployment level instead of the `serviceAccount` one is due to the `serviceAccount` being totally dynamic. 

https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/tower_service_account.yaml.j2#L1-L11
```yaml
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: '{{ meta.name }}'
  namespace: '{{ meta.namespace }}'
  labels:
    app.kubernetes.io/name: '{{ meta.name }}'
    app.kubernetes.io/part-of: '{{ meta.name }}'
```
 and https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/tower_deployment.yaml.j2#L30
```yaml
      serviceAccountName: '{{ meta.name }}'
```

@techBeck03 was the current implementation not sufficient to address your requirements?

techBeck03:
> Thanks for your PR @techBeck03, however, there is already a `imagePullSecrets` configured at the deployment level https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/tower_deployment.yaml.j2#L31-L34 instead of the `serviceAccoutName`.
> 
> I believe the reason why we are doing it at the deployment level instead of the `serviceAccount` one is due to the `serviceAccount` being totally dynamic.
> 
> https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/tower_service_account.yaml.j2#L1-L11
> 
> ```yaml
> ---
> apiVersion: v1
> kind: ServiceAccount
> metadata:
>   name: '{{ meta.name }}'
>   namespace: '{{ meta.namespace }}'
>   labels:
>     app.kubernetes.io/name: '{{ meta.name }}'
>     app.kubernetes.io/part-of: '{{ meta.name }}'
>     app.kubernetes.io/managed-by: awx-operator
>     app.kubernetes.io/component: awx
> ```
> 
> and https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/tower_deployment.yaml.j2#L30
> 
> ```yaml
>       serviceAccountName: '{{ meta.name }}'
> ```
> 
> @techBeck03 was the current implementation not sufficient to address your requirements?

Unfortunately this was not sufficient for me because the current `tower_image_pull_secret` doesn't apply to the `postgres` StatefulSet as mentioned in my initial comment.  If you add the `tower_image_pull_secret` to all image pulls then it would work.

Thanks!

tchellomello:
@techBeck03 yes, I see where you were going with this. I would say since we already have the `imagePullSecrets` implemented and to avoid having another redundant setting or modify something that would introduce a `break change`, I would just add the `imagePullSecret`  on the `PostgreSQL` deployment. 

Let's see with the others the direction they think we should go, but I think the above would provide the best approach and the least pain. 

Thanks again @techBeck03 !

techBeck03:
> @techBeck03 yes, I see where you were going with this. I would say since we already have the `imagePullSecrets` implemented and to avoid having another redundant setting or modify something that would introduce a `break change`, I would just add the `imagePullSecret` on the `PostgreSQL` deployment.
> 
> Let's see with the others the direction they think we should go, but I think the above would provide the best approach and the least pain.
> 
> Thanks again @techBeck03 !

My only other argument for my approach is you wouldn't have to ensure future deployments/statefulsets/etc have the `imagePullSecret` block.  All image pulls would inherit the pull secret from the service account using my proposed changes.  Thanks!

README.md
```diff
@@ -7,34 +7,37 @@ An [Ansible AWX](https://github.com/ansible/awx) operator for Kubernetes built w
 # Table of Contents
 
 <!--ts-->
-* [AWX Operator](#awx-operator)
```
shanemcd, line 10: This makes the diff unnecessarily noisy. Please revert this chunk.

shanemcd:
Actually, I agree with @tchellomello's points above. Please implement his feedback and we'll get this approved and merged.

tchellomello (changes requested):
@techBeck03

tchellomello:
@techBeck03 are you still working on this one?

tchellomello:
Closing this one as feature has been added at https://github.com/ansible/awx-operator/pull/295

Thanks @techBeck03

PR #208, opened 2021-04-13
Include support for nodeSelector and tolerations in AWX Postgres pod

a-eperez:
This change is based on `PR` #165 . It includes support for `nodeSelector` and `tolerations` for the `Postgres pod` that deploys awx-operator.

### DIFF
--- a/roles/installer/templates/tower_postgres.yaml.j2
+++ b/roles/installer/templates/tower_postgres.yaml.j2
@@ -60,6 +60,14 @@ spec:
             - name: postgres
               mountPath: '{{ tower_postgres_data_path | dirname }}'
               subPath: '{{ tower_postgres_data_path | dirname | basename }}'
+{% if tower_node_selector %}
+      nodeSelector:
+        {{ tower_node_selector | indent(width=8) }}
+{% endif %}
+{% if tower_tolerations %}
+      tolerations:
+        {{ tower_tolerations | indent(width=8) }}
+{% endif %}  
   volumeClaimTemplates:
     - metadata:
         name: postgres
--- a/roles/installer/templates/tower_postgres.yaml.j2
+++ b/roles/installer/templates/tower_postgres.yaml.j2
@@ -66,8 +66,8 @@ spec:
 {% endif %}
 {% if tower_postgres_tolerations %}
       tolerations:
-        {{ tower_tolerations | indent(width=8) }}
-{% endif %}  
+        {{ tower_postgres_tolerations | indent(width=8) }}
+{% endif %}
   volumeClaimTemplates:
     - metadata:
         name: postgres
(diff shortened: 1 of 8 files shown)


roles/installer/templates/tower_postgres.yaml.j2
```diff
@@ -60,6 +60,14 @@ spec:
             - name: postgres
               mountPath: '{{ tower_postgres_data_path | dirname }}'
               subPath: '{{ tower_postgres_data_path | dirname | basename }}'
+{% if tower_node_selector %}
```
tchellomello: @a-eperez I believe you want to use `tower_postgres_selector` right?

roles/installer/templates/tower_postgres.yaml.j2
```diff
@@ -60,6 +60,14 @@ spec:
...
+      nodeSelector:
+        {{ tower_node_selector | indent(width=8) }}
+{% endif %}
+{% if tower_tolerations %}
```
tchellomello: Same here should be `tower_postgres_tolerations`

roles/installer/templates/tower_postgres.yaml.j2
```diff
@@ -60,6 +60,14 @@ spec:
             - name: postgres
               mountPath: '{{ tower_postgres_data_path | dirname }}'
               subPath: '{{ tower_postgres_data_path | dirname | basename }}'
+{% if tower_node_selector %}
+      nodeSelector:
+        {{ tower_node_selector | indent(width=8) }}
```
tchellomello: And here

roles/installer/templates/tower_postgres.yaml.j2
```diff
@@ -60,6 +60,14 @@ spec:
...
+{% endif %}
+{% if tower_tolerations %}
+      tolerations:
+        {{ tower_tolerations | indent(width=8) }}
```
tchellomello: And here

tchellomello (changes requested):
@a-eperez Thanks for your PR!! 

Please have the PR addressed and I think it will be good to go! Thanks!

roles/installer/templates/tower_postgres.yaml.j2
a-eperez: Hi, this is indeed a mistake. I have fixed it in 1cc47f7. Thank you.

roles/installer/templates/tower_postgres.yaml.j2
a-eperez: Fixed it in d4d9d2a.

roles/installer/templates/tower_postgres.yaml.j2
a-eperez: Fixed it in 1cc47f7.

roles/installer/templates/tower_postgres.yaml.j2
a-eperez: Fixed it in d4d9d2a.

tchellomello (approved):
LGTM!! Requesting another ack from  @shanemcd  and @Spredzy so it can be merged!

PR #221, opened 2021-04-16
Use postgres config values

rooftopcellist:
This adds the ability for users to specify custom postgres config values for managed database deployments.  With this PR, these config values will be honored:
* database
* username
* password
* port

Previously, the tower_postgres.yml.j2 only looked in the default location for the tower postgres configuration secret.

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,7 +233,7 @@ stringData:
 type: Opaque
 ```
 
-> It is possible to set a specific username, password, port, or database, but still have the database managed by the operator. In this case, when creating the postgres-configuration secret, the `type: managed` field should be added.  
+> It is possible to set a specific username, password, or database, but still have the database managed by the operator. In this case, when creating the postgres-configuration secret, the `type: managed` field should be added.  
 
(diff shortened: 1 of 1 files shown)


rooftopcellist:
There were a couple typos, those are fixed now, CI is passing.

roles/installer/templates/tower_postgres.yaml.j2
```diff
@@ -54,7 +54,7 @@ spec:
             - name: POSTGRES_HOST_AUTH_METHOD
               value: '{{ postgres_host_auth_method }}'
           ports:
-            - containerPort: 5432
+            - containerPort: {{ awx_postgres_port | default('5432')}}
```
rooftopcellist: The quotes around this port var needed to be removed, and the default port was added as an extra precaution.

rooftopcellist:
The original changes this PR introduced got merged as part of the backup-restore work (https://github.com/ansible/awx-operator/pull/133).  This PR is no longer needed.

Issue #182, opened 2021-04-07
Please add instructions on how to update awx via awx-operator with kubertenetes without loosing data


OneHappyForever:
I have only seen instructions on how to install it the first time. 

Unless I run minikube delete, which will delete the files as well, is there a way to update awx via awx operator that will keep all the configs, postgresql database, etc. intact?

If there is, please direct me to the page, as I cannot find any easy way to do so, in any wiki/docs, either on github on or the ansible website.

I believe this will be immensely useful not just for me, but for all users.

shanemcd:
This is kind of already documented here: https://github.com/ansible/awx-operator#deploying-a-specific-version-of-awx

But I would caution against doing this. Certain versions of AWX will inherently require a new version of the operator. I think we may want to consider versioning the operator and awx together to make this more apparent.

@Spredzy thoughts?

MrAmbiG:
@shanemcd What you are suggesting is build and maintain our own custom image, we were hoping something like mentioning the version in the my-awx.yaml file itself. I do understand that an older version of operator may not support an awx release which got released after the operator but from an end user perspective, a backward compatibility with 3-4 versions do make a lot of sense..
```
---
apiVersion: awx.ansible.com/v1beta1
kind: AWX
metadata:
  name: awx
spec:
 version: 18
```
`kubectl apply -f my-awx.yaml`

shanemcd:
@MrAmbiG 

- This would not require you to build your own image. Just point to a different tag like `quay.io/ansible/awx:<whatever>`.
- We cannot ever guarantee any kind of backwards compatibility. This is a fast moving upstream project, and you will need to assume the risk of using bleeding edge software.

MrAmbiG:
@shanemcd 
branch tag: 0.8.0
file: awx-operator.yaml line 118 onwards
```
                # tower_ee_images:
                  # description: Registry path to the Execution Environment container to use
                  # type: array
                  # items:
                  #   type: object
                  #   properties:
                  #     name:
                  #       type: string
                  #     image:
                  #       type: string
                tower_ee_images:
                  - name: awx19
                     image: quay.io/ansible/awx
```
result:
[...]

PR #234, opened 2021-04-21
Stop telling folks to use devel

shanemcd:


tchellomello:
@shanemcd  seems to address the issue #182

tchellomello (approved):
LGTM! Yes, we need this!!

PR #241, opened 2021-04-23
Support custom config

kwangil-ha:
- Support rule path setting in ingress
- Support custom secret key setting in secret
- Support run as user and affinity setting in deployment

### DIFF
--- a/roles/installer/templates/tower_deployment.yaml.j2
+++ b/roles/installer/templates/tower_deployment.yaml.j2
@@ -117,8 +117,9 @@ spec:
         - image: '{{ tower_image }}'
           name: '{{ meta.name }}-task'
           imagePullPolicy: '{{ tower_image_pull_policy }}'
-{% if tower_task_privileged == true %}
           securityContext:
+            runAsUser: 0
+{% if tower_task_privileged == true %}
             privileged: true
 {% endif %}
 {% if tower_task_command %}
@@ -186,6 +187,8 @@ spec:
         - image: '{{ tower_ee_images[0].image }}'
           name: '{{ meta.name }}-ee'
           imagePullPolicy: '{{ tower_image_pull_policy }}'
+          securityContext:
+            runAsUser: 0
           args: ['receptor', '--config', '/etc/receptor.conf']
           volumeMounts:
             - name: "{{ meta.name }}-receptor-config"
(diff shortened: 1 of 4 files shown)


roles/installer/templates/tower_deployment.yaml.j2
```diff
@@ -117,8 +117,9 @@ spec:
         - image: '{{ tower_image }}'
           name: '{{ meta.name }}-task'
           imagePullPolicy: '{{ tower_image_pull_policy }}'
-{% if tower_task_privileged == true %}
           securityContext:
+            runAsUser: 0
```
shanemcd, line 121: I understand that probably made it work for *you*, but we aren't going to run as root by default. Please make this a configurable, where the default behavior should remain unchanged.

roles/installer/templates/tower_deployment.yaml.j2
```diff
@@ -186,6 +187,8 @@ spec:
         - image: '{{ tower_ee_images[0].image }}'
           name: '{{ meta.name }}-ee'
           imagePullPolicy: '{{ tower_image_pull_policy }}'
+          securityContext:
+            runAsUser: 0
```
shanemcd, line 191: Same as my comment above.

roles/installer/templates/tower_ingress.yaml.j2
```diff
@@ -19,7 +19,7 @@ spec:
     - host: '{{ tower_hostname }}'
       http:
         paths:
-          - path: /
+          - path: '{{ tower_ingress_rule_path | default('/') }}'
```
shanemcd: Please define a default variable in `defaults/main.yml` instead of using the `default` filter.

roles/installer/templates/tower_secret_key.yaml.j2
```diff
@@ -10,4 +10,4 @@ metadata:
     app.kubernetes.io/managed-by: awx-operator
     app.kubernetes.io/component: awx
 stringData:
-  secret_key: '{{ lookup('password', '/dev/null length=32 chars=ascii_letters,digits') }}'
+  secret_key: '{{ tower_custom_secret_key | lookup('password', '/dev/null length=32 chars=ascii_letters,digits') }}'
```
shanemcd: This seems like it would cause an error if `tower_custom_secret_key` is not set?

shanemcd (changes requested):
Please see my feedback in the comments above.

PR #245, opened 2021-04-24
Allow the user to set the ingress apiVersion

### LINKED ISSUE #205: minikube v1.19.0 issues with ingress-nginx validation webhook.
Minikube v1.19.0 updated ingress-nginx to v0.44.0. For what ever reason (still trying to figure it out), I get the following error while following logs on the awx-operator container:
```
(item=tower_ingress) => {\"ansible_loop_var\": \"item\", \"changed\": false, \"error\": 500, \"item\": \"tower_ingress\", \"msg\": \"Failed to apply object: b'{\\\"kind\\\":\\\"Status\\\",\\\"apiVersion\\\":\\\"v1\\\",\\\"metadata\\\":{},\\\"status\\\":\\\"Failure\\\",\\\"message\\\":\\\"Internal error occurred: failed calling webhook \\\\\\\\\\\"validate.nginx.ingress.kubernetes.io\\\\\\\\\\\": an error on the server (\\\\\\\\\\\"\\\\\\\\\\\") has prevented the request from succeeding\\\",\\\"reason\\\":\\\"InternalError\\\",\\\"details\\\":{\\\"causes\\\":[{\\\"message\\\":\\\"failed calling webhook \\\\\\\\\\\"validate.nginx.ingress.kubernetes.io\\\\\\\\\\\": an error on the server (\\\\\\\\\\\"\\\\\\\\\\\") has prevented the request from succeeding\\\"}]},\\\"code\\\":500}\\\\n'\", \"reason\": \"Internal Server Error\", \"status\": 500}
```

This may be related to: https://github.com/kubernetes/ingress-nginx/issues/5401

I was able to work-around the issue by running:
```
$ kubectl get validatingwebhookconfigurations
$ kubectl delete -A ValidatingWebhookConfiguration [configuration-name]
```
where in my case with minikube, the `[configuration-name]` is `ingress-nginx-admission`.

If I move back to minikube v1.18.1, everything works fine with both 0.8.0 and devel of awx-operator.

To reproduce (macos):
```
$ minikube config set vm-driver hyperkit
$ minikube start --cpus=4 --memory=8192
$ minikube addons configure ingress
$ minikube addons enable ingress
$ minikube addons enable metrics-server
$ minikube addons enable dashboard
$ kubectl apply -f https://raw.githubusercontent.com/ansible/awx-operator/devel/deploy/awx-operator.yaml
$ kubectl apply -f awx-deploy.yml
```
awx-deploy.yml:
```yaml
---
apiVersion: v1
kind: Namespace
metadata:
  name: awx
  labels:
    name: awx
---
kind: AWX
metadata:
  name: awx
  namespace: awx
  labels:
    name: awx
spec:
  tower_ingress_type: Ingress
  tower_hostname: awx-service.awx.svc.cluster.local
---
apiVersion: v1
kind: Secret
metadata:
  name: awx-admin-password
  namespace: awx
  labels:
    name: awx
stringData:
  password: changeme
```

Works: Minikube v1.18.1
Fails: Minikube v1.19.0

bhundven:
In #205, newer ingress is broken by not using networking.k8s.io/v1,
but this may not be the case for all users.

Allow apiVersion to be set by the user.

Default to networking.k8s.io/v1, but add a note in the README.md about
older k8s or ingress-nginx to use networking.k8s.io/v1beta1.

Signed-off-by: Bryan Hundven <bryanhundven@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,7 @@ spec:
 spec:
   ...
   ingress_type: Ingress
+  ingress_apiversion: networking.k8s.io/v1
   hostname: awx.mycompany.com
 ```
 
@@ -184,10 +185,13 @@ The following variables are customizable to specify the TLS termination procedur
 
 The following variables are customizable to specify the TLS termination procedure when `Ingress` is picked as an Ingress
 
-| Name                       | Description                              | Default       |
-| -------------------------- | ---------------------------------------- | ------------- |
-| ingress_annotations        | Ingress annotations                      | Empty string  |
-| ingress_tls_secret         | Secret that contains the TLS information | Empty string  |
+| Name                       | Description                              | Default             |
+| -------------------------- | ---------------------------------------- | ------------------- |
+| ingress_annotations        | Ingress annotations                      | Empty string        |
+| ingress_apiversion         | API Version for the Ingress              | extensions/v1beta1  |
+| ingress_tls_secret         | Secret that contains the TLS information | Empty string        |
+
+The default ingress_apiversion is `extensions/v1beta1`, but if you have a newer nginx-ingress deployment in k8s, you may need to change this to `networking.k8s.io/v1beta1`.
 
   * LoadBalancer
 
--- a/ansible/templates/crd.yml.j2
+++ b/ansible/templates/crd.yml.j2
@@ -86,6 +86,9 @@ spec:
                 ingress_annotations:
                   description: Annotations to add to the ingress
                   type: string
+                ingress_apiversion:
+                  description: apiVersion to use for the ingress
+                  type: string
                 ingress_tls_secret:
                   description: Secret where the ingress TLS secret can be found
                   type: string
--- a/deploy/awx-operator.yaml
+++ b/deploy/awx-operator.yaml
@@ -88,6 +88,9 @@ spec:
                 ingress_annotations:
                   description: Annotations to add to the ingress
                   type: string
+                ingress_apiversion:
+                  description: apiVersion to use for the ingress
+                  type: string
                 ingress_tls_secret:
                   description: Secret where the ingress TLS secret can be found
                   type: string
--- a/deploy/crds/awx_v1beta1_crd.yaml
+++ b/deploy/crds/awx_v1beta1_crd.yaml
@@ -86,6 +86,9 @@ spec:
                 ingress_annotations:
                   description: Annotations to add to the ingress
                   type: string
+                ingress_apiversion:
+                  description: apiVersion to use for the ingress
+                  type: string
                 ingress_tls_secret:
                   description: Secret where the ingress TLS secret can be found
                   type: string
--- a/deploy/olm-catalog/awx-operator/manifests/awx-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/awx-operator/manifests/awx-operator.clusterserviceversion.yaml
@@ -194,6 +194,11 @@ spec:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:io.kubernetes:Secret
         - urn:alm:descriptor:com.tectonic.ui:fieldDependency:ingress_type:Ingress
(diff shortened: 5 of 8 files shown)


bhundven:
@shanemcd lmk if you need any other changes. I'm happy to update this quickly.

shanemcd:
Unfortunately I don't think it will be so simple. When I tried to use this, I saw:


```
failed: [localhost] (item=tower_ingress) => changed=false
  ansible_loop_var: item
  error: 422
  item: tower_ingress
  msg: 'Failed to apply object: b''{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Ingress.extensions \\"awx-ingress\\" is invalid: spec.rules[0].http.paths[0].pathType: Required value: pathType must be specified","reason":"Invalid","details":{"name":"awx-ingress","group":"extensions","kind":"Ingress","causes":[{"reason":"FieldValueRequired","message":"Required value: pathType must be specified","field":"spec.rules[0].http.paths[0].pathType"}]},"code":422}\n'''
  reason: Unprocessable Entity
  status: 422
``` 

Per the k8s docs, it needs to look like this now:


```
spec:
  rules:
    - host: '{{ tower_hostname }}'
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: '{{ meta.name }}-service'
                port:
                  number: 80
```

At which point it will *only* work for `networking.k8s.io/v1`, so there's really no point to make it configurable.

shanemcd:
I should note that even when I made the above changes I still see the webhook error mentioned in https://github.com/ansible/awx-operator/issues/205

bhundven:
Well, we can update the template to switch the spec based on which apiVersion is set, but yea... we will have to wait for the next version of minikube.

bhundven:
I also removed the 'Closes' from the commit message and PR so it doesn't close 205

shanemcd (approved):
Nice work 😄

tchellomello (changes requested):
@bhundven thanks for great work.

I would say that would be nice to have this option also exposed as a `hidden` field at https://github.com/ansible/awx-operator/blob/devel/deploy/olm-catalog/awx-operator/manifests/awx-operator.clusterserviceversion.yaml so users can use configure it via UI as well. 

Do you mind adding it for future use? You might need to extend the `awxs.awx.ansible.com` CDR as well. 

But I'm ok if @shanemcd wants to merge like this since we can revisit this if needed in the future.

bhundven:
I will do that this evening.

bhundven:
@tchellomello Like that?

bhundven:
~~I can't see why test-local molecule scenario is failing. Could I get some help understanding what is going wrong?~~ @shanemcd I think I fixed this later with the fourth commit: Add missing type for apiversion to awx_v1beta1_crd

bhundven:
Forgot to add `type: string` to tower_ingress_apiversion in awx_v1beta1_crd.yml. Also rebased against latest devel.
@tchellomello Would you review?

bhundven:
I dropped the commit to try and fix the molecule int to string conversion warning. I think it broke the pipeline.

bhundven:
I am not sure why this is looping and never coming up? Is there something in molecule that needs to be updated?

bhundven:
I ran the test locally, and got this during prepare:
```
TASK [Fetch the kubeconfig] ****************************************************
fatal: [kind-test-local]: FAILED! => {"changed": false, "file": "/root/.kube/config", "msg": "unable to calculate the checksum of the remote file"}
```

bhundven:
That was fixed with a rebase. Now getting `Failed to find exact match for networking.k8s.io/v1.Ingress by [kind, name, singularName, shortNames]` when set to `extensions/v1beta1`, and `Failed to find exact match for .Ingress by [kind, name, singularName, shortNames]` when set to `networking.k8s.io/v1`. So I'm trying to sort that out now.

bhundven:
Ok, I fixed that, this passed on my test-local setup.

PR #247, opened 2021-04-25
Support custom config

kwangil-ha:
- Support rule path setting in ingress
- Support custom secret key setting in secret
- Support run as user and affinity setting in deployment


shanemcd (changes requested):
It seems like none of my feedback from https://github.com/ansible/awx-operator/pull/241 has been addressed.

PR #268, opened 2021-05-03
Update ingress template to remove k8s warning

endriu0:
Before this commit : Warning in k8s when checking ingress resource 
`Warning: extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress`

After this commit: Warning no longer appears and some future proofing in place


shanemcd:
Changing the API version alone won't be enough. See https://github.com/ansible/awx-operator/pull/245

PR #279, opened 2021-05-04
Persist secret names from old deployment & add them to the spec

rooftopcellist:
## Summary

  - renamed some more variables to be consistent with the pulp-operator
  - removed unneeded vars from backup & restore crds
  - added a way to parse spec at restore time by including vars to
    get around the issue of triply nested quotes when using to_json

### Why?

1. Creating the secrets with default names in the new namespace is undesirable because the AWX CR spec will still include names to custom secrets the user specified.  So the new deployment ends up with a spec that doesn't match the state of the secrets being used, which would be a problem for future backups.  

2. After persisting the secret names to the new AWX CR spec, another problem arose, the secrets that didn't have a custom secret are now not set in the spec.  So we need to extend the spec.  See verbose explanation below:

At backup time, the secret gets named: `{{ old_tower_name }}-secret-key`, and that gets set as a var and written to the backup pvc.  Then at restore time, that secret gets created as {{ old_tower_name }}-secret-key and the {{ new_tower_name}} CR gets created.  The installer for that looks for an existing {{ new_tower_name }}-secret-key secret and doesn't find it, so it generates one by the name `{{ old_tower_name }}-secret-key`.  This means we end up with duplicate secrets with different values, which is confusing to the end user and not expected.  

image

We don't know the name of the new deployment at the time of backup, so we are forced into passing the values into the spec vars.  We do this by combining the a dict of the secret names with the spec - [implemented here](https://github.com/rooftopcellist/awx-operator/pull/2/files#diff-15e71b7e97709f3557a1547aec4a94ddd121952ccc1b741308b45161ba8370d0R3-R34).

### DIFF
--- a/roles/restore/tasks/deploy_awx.yml
+++ b/roles/restore/tasks/deploy_awx.yml
@@ -8,9 +8,30 @@
       bash -c "cat '{{ backup_dir }}/awx_object'"
   register: awx_object
 
-- name: Set AWX spec variable from backup
+- name: Create temp file for spec dict
+  tempfile:
+    state: file
+  register: tmp_spec
+
+- name: Write spec vars to temp file
+  shell: |
+    echo '{{ awx_object.stdout }} > {{ tmp_spec.path }}'
+
+- include_vars: "{{ tmp_spec.path }}"
+  register: spec
+
+- name: Use include_vars to read in spec as a dict (because spec doesn't have quotes)
   set_fact:
-    awx_spec: "{{ awx_object.stdout }}"
+    awx_spec: "{{ spec.ansible_facts }}"
+
+- name: Set names of backed up secrets in the CR spec
+  set_fact:
+    awx_spec: "{{ awx_spec | combine ({ item.key : item.value }) }}"
+  with_items:
+    - {'key': 'tower_secret_key_secret', 'value': '{{ secret_key_secret_name }}'}
+    - {'key': 'tower_admin_password_secret', 'value': '{{ admin_password_secret_name }}'}
+    - {'key': 'tower_broadcast_websocket_secret', 'value': '{{ broadcast_websocket_secret_name }}'}
+    - {'key': 'tower_postgres_configuration_secret', 'value': '{{ postgres_configuration_secret_name }}'}
 
 - name: Deploy AWX
   k8s:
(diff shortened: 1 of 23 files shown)


roles/restore/tasks/deploy_awx.yml
```diff
@@ -8,9 +8,30 @@
       bash -c "cat '{{ backup_dir }}/awx_object'"
   register: awx_object
 
-- name: Set AWX spec variable from backup
+- name: Create temp file for spec dict
+  tempfile:
+    state: file
+  register: tmp_spec
```
rooftopcellist, line 14: If we could use `| to_json` when writing the spec to the backup, this business of creating a temp file, writing to it, then including vars from it could be avoided.  When [writing the file](https://github.com/ansible/awx-operator/pull/133/files#diff-3ced427495f8e83eeeefcf8bf0228eb0c2b7f0398758d64a46b49954d05528bcR24), we can't use `to_json` because the quotes in the resulting diction will be lost, so ansible treats it as an AnsibleUnsafeText type.  

If we could find a way to keep the quotes, we could just do:
```
- set_fact:
    awx_spec: "{{ lookup('file', 'awx_object') | from_json }}"
```

rooftopcellist: The feedback loop is slow, so here is a quick reproducer: https://gist.github.com/rooftopcellist/42a48b6d59488182176f8f5faa3fef0e

And here is my fixed version of that reproducer (which is used in the PR currently): https://gist.github.com/rooftopcellist/2943599dd731d5bfe4d160b93d37274e

roles/restore/tasks/deploy_awx.yml
```diff
@@ -8,9 +8,32 @@
...
+
+- name: Write spec vars to temp file
+  shell: |
+    echo '{{ awx_object.stdout }} > {{ tmp_spec.path }}'
+  changed_when: true
```
rooftopcellist: This failed the linter ([here](https://github.com/ansible/awx-operator/pull/279/checks?check_run_id=2501146267)).  Because this only writes to a tempfile which will be cleaned up when the process running the ansible playbook ends, I think this is OK.  If there is a better way to do this, please let me know.

rooftopcellist: I was able to remove this changed_when flag and make this a copy task.

PR #295, opened 2021-05-07
Scale down the new deployment before restoring

rooftopcellist:
Corrects a typo causing the awx-operator to never scale down the new awx deployment before restoring to it's db.  In it's current state, this task is be skipped (unless you happen to name the awxrestore object the same as your awx object.  

I also made it so that the `tower_image_pull_policy` variable applies to the postgres image as well, but am happy to remove it if people don't find this useful.

### DIFF
--- a/roles/installer/templates/tower_postgres.yaml.j2
+++ b/roles/installer/templates/tower_postgres.yaml.j2
@@ -30,6 +30,7 @@ spec:
     spec:
       containers:
         - image: '{{ tower_postgres_image }}:{{ tower_postgres_image_version }}'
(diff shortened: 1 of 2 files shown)


rooftopcellist:
I am no longer seeing this error during the restore role when scaling down the deployment:

```
"Failure\",\"message\":\"awxs.awx.ansible.com \\\\\"restore1\\\\\" not found\",\"reason\":\"NotFound\",\"details\":{\"name\":\"restore1\",\"group\":\"awx.ansible.com\",\"kind\":\"awxs\"},\"code\":404}
```

shanemcd (approved):

PR #296, opened 2021-05-07
Use custom archive format when migrating data

rooftopcellist:
This approach is compatible with the RH postgresql container.


rooftopcellist:
I just tested this, it still works with the default image as well:

image

I specified the pg creds to my old instance in the old-postgres-configuration secret, then deployed AWX and the data was properly migrated.  

image

shanemcd (approved):

Issue #291, opened 2021-05-07
How to specify NodePort?


VladoPortos:
I'm not sure right now how to specify NodePort with AWX operator.  Its not mentioned in documentation.

spec:
  tower_ingress_type: NodePort

Ok, but how to avoid random port assignment,

Usually I can do in service deployment something like:

```
spec:
  type: NodePort
  ports:
    - port: 80
      nodePort: 30080
      name: http
    - port: 443
      nodePort: 30443
      name: https
```

tchellomello:
@VladoPortos currently there is no support to manually specify the `NodePort` via operator (see https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/tower_service.yaml.j2#L48-L49)

However, that should be fairly simple to add (https://kubernetes.io/docs/concepts/services-networking/service/#nodeport). 

@VladoPortos do want to work on PR?

VladoPortos:
Hello @tchellomello I might have a look but need to finish some additional work first. I got to this while I was working on ansible role that deploys Minikube ( with docker backend ) and then AWX  which I did and it works fine :) just the HTTPS was point of interest. The classical ingress in kubernetes and minikube is more complicated than it needs to be and most people are just used to more to nginx reverse proxy. I was thinking to deploy AWX with tower_ingress_type: none and than just apply custom SVC with nodeport specified. than it can be passed further to role that setup nginx reverse proxy, without parsing the minikube get url command :)

VladoPortos:
Quick fix without me doing PR right now.

Deploy with:
`tower_ingress_type: NodePort` as normal ( or with none ) 

When AWX is deployed create file `awx_svc.yml`

```
---
apiVersion: v1
kind: Service
metadata:
  labels:
    app.kubernetes.io/component: awx
    app.kubernetes.io/managed-by: awx-operator
    app.kubernetes.io/name: awx
    app.kubernetes.io/part-of: awx
  name: awx-service-custom
  namespace: default
spec:
  ports:
    - name: http
      nodePort: 30444
      port: 80
      protocol: TCP
      targetPort: 8052
  selector:
    app.kubernetes.io/component: awx
    app.kubernetes.io/managed-by: awx-operator
    app.kubernetes.io/name: awx
  sessionAffinity: None
  type: NodePort
```

Change nodePort: to number you like in range: 30000-32767

And apply with:
minikube kubectl -- apply -f awx_svc.yml

This will make new service called awx-service-custom binded to port you have chosen. 

[...]

PR #310, opened 2021-05-16
Add configurable NodePort port number

jw-maynard:
My unknowledgeable stab at closing #291
I'm not sure if there are other parts of the operator that I need to make updates in that I'm missing.


jw-maynard:
Going to update this and re-submit.

Issue #304, opened 2021-05-11
Can't set '' as storageClassName value on pvcs


Zokormazo:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

As an openshift operator, sometimes I could want to bind my pvc to a pv without any storage class. 

To be able to do this on a cluster with a default sotrage class I should set `storageClassName: ''` on the pvc spec: 

https://docs.openshift.com/container-platform/4.7/storage/understanding-persistent-storage.html#pvc-storage-class_understanding-persistent-storage
> When a default storage class is configured, the PVC must explicitly ask for StorageClass or storageClassName annotations set to "" to be bound to a PV without a storage class.
>

But the operator will not set an storageClassName if the respective var's value is `''`:
https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/tower_postgres.yaml.j2#L97 
https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/tower_persistent.yaml.j2#L18
https://github.com/ansible/awx-operator/blob/devel/roles/backup/templates/backup_pvc.yml.j2#L10

So there is now way to set the explicit '' to bound the PVC to a PV without a storage class

##### ENVIRONMENT
* AWX version: 19.1.0
* Operator version: devel
* Kubernetes version:  OCP 4.7
* AWX install method: operator

##### STEPS TO REPRODUCE

try to set '' as storage_class for any of the storage_class vars

##### EXPECTED RESULTS

The PVC has `storageClassName: ''` on the spec

##### ACTUAL RESULTS

There is not `storageClassName` on the pvc spec

tchellomello:
Closing this as it was fixed by #315 and #318

PR #318, opened 2021-05-17
Default to storage class being undefined

rooftopcellist:
Follow-up for this PR: https://github.com/ansible/awx-operator/pull/315

Issue: https://github.com/ansible/awx-operator/issues/304

  * This is so that users can intentially set it to an empty string if they want to use the default storage class
  * conversely, now users can manually create a pvc that does not utilize the default storage class


tchellomello (approved):
:+1:  LGTM

PR #319, opened 2021-05-17
Use custom pg_dump format for faster restores

rooftopcellist:
The custom format for pg_dump is much more performant for large databases and should be used instead of the text format.  This is already the case for the `migrate_data.yml` logic, this PR updates the backup and restore roles to similarly use the custom format.


tchellomello (approved):
:+1:

PR #299, opened 2021-05-07
Unset ownerRefs in the installer instead of the finalizer

rooftopcellist:
Fixes: #228

Currently, if you deploy AWX to a project, then delete that project, the termination of the AWX object hangs.  

This is because openshift has already deleted the secrets for the AWX deployment (as per the project deletion), but the AWX finalizer is still trying to set the ownerReferences on those secrets to `null` even though they don't exist.  

The whole point of this logic in the finalizer is to only garbage collect the secrets when intended based on the value of `tower_garbage_collect_secrets`.  By default, the ownerReference is set to the object that created the secret (injected at the time of creation).  Setting these ownerReference's to `null` at the end of the installer instead of in the finalizer avoids this conflict with project deletions.

PR #323, opened 2021-05-18
Fully remove finalizer

rooftopcellist:
This is a follow-up to this PR: https://github.com/ansible/awx-operator/pull/299

I left in an empty finalizer directory because it errored without it.  @Zokormazo pointed out that it was failing because it was still referenced in the watches.yml.  This PR removes it completely.


Zokormazo (approved):

tchellomello (approved):
:+1:

PR #331, opened 2021-05-23
Do not prepend variables name with tower_

Spredzy:

PR #339, opened 2021-05-27
Update README.md

ivaanko:
Hi!

Thanks a lot for your work!
I found some discrepancies between README and actual CRD.

Cheers,
Ivan

### DIFF
--- a/README.md
+++ b/README.md
@@ -485,17 +485,17 @@ Example spec file for volumes and volume mounts
 ---
     spec:
     ...
-      ee_extra_volume_mounts: |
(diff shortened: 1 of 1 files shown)


tchellomello:
@ivaanko thanks for your PR, however the changes you pointed out were implemented by the PR https://github.com/ansible/awx-operator/pull/331. 

If you check the `devel` then you will have it.  The new changes will be present in the next release. 

If you are using the `0.9.0` version, then you can follow the docs on that branch at https://github.com/ansible/awx-operator/tree/0.9.0

PR #348, opened 2021-05-31
Run redis service as redis user

aperigault:
Hi,

by default redis container launch redis as root. It is a problem on kubernetes clusters with restrictive psp. Redis container is built with a redis user inside so I propose to use it.


shanemcd:
We'll want to run this through some tests before it can be considered. I mostly want to make sure it doesn't break on OCP, given it has a more restrictive security policy enabled by default.

mo-saeed:
Hope this get merged, we have same issue.

mo-saeed:
@shanemcd Any updates on this ?

Spredzy:
@rooftopcellist Can you give that a spin and see if it works on OCP please ?

mo-saeed:
@Spredzy Unfortunately I have no OCP setup to test this. Could you please elaborate why this shouldn't work in OCP or any other platform, this is to run redis as non-root user.

I have changed the deployment manually on EKS cluster and it works fine.

mo-saeed:
@rooftopcellist @Spredzy Any updates here ? this is really blocking us right now from deploying awx, would appreciate if you can move this forward somehow. Thanks

gigawhat:
I ran into this issue as well. Our default psp doesn't allow running containers as root. In the mean time I've switched to the bitnami redis container. This container does not run as root.

mhahl:
In OpenShift, When a Pod is deployed to a project, by default, a unique UID is allocated and used to execute the Pod,
setting the `securityContext` in this case seems to break the Deployment unfortunately.

```---
kind: AWX
metadata:
  name: awx-demo
spec:
  service_type: ClusterIP
  ingress_type: route
  hostname: awx.apps.ocp.hahl.id.au
```

image

metashock:
@Wolskie Thank you for checking this!

I'm running into the same problem. Can't we extend the AWX crd and add an optional security_context property to it and render the securityContext only when this new field is not blank? Openshift users would just leave it blank, other users could set it.

shanemcd:
>  Can't we extend the AWX crd and add an optional security_context property

Yes, would accept a patch that does this conditionally.

shanemcd:
Closing in favor of https://github.com/ansible/awx-operator/pull/638

Issue #345, opened 2021-05-31
Backup is broken


Zokormazo:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Backup is broken

##### ENVIRONMENT
* AWX version: 19.1
* Operator version: devel
* Kubernetes version: OCP 4.6
* AWX install method: operator

##### STEPS TO REPRODUCE

1. Install operator
2. Deploy an awx instance
3. Backup

##### EXPECTED RESULTS

Get a backup

##### ACTUAL RESULTS

Operator fails on https://github.com/ansible/awx-operator/blob/devel/roles/backup/tasks/postgres.yml#L30-L41

##### ADDITIONAL INFORMATION

AWX CR:
```yaml
---
apiVersion: awx.ansible.com/v1beta1
kind: AWX
metadata:
  name: tower
  namespace: backup
spec:
  admin_user: admin
  admin_email: test@example.com
  ingress_type: Route
  hostname: tower
  replicas: 1
```
AWXBackup CR:
```
---
apiVersion: awx.ansible.com/v1beta1
kind: AWXBackup
metadata:
  name: awxbackup1
  namespace: backup
spec:
  deployment_name: tower
```

##### AWX-OPERATOR LOGS
[awx-operator.log](https://github.com/ansible/awx-operator/files/6568722/awx-operator.log)

Spredzy:
Fixed by https://github.com/ansible/awx-operator/pull/354

PR #297, opened 2021-05-07
Make postgres sts labels consistent with k8s recommendations & pulp-operator

rooftopcellist:
This PR updates the postgres labels to conform with the recommendations in the kubernetes docs and to remain consistent with the pulp-operator.  
  - k8s recommended labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
  - [pulp-operator labels](https://github.com/pulp/pulp-operator/blob/31a60ca80f3e5484da1687d1d232e536e54eaa5e/roles/postgres/templates/postgres.yaml.j2#L7-L23)

PR #354, opened 2021-06-01
Update name of default postgres label selector for backup role

rooftopcellist:
Issue: https://github.com/ansible/awx-operator/issues/345

As mentioned by @Zokormazo , the backup is broken because the postgres pod is never found.  This is because the default value for the postgres_label_selector var isn't set correctly.  This value changed in this PR - https://github.com/ansible/awx-operator/pull/297/files, but this instance was missed.


Zokormazo:
Verified, this fixes #345

Spredzy (approved):

PR #358, opened 2021-06-02
nginx: Improve security configuration

Spredzy:


roles/installer/templates/config.yaml.j2
```diff
@@ -214,6 +217,13 @@ data:
                 {%- endif %}
                 proxy_set_header X-Forwarded-Port 443;
                 uwsgi_param HTTP_X_FORWARDED_PORT 443;
+
+                add_header Strict-Transport-Security max-age=15768000;
```
tchellomello, line 221: :+1: It's important to keep this like here because once you override the `add_header` in a block you must declare all options again. @Spredzy++

tchellomello (approved):
:+1:  LGTM @Spredzy
Thanks!

Zokormazo (approved):

PR #361, opened 2021-06-02
Use own attribute for control_plane_ee

Zokormazo:
Signed-off-by: Julen Landa Alustiza <jlanda@redhat.com>

### DIFF
--- a/ansible/templates/crd.yml.j2
+++ b/ansible/templates/crd.yml.j2
@@ -149,6 +149,9 @@ spec:
                         type: string
                       image:
                         type: string
+                control_plane_ee_image:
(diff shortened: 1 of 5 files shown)


kdelee (commented):
How does this work when we need to have a credential to pull it, is there some place else to specify the ImagePullSecret ?

Zokormazo:
> How does this work when we need to have a credential to pull it, is there some place else to specify the ImagePullSecret ?

yes. `image_pull_secret` is set as `ImagePullSecret` on pod spec so you can use that to specify your pull secrets for the control plane ee.

Note that it will not be converted to a controller docker registry credential and assigned to the execution environment, it will just allow to pull the controlplane embedded ee

PR #353, opened 2021-06-01
olm-catalog: Add missing statuses

Spredzy:

PR #362, opened 2021-06-02
olm-catalog: Add statuses to schemas for awxbackup and awxrestore objects

rooftopcellist:
This PR adds statuses to the olm-catalog and crd schema's.  It also adds in the Telemetry statuses.

### DIFF
--- a/deploy/olm-catalog/awx-operator/manifests/awx.ansible.com_awxs_crd.yaml
+++ b/deploy/olm-catalog/awx-operator/manifests/awx.ansible.com_awxs_crd.yaml
@@ -385,15 +385,6 @@ spec:
               image:
                 description: URL of the image used for the deployed instance
                 type: string
-              postgresConfigurationSecret:
-                description: Postgres Configuration secret name of the deployed instance
-                type: string
-              broadcastWebsocketSecret:
-                description: Broadcast websocket secret name of the deployed instance
-                type: string
-              secretKeySecret:
-                description: Secret key secret name of the deployed instance
-                type: string
               migratedFromSecret:
                 description: The secret used for migrating an old instance.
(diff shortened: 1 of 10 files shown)


shanemcd:
Please don't update the OLM files by hand. See https://github.com/ansible/awx-operator/pull/364

deploy/olm-catalog/awx-operator/manifests/awx.ansible.com_awxs_crd.yaml
```diff
@@ -385,15 +385,6 @@ spec:
...
                 type: string
-              postgresConfigurationSecret:
-                description: Postgres Configuration secret name of the deployed instance
-                type: string
-              broadcastWebsocketSecret:
-                description: Broadcast websocket secret name of the deployed instance
-                type: string
-              secretKeySecret:
-                description: Secret key secret name of the deployed instance
-                type: string
```
rooftopcellist, line 396: This is getting deleted because it is now duplicated.  It was added by @Spredzy [here](https://github.com/ansible/awx-operator/pull/353/files), then added again when the bundle was generated [here](https://github.com/ansible/awx-operator/pull/364)

rooftopcellist:
I updated the templated/generated files using:

```bash
$ ansible-playbook ansible/chain-operator-files.yml
$ operator-sdk generate bundle --operator-name awx-operator --version 0.10.0
```

shanemcd (changes requested):
We are on version `0.10.0`, not `10.0.0` 😅

rooftopcellist:
_phew_ thanks for catching that typo @shanemcd!

shanemcd (approved):

PR #364, opened 2021-06-02
Update OLM metadata

shanemcd:
I generated this by running: 

```
$ operator-sdk generate bundle --operator-name awx-operator --version 0.0.1
```


tchellomello (approved):
WOW!! Thanks for sharing this!! I was modifying those files by hand and totally missed this nifty shortcut!!

 @shanemcd++ @rooftopcellist++

Issue #320, opened 2021-05-18
AWXBackup should capture pg_dump result and report in case of error


tchellomello:
cc: @rooftopcellist 

##### ISSUE TYPE
 - Bug Report

##### SUMMARY
AWXBackup should capture pg_dump result and report in case of error
##### ENVIRONMENT

* AWX version: X.Y.Z
* Operator version: devel
* Kubernetes version: 1.20.7
* AWX install method: k8s

##### STEPS TO REPRODUCE

`awx-operator`  should be more verbose when a `pg_dump` fails. (**ignore the version mismatch error. This is just to force a pg_dump to fail**)

https://github.com/ansible/awx-operator/blob/devel/roles/backup/tasks/postgres.yml#L89-L95

##### EXPECTED RESULTS

Should we raise an event? Log message? Debug + Pause? Should we update the status when checking the `awxbackup` kind?

##### ACTUAL RESULTS

`AWXBackup` finishes without error even when it fails
```yaml
status:
  backupClaim: awx-toca-backup-claim
  backupDirectory: /backups/tower-openshift-backup-2021-05-18-01:46:01
  conditions:
  - ansibleResult:
      changed: 0
      completion: 2021-05-18T01:47:04.616452
      failures: 0
      ok: 4
      skipped: 7
    lastTransitionTime: "2021-05-18T01:45:47Z"
    message: Awaiting next reconciliation
    reason: Successful
    status: "True"
    type: Running
```

##### ADDITIONAL INFORMATION

```yaml
kubectl get awxbackup -o yaml awx-toca-backup-051721a                                                                         21:53:21
apiVersion: awx.ansible.com/v1beta1
kind: AWXBackup
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"awx.ansible.com/v1beta1","kind":"AWXBackup","metadata":{"annotations":{},"name":"awx-toca-backup-051721a","namespace":"default"},"spec":{"deployment_name":"awx-toca"}}
  creationTimestamp: "2021-05-18T01:45:47Z"
  generation: 2
  labels:
    app.kubernetes.io/component: awx
    app.kubernetes.io/managed-by: awx-operator
    app.kubernetes.io/name: awx-toca-backup-051721a
    app.kubernetes.io/operator-version: operator_labels_complementary
    app.kubernetes.io/part-of: awx-toca-backup-051721a
  managedFields:
  - apiVersion: awx.ansible.com/v1beta1
[...]

PR #372, opened 2021-06-04
Fail task if pg_dump fails in backup role

rooftopcellist:
Issue: https://github.com/ansible/awx-operator/issues/320

Previously, if the `pg_dump` exec task failed in the backup role, it would fail silently and keep running.  

3e3c6b0e96fc4a18a8bc5af215de7067

> The output above was from running the backup role directly as a playbook with the status tasks commented out

Now, it will fail:

image

Similarly, if I deploy the operator and create a backup object, it will set the Failed status condition.

```
  status:
    conditions:
    - lastTransitionTime: "2021-06-04T15:15:17Z"
      reason: Failed
      status: "False"
      type: Failure
```

#### Additional Info

As part of this PR, I made `postgres_image` and `postgres_image_version` configurable, which made testing this a bit easier.  It also servce the use case of a user who has an external postgres database they manage, but would still like to use our backup/restore roles.

### DIFF
--- a/ansible/templates/awxbackup_crd.yml.j2
+++ b/ansible/templates/awxbackup_crd.yml.j2
@@ -46,6 +46,12 @@ spec:
                 postgres_label_selector:
                   description: Label selector used to identify postgres pod for backing up data
                   type: string
+                postgres_image:
+                  description: Registry path to the PostgreSQL container to use
+                  type: string
+                postgres_image_version:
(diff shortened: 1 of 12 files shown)


tchellomello (approved):
Awesome @rooftopcellist **thank you**

LGTM!

rooftopcellist:
Thanks for taking a look @tchellomello !

PR #374, opened 2021-06-06
- Added ability to declare namespace via a variable

mkimberley:
- Avoids creating role binding in default namespace
- Avoids creating service account in default namespace

Signed-off-by: Matt Kimberley <matt.kimberley@hotmail.com>

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -2,3 +2,4 @@ operator_image: quay.io/ansible/awx-operator
 operator_version: 0.10.0
 pull_policy: Always
 ansible_debug_logs: "false"
+awx_namespace: "awx"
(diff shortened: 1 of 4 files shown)


mkimberley:
Operator still creates pods in the default namespace under kubernetes, which is different behaviour to that seen in openshift with this change

mkimberley:
Operator still creates pods in the default namespace under kubernetes, which is different behaviour to that seen in openshift with this change

cnukwas:
@mkimberley , came across this as I am looking to install AWX Operator CR/CRDs and a single instance in a given namespace, rather than installing at cluster level. Based on above comment, this would work as expected in OpenShift, is that right?
Also from which version is this feature available as I was looking at 0.12 due to known LDAP integration issues with 0.13 version of the operator?

PR #356, opened 2021-06-02
Use awx-manage register_default_execution_environments

Zokormazo:
Signed-off-by: Julen Landa Alustiza <jlanda@redhat.com>

PR #379, opened 2021-06-07
Register ees

kdelee:
This is https://github.com/ansible/awx-operator/pull/356 but rebased on devel + with new change to use https://github.com/ansible/awx/pull/10377

### DIFF
--- a/ansible/templates/crd.yml.j2
+++ b/ansible/templates/crd.yml.j2
@@ -151,6 +151,8 @@ spec:
                         type: string
                 control_plane_ee_image:
                   description: Registry path to the Execution Environment container image to use on control plane pods
+                ee_pull_credentials_secret:
+                  description: Secret where pull credentials for registered ees can be found
                   type: string
                 image_pull_policy:
                   description: The image pull policy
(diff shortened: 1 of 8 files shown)


Zokormazo:
Seems good to me. Rerunned the pr check

deploy/crds/awx_v1beta1_crd.yaml
```diff
@@ -151,6 +151,8 @@ spec:
                         type: string
                 control_plane_ee_image:
                   description: Registry path to the Execution Environment container image to use on control plane pods
```
Zokormazo, line 153: missing type here

Zokormazo (dismissed):

deploy/crds/awx_v1beta1_crd.yaml
kdelee: what is type

kdelee: oh string I guess

ansible/templates/crd.yml.j2
```diff
@@ -151,6 +151,8 @@ spec:
                         type: string
                 control_plane_ee_image:
                   description: Registry path to the Execution Environment container image to use on control plane pods
```
kdelee, line 153: @Zokormazo so I think when I did the rebase I messed this up. all these need to have a type, right?

PR #384, opened 2021-06-09
Add script to generate files after crd.yml.j2 changes

rooftopcellist:
This script may be useful for other people making changes to template files.  This will help keep people from editing the generated files manually when variables/attributes are added.  

cc @tchellomello

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-## This script will build 3 images awx-{operator,bundle,catalog}
+## This script will generate a bundle manifest, build 3 images awx-{operator,bundle,catalog}
 ## and push to the $REGISTRY specified.
 ##
 ## The goal is provide an quick way to build a test image.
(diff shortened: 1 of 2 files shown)


rooftopcellist:
I also added a step top generate the bundle manifest to the `build.sh`.

shanemcd (changes requested):
-1 on adding a the shell script at the root of the repo. Please put it in `scripts`.

rooftopcellist:
woops, I meant to put that in the scripts directory, thanks for catching that, it is updated.

scripts/build.sh
```diff
@@ -78,6 +78,7 @@ build_operator_image() {
 
 build_bundle_image() {
   echo "Building and pushing $BUNDLE_IMAGE image"
+  operator-sdk generate bundle --operator-name awx-operator --version $TAG
```
tchellomello, line 81: That is great so we can have the option to generate the bundle here as well!

tchellomello (approved):
:fire: sweet!!

scripts/build.sh
rooftopcellist: One less step when testing changes that affect the bundle :smile:

PR #401, opened 2021-06-15
Added ability to trust a custom bundle CA

### LINKED ISSUE #376: Add private CAs to the containers
Hi,
to allow access to TLS sites using private CAs we need to add CA certificates to the AWX containers. Is there a native way to do so using AWX operator?

I used extra_volumes/ee_extra_volume_mounts to get crt files to /etc/pki/ca-trust/source/anchors/ but a run of update-ca-trust is missing. Is there a native way to accomplish adding own CAs?

Currently we got two use cases for that:
1. fetching projects from GIT
2. using the "uri" module in roles
thanx in advance
Max

tchellomello:
Fixes: https://github.com/ansible/awx-operator/issues/376

This PR introduces the `initContainer`  to prepare the RHEL PKI for the sub-containers used in AWX. This approach allows having the `update-trust-ca`  executed one instead for each container via `lifecycle`.  

See the https://github.com/ansible/awx-operator/issues/376 for the full testing report

Creating secret:
```sh
$ cat Toca_ROOT_CA.crt  Toca_Intermediate_CA.crt  > /tmp/bundle-ca.crt
$ kubectl create secret generic  awx-ssl-ca-custom  --from-file=bundle-ca.crt=/tmp/bundle-ca.crt
```

Modifying `awx` spec:

```yaml
kind: AWX
....
spec:
  bundle_cacert_secret: awx-ssl-ca-custom
....
```

Result:
```sh
(py39) mdemello@storm ~> kubectl iexec awx /bin/bash                                                                                                                                                                                     00:53:40
Namespace: default | Pod: ✔ awx-ssl-ca-6cccf6577d-jzrk9
Container: ✔ awx-ssl-ca-task
bash-4.4$ ls -la /etc/pki/ca-trust/source/anchors/bundle-ca.crt 
-rw-r--r--. 1 root root 4086 Jun 11 04:51 /etc/pki/ca-trust/source/anchors/bundle-ca.crt
```

### DIFF
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -111,6 +111,8 @@ redis_image: docker.io/redis
 redis_image_version: latest
 postgres_image: postgres
 postgres_image_version: 12
+init_container_image: quay.io/centos/centos
+init_container_image_version: 8
 image_pull_policy: IfNotPresent
 image_pull_secret: ''
 
@@ -205,6 +207,9 @@ ca_trust_bundle: "/etc/pki/tls/certs/ca-bundle.crt"
 # Secret to lookup that provides the LDAP CACert trusted bundle
 #
 ldap_cacert_secret: ''
+
+# Secret to lookup that provides the custom CA trusted bundle
+bundle_cacert_secret: ''
 
 # Whether secrets should be garbage collected
 # on teardown
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ An [Ansible AWX](https://github.com/ansible/awx) operator for Kubernetes built w
          * [Deploying a specific version of AWX](#deploying-a-specific-version-of-awx)
          * [Privileged Tasks](#privileged-tasks)
          * [Containers Resource Requirements](#containers-resource-requirements)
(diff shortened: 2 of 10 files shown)


roles/installer/defaults/main.yml
```diff
@@ -111,6 +111,8 @@ redis_image: docker.io/redis
 redis_image_version: latest
 postgres_image: postgres
 postgres_image_version: 12
+init_container_image: quay.io/centos/centos
```
rooftopcellist, line 114: I agree with the approach of this PR, and it works, but is using this image OK? @wenottingham @Spredzy

wenottingham: I would assume in other places this would be overriden. @tchellomello ?

tchellomello: Yes @wenottingham. The variables `init_container_image` and `init_container_image_version` can be overriden.

rooftopcellist (approved):
After rebasing, this looks good to me.  Nice work @tchellomello !

shanemcd (approved):

PR #404, opened 2021-06-16
backup: Ensure colon are preserved in awx spec

### LINKED ISSUE #370: Restore fails if a value in the deployment has colon
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
In awx object spec I have `ingress_annotations: 'cert-manager.io/cluster-issuer: letsencrypt'`, the value that has a colon. In such a case the awx object does not serialize correctly and restore fails with "unknown playbook failure".

##### ENVIRONMENT
* AWX version: 19.2.0
* Operator version: 0.10.0
* Kubernetes version: v1.21.0+k3s1
* AWX install method: kubernetes

##### STEPS TO REPRODUCE

Backup:

```
---
kind: AWXBackup
metadata:
  name: awxbackup-manual-2021-06-04-01
spec:
  deployment_name: awx
  postgres_label_selector: app.kubernetes.io/instance=postgres-awx
  backup_pvc: manual-claim
```
Restore:

```
---
kind: AWXRestore
metadata:
  name: awxrestore-init
spec:
  deployment_name: awx
  backup_pvc: manual-claim
  backup_dir: '/backups/tower-openshift-backup-2021-06-04-11:42:15'
  backup_pvc_namespace: default
```

##### EXPECTED RESULTS

Restore should succeed

##### ACTUAL RESULTS

Restore fails with "unknown playbook failure".

##### ADDITIONAL INFORMATION

This is what persisted `awx_object` looks like:

```
{admin_user: admin, api_version: awx.ansible.com/v1beta1, create_preload_data: True, deployment_type: awx, garbage_collect_secrets: False, hostname: nevermind.domain.tld, image_pull_policy: IfNotPresent, ingress_annotations: cert-manager.io/cluster-issuer: letsencrypt, ingress_tls_secret: awx-le-certificate, ingress_type: Ingress, kind: AWX, loadbalancer_port: 80, loadbalancer_protocol: http, projects_persistence: False, projects_storage_access_mode: ReadWriteMany, projects_storage_size: 8Gi, replicas: 1, route_tls_termination_mechanism: Edge, task_privileged: False}
``` 
Note how there is no quotes around `cert-manager.io/cluster-issuer: letsencrypt`.  The reason for this is that [this line](https://github.com/ansible/awx-operator/blob/af5983cb6879cbb2b653af142d9a65d7f7d96117/roles/backup/tasks/awx-cro.yml#L24) does not quite work as the author intended.

Suggested fix:

```
bash -c 'echo "$0" > {{ backup_dir }}/awx_object' {{ awx_spec|quote }}
```

##### AWX-OPERATOR LOGS

Can be found here: <https://gist.github.com/AndrewSav/e9e73d9b4ab19341bf4926707ec52540>

Spredzy:
Due to the way the command is run, the field with a colon in it will
generate a broken yaml serialization. This commit aims to keep it
accurate.

Proof of working as expected:

```
---
- hosts: localhost
  gather_facts: false
  vars:
    awx_spec:
      foo: bar
      tmp: 'foo: bar'
      task_extra_env: |
        - name: http_proxy
          value: "http://proxy.domain.tld:3128"
  tasks:
    - debug:
        var: '{{ awx_spec | to_nice_yaml | quote }}'

    - name: Store spec definition
      # command: bash -c "echo '{{ awx_spec }}' > awx_object"
      command: bash -c 'echo "$0" > awx_object' {{ awx_spec | quote  }}

    - name: Load spec definition
      command: bash -c "cat awx_object"
      register: awx_object

    - name: Create temp file for spec dict
      tempfile:
        state: file
      register: tmp_spec

    - name: Write spec vars to temp file
      copy:
        content: "{{ awx_object.stdout }}"
        dest: "{{ tmp_spec.path }}"
        mode: '0644'

    - name: Include spec vars to save them as a dict
      include_vars: "{{ tmp_spec.path }}"
      register: spec

    - name: Use include_vars to read in spec as a dict (because spec doesn't have quotes)
      set_fact:
        awx_spec: "{{ spec.ansible_facts }}"

    - debug:
        var: awx_spec
```

With the fix on (above snipped) output is:

```
TASK [debug] ***********************************************************************************************************************************************************************************
ok: [localhost] => 
  awx_spec:
    foo: bar
    task_extra_env: |-
      - name: http_proxy
        value: "http://proxy.domain.tld:3128"
    tmp: 'foo: bar'
```

Without it, the playbook run would crash

```
TASK [Include spec vars to save them as a dict] ************************************************************************************************************************************************
fatal: [localhost]: FAILED! => changed=false 
  ansible_facts: {}
  ansible_included_var_files: []
  message: |-
    We were unable to read either as JSON nor YAML, these are the errors we got from each:
    JSON: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
  
    Syntax Error while loading YAML.
      did not find expected ',' or '}'
  
    The error appears to be in '/tmp/ansible.ea5z_u_z': line 1, column 20, but may
  
    {foo: bar, tmp: foo: bar, task_extra_env: - name: http_proxyn value: http://proxy.domain.tld:3128n}
                       ^ here
```

Fixes: https://github.com/ansible/awx-operator/issues/370

### DIFF
--- a/roles/backup/tasks/awx-cro.yml
+++ b/roles/backup/tasks/awx-cro.yml
@@ -21,4 +21,4 @@
     namespace: "{{ backup_pvc_namespace }}"
     pod: "{{ meta.name }}-db-management"
     command: >-
-      bash -c "echo '{{ awx_spec }}' > {{ backup_dir }}/awx_object"
+      bash -c 'echo "$0" > {{ backup_dir  }}/awx_object' {{ awx_spec | quote  }}


Zokormazo (approved):

rooftopcellist (approved):

rooftopcellist:
This works for me for both issues brought up in the ticket, thanks for the fix!

PR #409, opened 2021-06-17
AWX Operator 0.11.0

shanemcd:


deploy/awx-operator.yaml
```diff
@@ -711,7 +711,7 @@ spec:
       serviceAccountName: awx-operator
       containers:
         - name: awx-operator
-          image: "quay.io/ansible/awx-operator:0.10.0"
+          image: "quay.io/shanemcd/awx-operator:0.11.0"
```
Zokormazo: sure? :)

shanemcd: You are too fast.... I fixed almost immediately 😄

PR #430, opened 2021-06-24
Backup ingress, bundle & ee pull secrets

rooftopcellist:
# Summary

Follow-up from conversation here: https://github.com/ansible/aap-qa/pull/502

The new `ee_pull_credentials_secret` was not on the list of secrets to be backed up.  I also noticed we should be backing up the bundle cacert and the ingress tls secret.  

This PR adds the following secrets to the list:

    - ingress_tls_secret
    - bundle_cacert_secret
    - ee_pull_credentials_secret

This is the commit that introduced the ee pull secret - https://github.com/ansible/awx-operator/pull/379/files

### DIFF
--- a/roles/backup/tasks/secrets.yml
+++ b/roles/backup/tasks/secrets.yml
@@ -18,8 +18,11 @@
   include_tasks: dump_secret.yml
   loop:
     - route_tls_secret
+    - ingress_tls_secret
     - ldap_cacert_secret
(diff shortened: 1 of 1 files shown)


Zokormazo (approved):

rooftopcellist:
Waiting for tests to pass, then I will merge this.

PR #435, opened 2021-06-28
Preserve newlines in AWX spec

rooftopcellist:
Issue & context: https://github.com/ansible/awx-operator/issues/370#issuecomment-868851991

This will preserve newlines in awx_spec vars.  Below is an example that was problematic that this fixes:

```
  spec:
    admin_email: admin@localhost
    admin_user: admin
    create_preload_data: true
    development_mode: false
    extra_volumes: |
      - name: myvolume
        persistentVolumeClaim:
          claimName: manual-claim
```

Previously, this would end up looking like this in the resulting awx object's spec:

```
    extra_volumes: '- name: myvolume\n  persistentVolumeClaim:\n    claimName: manual-claim\n'
```

### DIFF
--- a/roles/backup/tasks/awx-cro.yml
+++ b/roles/backup/tasks/awx-cro.yml
@@ -10,11 +10,8 @@
 
 - name: Set AWX object
   set_fact:
-    _awx: "{{ _awx_cro['resources'][0] }}"
-
-- name: Set user specified spec
-  set_fact:
-    awx_spec: "{{ _awx['spec'] }}"
+    awx_spec:
+      spec: "{{ this_awx['resources'][0]['spec'] }}"
 
 - name: Set names of backed up secrets in the CR spec
   set_fact:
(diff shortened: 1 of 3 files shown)


rooftopcellist:
I tested this out by:
1. Deploying the awx-operator with these changes
2. Deploying AWX by creating a CR object
3. Backing up AWX by creating a CR awxbackup object
4. Restoring AWX by creating a CR awxrestore object

The spec of the resulting AWX CR object is formatted correctly:

```
[chadams@chadams-work awx-operator]$ oc -n ca-dev get awx -o yaml | grep ' spec:' -A 8
  spec:
    admin_email: admin@localhost
    admin_user: admin
    create_preload_data: true
    development_mode: false
    extra_volumes: |
      - name: myvolume
        persistentVolumeClaim:
          claimName: manual-claim
```

PR #441, opened 2021-06-30
Fix var so that generated secret names are added to spec backup

rooftopcellist:
Issue: https://github.com/ansible/tower/issues/5185

The generated secret names were not getting added to the awx_spec at backup time.  As a result, at restore time, the postgres configuration secret was not known to the AWX object.  

After the fix:

```
root@debug-db-management:/backups/tower-openshift-backup-2021-06-30-09:50:30# ls
awx_object  secrets.yml  tower.db
root@debug-db-management:/backups/tower-openshift-backup-2021-06-30-09:50:30# cat awx_object 
spec: {admin_email: admin@localhost, admin_password_secret: awx-admin-password, admin_user: admin,
  broadcast_websocket_secret: awx-broadcast-websocket, create_preload_data: true,
  development_mode: false, garbage_collect_secrets: false, image_pull_policy: Always,
  ingress_type: Route, loadbalancer_port: 80, loadbalancer_protocol: http, postgres_configuration_secret: awx-postgres-configuration,
  projects_persistence: false, projects_storage_access_mode: ReadWriteMany, projects_storage_size: 8Gi,
  replicas: 1, route_tls_termination_mechanism: Edge, secret_key_secret: awx-secret-key,
  service_type: clusterip, task_privileged: false}
```

### DIFF
--- a/roles/restore/tasks/postgres.yml
+++ b/roles/restore/tasks/postgres.yml
@@ -1,5 +1,9 @@
 ---
 
+- name: Set Postgres Configuration Secret name
+  set_fact:
+    postgres_configuration_secret: "{{ spec['postgres_configuration_secret'] | default(postgres_configuration_secret) }}"
+
 - name: Check for specified PostgreSQL configuration
   k8s_info:
     kind: Secret
--- a/roles/backup/tasks/awx-cro.yml
+++ b/roles/backup/tasks/awx-cro.yml
@@ -10,17 +10,21 @@
 
 - name: Set AWX object
   set_fact:
-    awx_spec:
-      spec: "{{ this_awx['resources'][0]['spec'] }}"
+    _awx: "{{ this_awx['resources'][0]['spec'] }}"
 
 - name: Set names of backed up secrets in the CR spec
   set_fact:
-    awx_spec: "{{ awx_spec | combine ({ item.key : item.value }) }}"
+    _awx: "{{ _awx | combine ({ item.key : item.value }) }}"
   with_items:
     - {"key": "secret_key_secret", "value": "{{ this_awx['resources'][0]['status']['secretKeySecret'] }}"}
     - {"key": "admin_password_secret", "value": "{{ this_awx['resources'][0]['status']['adminPasswordSecret'] }}"}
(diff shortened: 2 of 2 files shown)


Zokormazo (approved):

tchellomello (approved):
:+1:

roles/restore/tasks/postgres.yml
```diff
@@ -1,5 +1,9 @@
 ---
 
+- name: Set Postgres Configuration Secret name
+  set_fact:
+    postgres_configuration_secret: "{{ spec['postgres_configuration_secret'] | default(postgres_configuration_secret) }}"
+
```
rooftopcellist, line 6: This default was getting used instead of the one set in the awx spec - https://github.com/ansible/awx-operator/blob/devel/roles/restore/vars/main.yml#L10-L14

I only left those defaults in in case users specified `backup_dir` and `backup_pvc` instead of just the `backup_name` for the CR.

Solution, we get the postgres-configuration-secret-name from the backup's awx_object (spec).  A valid backup will always have the pg secret name in the spec.  

```
TASK [restore : debug] ***********************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": "dev-restore-postgres-configuration"
}
TASK [restore : Set Postgres Configuration Secret name] **************************************************************************************************************************************************************************************
ok: [localhost]
TASK [restore : debug] ***********************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": "awx-postgres-configuration"
}
```

rooftopcellist:
It looks like that just uncovered the real issue:

```
 TASK [Check for pending migrations] ******************************** 
[0;31mfatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to execute on pod dev-restore-7bdc557b56-fw6wp due to : (0)\nReason: Handshake status 500 Internal Server Error\n"}
```

```
django.db.utils.OperationalError: FATAL:  password authentication failed for user "awx"

```

rooftopcellist:
It looks like the 500 I was seeing was an old awx-operator hanging out.

PR #456, opened 2021-07-06
Update README.md

nntrn:
Add demonstration for running kubectl inside minikube and add step to create alias


shanemcd (approved):
🤘 thanks @nntrn!

PR #460, opened 2021-07-07
Set no_log on all tasks that could leak secrets in logs

rooftopcellist:
Related to https://github.com/pulp/pulp-operator/pull/160/files

This adds `no_log: true` to all tasks that could potentially leak secret values or database credentials.


Spredzy (approved):

rooftopcellist:
I manually deployed awx-operator, awx, then did a backup & restore as a sanity check.  I didn't run into any issues, and could log in to the resulting deployment and run a job successfully.

PR #464, opened 2021-07-12
Define ingress path as variable for deployments

### LINKED ISSUE #455: Ingress Path / Causes 404 when running eks + alb in aws
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
The default ingress config uses / for the path, when running in AWS/EKS using the aws-load-balancer-controller, it adds a listener with a rule:

IF
Host is awx.domain
Path is /

Any requests to not / paths (such as those fetching static resources) end in a 404.

##### ENVIRONMENT
* AWX version: 19.2.2
* Operator version: 0.12.0
* Kubernetes version: 1.19.x
* AWX install method: operator on eks

##### STEPS TO REPRODUCE

Provision awx into the following ingress annotations:

```yaml
kind: AWX
metadata:
  name: awx
  namespace: awx
spec:
  # Base Configuration
  replicas: 1
  # Connction Info
  hostname: <host>
  ingress_type: Ingress
  service_type: ClusterIP
  ingress_annotations: |
    kubernetes.io/ingress.class: alb
    alb.ingress.kubernetes.io/tags: Name=k8s-awx-ingress
    alb.ingress.kubernetes.io/inbound-cidrs: "<cidr>"
    alb.ingress.kubernetes.io/group.name: "k8s-awx-ingress"
    alb.ingress.kubernetes.io/scheme: internal
    alb.ingress.kubernetes.io/target-type: 'ip'
    alb.ingress.kubernetes.io/actions.ssl-redirect: '{"Type": "redirect", "RedirectConfig":{ "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}'
    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS":443}]'
    service.beta.kubernetes.io/aws-load-balancer-ssl-cert: <arn>
    service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: '60'
    service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: 'true'
    service.beta.kubernetes.io/aws-load-balancer-ssl-negotiation-policy: ELBSecurityPolicy-TLS-1-2-2017-01
```

##### EXPECTED RESULTS

AWX Deploys and is accessible via URL

##### ACTUAL RESULTS

AWX resolves by URL by all static content fails to load with 404s

##### ADDITIONAL INFORMATION

Solution:
Removing / from the ingress removes this rule from the ALB and allows awx to function.
Idea would be to add the ingress path as a configurable option, defaulting to / which would allow us in this case to use an empty string which removes the alb rule (verified manually)

##### AWX-OPERATOR LOGS
N/A

jitterjuice:
This PR provides the ingress path on ingress deployment as a variable that defaults to '/'. By overriding this value, custom ingress can be applied for running on cloud-providers using external LoadBalancers.

For the issue, please see: https://github.com/ansible/awx-operator/issues/455.

Fixes: https://github.com/ansible/awx-operator/issues/455

Updates the defaults for the installer role, the crd's and the documentation. Tested build against local Kubernetes by updating the ingress_path to be '' and '/'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,7 @@ The following variables are customizable when `ingress_type=ingress`. The `ingre
 | ingress_annotations        | Ingress annotations                      | Empty string                 |
 | ingress_tls_secret         | Secret that contains the TLS information | Empty string                 |
 | hostname                   | Define the FQDN                          | {{ meta.name }}.example.com  |
+| ingress_path               | Define the ingress path to the service   | /                            |
 
 ```yaml
 ---
--- a/ansible/templates/crd.yml.j2
+++ b/ansible/templates/crd.yml.j2
@@ -86,6 +86,9 @@ spec:
                     - ingress
                     - Route
                     - route
+                ingress_path:
+                  description: The ingress path used to reach the deployed service
+                  type: string
                 ingress_annotations:
                   description: Annotations to add to the Ingress Controller
                   type: string
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -6,9 +6,9 @@ LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=awx-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
 LABEL operators.operatorframework.io.bundle.channel.default.v1=alpha
-LABEL operators.operatorframework.io.metrics.project_layout=ansible
 LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
 LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v0.19.4
+LABEL operators.operatorframework.io.metrics.project_layout=ansible
 
 COPY deploy/olm-catalog/awx-operator/manifests /manifests/
 COPY deploy/olm-catalog/awx-operator/metadata /metadata/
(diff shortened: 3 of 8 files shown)


MrBones757:
The operator has been built locally and deployed as an in-place upgrade to 0.12 - testing addition of this path and param update succeeded on an existing awx cluster

tchellomello (changes requested):
Thanks so much for your contribution @js-rwwa. I think you just need to address a small thing. 

I can see you manually edited the files because the file `ansible/templates/crd.yml.j2` is not included on your PR. This `ansible/templates/crd.yml.j2` is the source file we use to generate the remaining files to the operator. 

You can use this script `https://github.com/ansible/awx-operator/blob/devel/scripts/generate-files.sh` to generate the changes for you. 

Once this gets addressed, looks good to be merged! 

Thank you!

jitterjuice:
My apologies @tchellomello , I did miss that in the readme. I have updated ansible/templates/crd.yml.j2 and run the generate-files with TAG=0.12.0.

Could you please review the files to see if anything else needs changing? 

Thank you!

tchellomello (approved):
Thanks for fixing it @js-rwwa

It looks good to me.  @shanemcd  @rooftopcellist  @Spredzy  can we have another set of eyes on it?

rooftopcellist (approved):
lgtm, thanks for the contribution @js-rwwa!

PR #490, opened 2021-08-08
Update API versions removed from kubes 1.22

smuth4:
The ingress change will lock out versions <1.19, if desired I can add a `ingress_api_version` or something similar.

### DIFF
--- a/ansible/templates/role.yml.j2
+++ b/ansible/templates/role.yml.j2
@@ -37,6 +37,11 @@ rules:
       - daemonsets
       - replicasets
       - statefulsets
+    verbs:
+      - '*'
+  - apiGroups:
+      - networking.k8s.io
+    resources:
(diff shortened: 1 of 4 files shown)


rtsp:
`deploy.yaml`  is actually generated from templates in `ansible/templates/`.

You'll need to update `ansible/templates/role.yml.j2` like you did with `ClusterRole` in `awx-operator.yaml`.

https://github.com/rtsp/awx-operator/blob/k8s-v1.22-api-removed/ansible/templates/role.yml.j2#L42-L47

smuth4:
Thanks, fixed.

Spredzy:
@smuth4 I see https://github.com/ansible/awx-operator/pull/508 got merged fixing the same issue. Would you consider it enough for closing this one ?

PR #495, opened 2021-08-11
use awx-ee:latest

kdelee:
We are updating the requirements in awx to get the latest receptor and runner in the task container,
we should also have the latest in the EE

see https://github.com/ansible/awx/pull/10861 and https://github.com/ansible/awx/pull/10858

### DIFF
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -134,9 +134,9 @@ init_container_extra_volume_mounts: ''
 
 ee_images:
   - name: AWX EE 0.5.0
-    image: quay.io/ansible/awx-ee:0.5.0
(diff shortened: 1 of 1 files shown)


roles/installer/defaults/main.yml
```diff
@@ -134,9 +134,9 @@ init_container_extra_volume_mounts: ''
 
 ee_images:
   - name: AWX EE 0.5.0
```
simaishi: Update this too?

kdelee: thanks

kdelee:
```suggestion
  - name: AWX EE (latest)
```

rooftopcellist (approved):

PR #485, opened 2021-07-31
pgsql uses initContainer to address FS permissions

tchellomello:
When using `local-path-provisioner` driver it might occur of the `postgresql` user to be unable to write on the `pvc` mounted at `{{ postgres_data_path }}`. 

To address this, we can use a `initContainer` which will be called only the first time the PostgreSQL statefulset gets created to address the directory to be writable by the `postgresql` user. 

This approach worked well when using the `postgres:12 ` image, however, we need to test using the `rhel8/postgresql-12` as well. 

### TODO
- [ ] - Test using the `rhel8/postgresql-12` image on OCP *cc*: @rooftopcellist 
- [x] - Test using a `k3s` with `local-path-provisioner`  with `postgres:12` image 

Fixes: #483
Fixes: #475

```sh
$ kubectl get pvc                                                                                                                                                00:45:39
NAME                      STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
awx-projects-claim        Bound    pvc-33b08fb9-2af1-4352-b805-49c9686ddfbc   2Gi        RWO            local-path     27m
postgres-awx-postgres-0   Bound    pvc-786afed3-6a14-4bc0-9c54-e39015b481f2   3Gi        RWO            local-path     20m


$ kubectl get pods -w                              00:38:58
NAME                            READY   STATUS    RESTARTS   AGE
awx-operator-5bc776b4d4-d9ww2   1/1     Running   0          4m41s
awx-postgres-0                  1/1     Running   0          4m3s
awx-d67898cd9-k6jrc             4/4     Running   0          3m48s

$ kubectl iexec awx-postgres-0 /bin/bash 
root@awx-postgres-0:/# namei  -xmolv /var/lib/postgresql/data/pgdata/
f: /var/lib/postgresql/data/pgdata/
Drwxr-xr-x root     root     /
drwxr-xr-x root     root     var
drwxr-xr-x root     root     lib
drwxr-xr-x postgres postgres postgresql
Drwx------ postgres root     data
drwx------ postgres root     pgdata

PR #499, opened 2021-08-12
Revert initContainer change to address FS permissions issues

rooftopcellist:
I merged this pre-maturely.  Unfortunately it will not work with the Red Hat postgresql image as well in it's current state.   Let's handle this in the next release.  

Reverts: https://github.com/ansible/awx-operator/pull/485


shanemcd (approved):

tchellomello:
Thanks for reverting it @rooftopcellist. Let's review the plan next week to better address this problem.

PR #512, opened 2021-08-27
Add clarification for inexperienced users

thorian93:
We hit that issue ourselves and it took us weeks until someone pointed out our error.
As it feels like this could hit other users inexperienced with kubernetes too, I propose this note.


tchellomello (approved):
Thanks @thorian93

alexbmw00:
Hi Guys,

I got solved a problem with recreate a credentials in Ansible AWX (Credentials entry) , perhaps a bug after upgrade operator.

"AWX 2.2.1 provided by Ansible"

PR #523, opened 2021-09-05
Add Uninstall section

eoq:
Just a quick instruction on how to remove an AWX deployment from a new user who struggled with it for a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -771,7 +771,7 @@ $ kubectl delete awx awx-demo
 awx.awx.ansible.com "awx-demo" deleted
 ```
 
-Deleting an AWX instance will remove all related deployments and statefulsets, however, persistent volumes and secrets will remain. To enforce secrets also getting removed, you can use garbage_collect_secrets: true.
+Deleting an AWX instance will remove all related deployments and statefulsets, however, persistent volumes and secrets will remain. To enforce secrets also getting removed, you can use `garbage_collect_secrets: true`.
 
 ### Upgrading
 
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ An [Ansible AWX](https://github.com/ansible/awx) operator for Kubernetes built w
          * [Exporting Environment Variables to Containers](#exporting-environment-variables-to-containers)
          * [Extra Settings](#extra-settings)
          * [Service Account](#service-account)
+      * [Uninstall](#uninstall)
    * [Upgrading](#upgrading)
    * [Contributing](#contributing)
    * [Release Process](#release-process)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -760,6 +761,17 @@ Example configuration of environment variables
       eks.amazonaws.com/role-arn: arn:aws:iam::<ACCOUNT_ID>:role/<IAM_ROLE_NAME>
 ```
 
+
+### Uninstall ###
+
+To uninstall the AWX deployment you need to direct AWX Operator to initiate the removal which will delete all AWX objects except for persistent volumes.
```
tchellomello: Well, that is not completely true if the user has the `garbage_collect_secrets: true` on the SPEC. So maybe we could rephrase this statement for:

_To uninstall an AWX deployment instance, you basically need to remove the `AWX` kind related to that instance. For example, to delete a `awx` instance named `awx-demo`, you would do:_

```bash
$ kubectl delete awx awx-demo
awx.awx.ansible.com "awx-demo" deleted
```

_Deleting an `AWX` instance will remove all deployments and statefulsets related, however, `persistence volumes `and `secrets` will remain. To enforce `secrets` get also removed, you can use the `garbage_collect_secrets: true`._

tchellomello (changes requested):
Thanks for your contribution @eoq 

I added some notes on your PR. I don't want to be too prescriptive. You can rephrase the text if you want, however,  let's make sure a similar message is there. 

Thank you

README.md
eoq: Great. I will update it per your advise.

README.md
```diff
@@ -760,6 +761,18 @@ Example configuration of environment variables
...
+
+To uninstall an AWX deployment instance, you basically need to remove the AWX kind related to that instance. For example, to delete an AWX instance named awx-demo, you would do:
+
+```bash
+$ kubectl delete awx awx-demo
+awx.awx.ansible.com "awx-demo" deleted
+```
+
+Deleting an AWX instance will remove all related deployments and statefulsets, however, persistent volumes and secrets will remain. To enforce secrets also getting removed, you can use garbage_collect_secrets: true.
+
```
tchellomello, line 775: Could you please add the `garbage_collect_secrets: true` between single-quote?

tchellomello (changes requested):

tchellomello (approved):
Thanks for working on this @eoq 

We really appreciate your assistance with the project!

**cc**: @rooftopcellist let's see what @rooftopcellist  says about and then we can merge it!

rooftopcellist (approved):
Thank you for contributing this to the docs @eoq !   These notes look good to me.

PR #505, opened 2021-08-20
Add redis_resource_requirements

beenje:
In our OKD cluster all containers must have quotas.

The Redis container currently doesn't have resources configured.
I added the `redis_resource_requirements` variable.

PR #527, opened 2021-09-07
Added Redis Container resource requirements

rajatguptarg:
Currently, the Redis container does not have a variable to set the Kubernetes resource quota. This will add another variable in `redis_resource_requirements` in the spec section, which can be customized depending on the use case.

This will also enhance the Kubernetes cluster security, as the Redis container will also have a resource definition.


rajatguptarg:
Closing in favour of https://github.com/ansible/awx-operator/pull/505

PR #528, opened 2021-09-07
Running as user nobody if task_privilage is false

rajatguptarg:
If the `task_privileged` is `false`, then the container should not run as root, this will open security vulnerabilities in the Kubernetes cluster. If the user sets the value of `task_privileged` to `false`, then the container should run as user `nobody`. 

This will enhance the Kubernetes cluster security and the variable needs to set explicitly to `true`,  if the user wants to run awx as root.

### DIFF
--- a/roles/installer/templates/deployment.yaml.j2
+++ b/roles/installer/templates/deployment.yaml.j2
@@ -178,6 +178,9 @@ spec:
 {% if task_privileged == true %}
           securityContext:
             privileged: true
+{% else %}
+          securityContext:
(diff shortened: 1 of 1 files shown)


shanemcd:
This will not work on OpenShift as the default restricted scc will not allow you to specify runAsUser.

I'm also not sure why you need to do this since by default we run as user 1000: https://github.com/ansible/awx/blob/devel/tools/ansible/roles/dockerfile/templates/Dockerfile.j2#L292

PR #529, opened 2021-09-07
Mount receptor config in awx-task container

rooftopcellist:
When deploying an AWX app image from awx/devel with the awx-operator, the containers deployed and run, but after migrations run I am see this in the awx-task logs:

```
FileNotFoundError: [Errno 2] No such file or directory: '/etc/receptor/receptor.conf'
```

This PR mounts the receptor.conf where the awx app expects it.  Now, when deploying a fresh image built from awx/devel with the awx-operator, I no longer see the errors in the task container and can log in and run jobs.

### DIFF
--- a/roles/installer/templates/deployment.yaml.j2
+++ b/roles/installer/templates/deployment.yaml.j2
@@ -222,6 +222,10 @@ spec:
               mountPath: "/var/run/awx-rsyslog"
             - name: rsyslog-dir
               mountPath: "/var/lib/awx/rsyslog"
+            - name: "{{ meta.name }}-receptor-config"
+              mountPath: "/etc/receptor/receptor.conf"
+              subPath: receptor.conf
+              readOnly: true
             - name: receptor-socket
               mountPath: "/var/run/receptor"
             - name: "{{ meta.name }}-projects"


shanemcd (approved):

AlanCoding:
The only reason we read this is so that we can get the location of the receptor sockfile... somewhere like `/var/lib/awx-receptor/receptor.sock`, is this still all going to work? I'm not as familiar with the OpenShift AWX process tree as I should be. Does receptor run inside the same container? If it does, we're probably all fine :ok_hand:

roles/installer/templates/deployment.yaml.j2
```diff
@@ -222,6 +222,10 @@ spec:
               mountPath: "/var/run/awx-rsyslog"
             - name: rsyslog-dir
               mountPath: "/var/lib/awx/rsyslog"
+            - name: "{{ meta.name }}-receptor-config"
+              mountPath: "/etc/receptor/receptor.conf"
+              subPath: receptor.conf
+              readOnly: true
```
tchellomello, line 228: @rooftopcellist  I have not tested yet (I'll test it tomorrow), but `awx` code reads the `/etc/receptor/receptor.conf` to find the `socket` location to dispatch work (see https://github.com/ansible/awx/pull/10932) . 

Looking at the `configmap`, the socket will be available at `filename: /var/run/receptor/receptor.sock`  with perms `0660` 
```yaml
- control-service:
    service: control
    filename: /var/run/receptor/receptor.sock
    permissions: 0660
```

roles/installer/templates/deployment.yaml.j2
```diff
@@ -222,6 +222,10 @@ spec:
...
             - name: receptor-socket
               mountPath: "/var/run/receptor"
```
tchellomello, line 230: That socket gets created by the `ee` container (https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/deployment.yaml.j2#L265) however, it seems the same socket gets mounted on the `task` container which allows the `task` container to issue commands. 

@rooftopcellist  If you still have the environment, are you able to connect on the `task` container and then do:

```bash
ls -la /var/run/receptor/receptor.sock
nc -U /var/run/receptor/receptor.sock
status

OR
RECEPTORCTL_SOCKET=/var/run/receptor/receptor.sock receptorctl status
``` 

I think it will work as you mentioned you were able to run jobs. If that is the case, the PR looks good to me!!

tchellomello:
nit comment:
@rooftopcellist maybe one thing we might need to do in the future is to use  `/etc/receptor/receptor.conf` path in all containers instead of `/etc/receptor.conf`, especially because when the TLS stuff arrives. 

**see**:
https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/deployment.yaml.j2#L265 
https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/deployment.yaml.j2#L276

Zokormazo:
```
        - name: elijah-awx-receptor-config
          readOnly: true
          mountPath: /etc/receptor/receptor.conf
          subPath: receptor.conf
        - name: receptor-socket
          mountPath: /var/run/receptor
```

receptor.conf points to /var/run/receptor/receptor.sock and the socket is there

I can run jobs on this awx instance

Zokormazo (approved):

roles/installer/templates/deployment.yaml.j2
rooftopcellist: @tchellomello Yep, the .sock file is created where we expect it to be:

```
bash-4.4$ ls -la /var/run/receptor/receptor.sock
srw-rw----. 1 awx 1000670000 0 Sep  8 12:48 /var/run/receptor/receptor.sock

```

Receptorctl status:

```
bash-4.4$ source /var/lib/awx/venv/awx/bin/activate
(awx) bash-4.4$ RECEPTORCTL_SOCKET=/var/run/receptor/receptor.sock receptorctl status
Node ID: awx-59fd4465c6-zp9qn
Version: -0.git20210902.
System CPU Count: 4
System Memory MiB: 15013

Node         Service   Type       Last Seen             Tags
awx-59fd4465c6-zpbash-4.4$ ls -la /var/run/receptor/receptor.sock
srw-rw----. 1 awx 1000670000 0 Sep  8 12:48 /var/run/receptor/receptor.sock
9qn control   Stream     2021-09-08 12:54:12   {'type': 'Control Service'}

Node         Work Types
awx-59fd4465c6-zp9qn local, kubernetes-runtime-auth, kubernetes-incluster-auth
```

I wasn't able to do the `nc` command to connect because we don't have nc in the container.

rooftopcellist:
Re: https://github.com/ansible/awx-operator/pull/529#issuecomment-914898760, I see what you mean, it would be better to use /etc/receptor/receptor.conf everywhere (in this case in the awx-ee container). 

I'll merge this and handle that in a separate PR.  Thanks!

tchellomello (commented):
Awesome!! LGTM good job @rooftopcellist

PR #535, opened 2021-09-08
nodePort paramater missing when using service_type: nodeport

srinathman:
It seems that the "nodeport_port" is not actually setting nodeport to the desired static value, instead a random port is assigned. 
As per https://kubernetes.io/docs/concepts/services-networking/service/, a "nodePort" parameter must be provided for static nodeport_port, otherwise k8s will pick up a random port.
Just setting "port" and "targetPort" is not enough.

### DIFF
--- a/roles/installer/templates/service.yaml.j2
+++ b/roles/installer/templates/service.yaml.j2
@@ -44,6 +44,7 @@ spec:
       protocol: TCP
       targetPort: 8052
       name: http
+      nodePort: {{ nodeport_port }}
   type: NodePort
 {% endif %}
   selector:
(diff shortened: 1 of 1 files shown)


shanemcd:
After reading the docs you linked to... shouldn't `port: {{ nodeport_port }}` get changed to `port: 8052` ?

srinathman:
I wasn't sure why the original code was like, but yes, I believe it should be 8052

shanemcd:
Sorry for the delay here, this slipped through the cracks. I see there is a conflict now so I'm going to close this in favor of https://github.com/ansible/awx-operator/pull/789.

PR #541, opened 2021-09-14
Move to namespace-scoped operator for better security & isolation

rooftopcellist:
Be more intentional about what permissions the awx-operator service account has.  Restrict access where possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 This is a list of high-level changes for each release of `awx-operator`. A full list of commits can be found at `https://github.com/ansible/awx-operator/releases/tag/<version>`.
 
+# 0.14.0 (TBA)
+
+- Starting with awx-operator 0.14.0, AWX can only be deployed in the namespace that the operator exists in. See [upgrade docs](#upgrading) for necessary cleanup actions. (Christian Adams) - 58c3ebf (breaking change)
+
 # 0.10.0 (Jun 1, 2021)
 
 - Make tower_ingress_type to respect ClusterIP definition (Marcelo Moreira de Mello) - e37c091 (breaking_change)
--- a/README.md
+++ b/README.md
@@ -795,6 +795,11 @@ To upgrade AWX, it is recommended to upgrade the awx-operator to the version tha
 Apply the awx-operator.yml for that release to upgrade the operator, and in turn also upgrade your AWX deployment.
 
 
+**Cluster-scope to Namespace-scope considerations**
+
+Starting with awx-operator 0.14.0, AWX can only be deployed in the namespace that the operator exists in. This is called a namespace-scoped operator. If you are upgrading from an earlier version, you will want to 
+delete your existing `awx-operator` service account, role and role binding.
+
 ## Contributing
 
 Please visit [our contributing guidelines](https://github.com/ansible/awx-operator/blob/devel/CONTRIBUTING.md).
--- a/ansible/deploy-operator.yml
+++ b/ansible/deploy-operator.yml
@@ -5,7 +5,7 @@
 - name: Deploy Operator
   hosts: localhost
   vars:
-    k8s_namespace: "default"
+    k8s_namespace: "{{ namespace | default('default') }}"
     obliterate: no
 
   collections:
--- a/ansible/templates/operator.yml.j2
+++ b/ansible/templates/operator.yml.j2
@@ -22,9 +22,11 @@ spec:
             - mountPath: /tmp/ansible-operator/runner
               name: runner
           env:
-            # Watch all namespaces (cluster-scoped).
+            # Watch one namespace (namespace-scoped).
             - name: WATCH_NAMESPACE
-              value: ""
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
             - name: POD_NAME
               valueFrom:
                 fieldRef:
--- a/ansible/templates/role.yml.j2
+++ b/ansible/templates/role.yml.j2
@@ -11,7 +11,13 @@ rules:
       - routes
       - routes/custom-host
     verbs:
-      - '*'
+      - get
+      - list
+      - create
+      - delete
+      - patch
+      - update
+      - watch
   - apiGroups:
       - ""
       - "rbac.authorization.k8s.io"
@@ -28,7 +34,13 @@ rules:
       - roles
       - rolebindings
     verbs:
-      - '*'
+      - get
+      - list
+      - create
+      - delete
+      - patch
+      - update
+      - watch
   - apiGroups:
       - apps
       - networking.k8s.io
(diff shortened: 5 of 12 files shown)


rooftopcellist:
This PR has been updated to move the `awx-operator` from a **cluster-scoped operator** to a **namespace-scoped operator**.  At a high level this means only Roles for service accounts, no ClusterRoles, and it also means that operators & AWX deployments will have a 1-to-1 relationship.  

  * This means one operator can only deploy to a single namespace
  * This increases security, the awx-operator SA has less cluster-wide
    access
  * Safer upgrades: upgrade is sandboxed to a single namespace, so this will help with conflicts with other awx-operators (or operators in general) in a cluster.  
  * If AWX deployments are needed in multiple namespaces, multiple
    awx-operators can be deployed to accomplish this.

rooftopcellist:
Looks like CI is borked.  I am able to run `molecule test -s test-minikube` locally without issue, I tried with k8s 1.20, and 1.19.2 (what our CI uses).

```
TASK [Verify there is one AWX pod] *********************************************
ok: [localhost] => {
    "changed": false,
    "msg": "All assertions passed"
}

localhost                  : ok=21   changed=8    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0

INFO     Running test-minikube > destroy
WARNING  Skipping, instances are delegated.
INFO     Pruning extra files from scenario ephemeral directory
[chadams@chadams-work awx-operator]$ minikube version
minikube version: v1.17.1
```

rooftopcellist:
CI will be fixed by this commit, a rebase will be required once it merges - #552

Spredzy:
@rooftopcellist  Merged now :)

tchellomello (approved):
Beautiful work @rooftopcellist++ 

Quick question for further PR: How are we going to tackle the upgrade progress? We need to create a route to delete the older `ClusterRole` and `ClusterRoleBinding` objects, right?

tchellomello:
Talking with @rooftopcellist  he brought a good point. Because the `awx-operator` the new deployed instance will not have to the `ClusterRole*` objects, so we need to document this very well!!

rooftopcellist:
Thanks for the review @tchellomello, as you mentioned, the new operator will not have permission to be able to clean up the old service account, so manual user intervention is required.  I added some notes in the upgrade docs (https://github.com/ansible/awx-operator/pull/541/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R798-R802) and a changelog entry for our next release.

craph:
Hello @rooftopcellist , @tchellomello , @Spredzy 

For this PR is it possible to add some documentation to explain how to use it ? I mean, add an example to deploy awx-operator to a desired namespace instead of default ?

This will be very usefull for the community.

Thank you.
Best regards,

cnukwas:
Came across this post while looking for this feature, looking forward for the PR to be available in next release and related documentation.

ddevalco:
This PR reads as though the operator should not be using clusterroles anymore when doing an install, however even with me using the latest release 0.15.0 I encounter errors deploying this that refer to the cluster roles and cluster role binding.

Example:

`"system:authenticated" "system:cattle:authenticated"]) is attempting to grant RBAC permissions not currently held:
{APIGroups:["coordination.k8s.io"], Resources:["leases"], Verbs:["get" "list" "watch" "create" "update" "patch" "delete"]}
Error from server (Forbidden): error when retrieving current configuration of:
Resource: "rbac.authorization.k8s.io/v1, Resource=clusterroles", GroupVersionKind: "rbac.authorization.k8s.io/v1, Kind=ClusterRole"
Name: "awx-operator-metrics-reader", Namespace: ""
from server for: "STDIN": clusterroles.rbac.authorization.k8s.io "awx-operator-metrics-reader" is forbidden: User "u-w4453qrfpa" cannot get resource "clusterroles" in API group "rbac.authorization.k8s.io" at the cluster scope
Error from server (Forbidden): error when retrieving current configuration of:
Resource: "rbac.authorization.k8s.io/v1, Resource=clusterroles", GroupVersionKind: "rbac.authorization.k8s.io/v1, Kind=ClusterRole"
Name: "awx-operator-proxy-role", Namespace: ""
from server for: "STDIN": clusterroles.rbac.authorization.k8s.io "awx-operator-proxy-role" is forbidden: User "u-w4453qrfpa" cannot get resource "clusterroles" in API group "rbac.authorization.k8s.io" at the cluster scope
Error from server (NotFound): error when creating "STDIN": roles.rbac.authorization.k8s.io "awx-operator-awx-manager-role" not found
Error from server (NotFound): error when creating "STDIN": roles.rbac.authorization.k8s.io "awx-operator-leader-election-role" not found
Error from server (Forbidden): error when retrieving current configuration of:
Resource: "rbac.authorization.k8s.io/v1, Resource=clusterrolebindings", GroupVersionKind: "rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding"
Name: "awx-operator-proxy-rolebinding", Namespace: ""
from server for: "STDIN": clusterrolebindings.rbac.authorization.k8s.io "awx-operator-proxy-rolebinding" is forbidden: User "u-w4453qrfpa" cannot get resource "clusterrolebindings" in API group "rbac.authorization.k8s.io" at the cluster scope
make: *** [deploy] Error 1`

Is there something special you have to do in order to avoid the cluster role stuff because I can't seem to deploy in this situation?

blackluck:
Hello, I just started to read awx operator docs and configs so I could miss things, please help me understand how this should work.
This breaking change mentions "If AWX deployments are needed in multiple namespaces, multiple awx-operators can be deployed to accomplish this." but how will it work with the cluster scoped objects like crd, clusterrole etc (like it mentioned in #705 ) especially when upgrading an operator?

For example we need 2 awx install  in separate namspaces, there is ns1 and ns2, in both of them installed awx operator 0,15 (since 0.14 I need to install operator in both) and 1-1 awx instance.
Starting upgrade ns1 awx operator (maybe/probably) will change awx.ansible.com crds (because crd is cluster scoped object), so it could affect ns2 awx instance too. So "upgrade is sandboxed to a single namespace" isn't fully solved, right?

Crd upgrades are problems to other operators too for the same reason, for example some take crd install separately so if there are multiple operator in the cluster, those operator upgrades won't effect crd changes and crd change need to do in another task as cluster level modification (it still could effects all installed operator, just more controlled when).

I understand the motive to restrict operator access as much as needed, for example with operator it's used to be a nice option to have namespace list or namespace selector for crd resources watch (if you install a cluster scoped awx operator, but only want to watch in ns1 and ns2 the awx, awxrestore and awxbackup resources). But I didn't see any operator with crds really namespace scoped (yet).

Btw is there a way to still install awx operator cluster scoped (> 0.14 version)?
Because we want to install separate awx instances in multiple namespaces and can't really see yet how this operator install and more importantly the upgrades could work with multiple operators and I feel less problematic to install one cluster scoped operator?

PR #551, opened 2021-09-17
Update crd & awx-operator.yml templated files

rooftopcellist:
Update templated files.


rooftopcellist:
CI will be fixed by this commit, a rebase will be required once it merges - https://github.com/ansible/awx-operator/pull/552

PR #556, opened 2021-09-21
[Doc] Rework explanations for version selection

vrubiolo:
Use a shell snippet so that the commands are readily usable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,10 +106,11 @@ Let's create an alias for easier usage:
 $ alias kubectl="minikube kubectl --"
 ```
 
-Now you need to deploy AWX Operator into your cluster. Start by going to https://github.com/ansible/awx-operator/releases and making note of the latest release. Replace `<TAG>` in the URL `https://raw.githubusercontent.com/ansible/awx-operator/<TAG>/deploy/awx-operator.yaml` with the version you are deploying.
+Now you need to deploy the AWX Operator into your cluster. Start by going to the [release page](https://github.com/vrubiolo/awx-operator/releases) and make note of the latest release. Replace `$TAG` below with that version.
 
 ```bash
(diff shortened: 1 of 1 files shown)


vrubiolo:
This should smooth the experience a bit.

Ideally, I'd like to have a way to automatically insert the latest release version tag into the snippet in question but I don't know how to do that in _the documentation_ (links like [this one](https://stackoverflow.com/questions/18995854/how-can-i-use-github-api-to-get-all-tags-or-releases-for-a-project) show how to do it from the API itself).

README.md
```diff
@@ -106,10 +106,11 @@ Let's create an alias for easier usage:
 $ alias kubectl="minikube kubectl --"
 ```
 
-Now you need to deploy AWX Operator into your cluster. Start by going to https://github.com/ansible/awx-operator/releases and making note of the latest release. Replace `<TAG>` in the URL `https://raw.githubusercontent.com/ansible/awx-operator/<TAG>/deploy/awx-operator.yaml` with the version you are deploying.
+Now you need to deploy the AWX Operator into your cluster. Start by going to the [release page](https://github.com/vrubiolo/awx-operator/releases) and make note of the latest release. Replace `$TAG` below with that version.
```
rooftopcellist, line 109: @vrubiolo do you mind updated the link here to point to the ansible/awx-operator/releases page instead of your fork?

vrubiolo: Hi and thanks for your feedback.

Given the recent changes via https://github.com/ansible/awx-operator/commit/d4c8fd67f3a39b1a2c6f2166a3f2bc5958861c4b#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R104 and the move to a `kustomize`-based workflow, I think this PR is not as useful anymore: the version to target is done via the branc checkout of the git repo.

I would thus tend to think we can close it, unless you object. What do you think?

shanemcd:
Closing because this is now completely different after upgrading to operator-sdk 1.x.

Issue #582, opened 2021-10-05
how to upgrade from 0.13.0 to 0.14.0


nbari:
Hi, sorry if this is very basic, but wondering how to upgrade to latest `0.14.0`, I installed the operator `0.13.0` using: 

    kubectl apply -f https://raw.githubusercontent.com/ansible/awx-operator/devel/deploy/awx-operator.yaml

> The URL is not present anymore

Should I edit the current operator and change the version, something like:

     kubectl edit deployment.apps/awx-operator

Could someone please share the steps?

vrubiolo:
Hi @nbari , it looks like you might need to delete the current operator Deployment and recreate it, cf https://github.com/ansible/awx-operator#upgrading

tgruezi:
hi @vrubiolo is there a way to delete the operator and without removing all the resources?

nbari:
Is there a way to delete the current operator without removing the exiting resources?

vrubiolo:
Sorry I don't know. I think so (as the operator deployment and AWX itself are two different things) but I cannot be sure as I have not done it.
This warrants a doc update to me. You might want to ask in the IRC channel as well.

kurokobo:
In my case, to upgrade AWX Operator and related AWX instance, delete existing AWX Operator and some unnecessary resources first,

```bash
kubectl -n default delete deployment awx-operator
kubectl -n default delete serviceaccount awx-operator
kubectl -n default delete clusterrolebinding awx-operator
kubectl -n default delete clusterrole awx-operator
```

then simply re-deploy new AWX Operator.

```bash
# Prepare required files
git clone https://github.com/ansible/awx-operator.git
cd awx-operator
git checkout 0.14.0

# Deploy new AWX Operator
export NAMESPACE=<Name of the namespace where your old AWX instanse exists>
make deploy
```

Once the new AWX Operator is up and running, it will start rolling out a new version of the AWX instance automatically based on the old awx resource definition.

I don't know if it will suit your use case perfectly because of the particular configuration for my use case, but there are some notes in my repository that might be helpful: <https://github.com/kurokobo/awx-on-k3s/blob/main/tips/upgrade-operator.md>

It is strongly recommended that [making a backup](https://github.com/ansible/awx-operator/tree/devel/roles/backup) before you start working.

vrubiolo:
[...]

PR #592, opened 2021-10-07
feat(readme): add k8s cluster setup instructions

mrwiora:
for those running a k8s cluster, the instructions are not very clear. 
Especially for those ones which were formerly using 0.13 and the deploy/awx-operator.yml file.
This was discussed in issue https://github.com/ansible/awx-operator/issues/582 and this PR adds instructions that helps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,13 +203,14 @@ For an example using the Nginx Controller in Minukube, don't miss our [demo vide
 ### Basic Install on existing cluster
 
 For those running a whole K8S Cluster the steps to set up the awx-operator are:
-```bash
-# Prepare required files
+
+```
+$ Prepare required files
 git clone https://github.com/ansible/awx-operator.git
 cd awx-operator
-git checkout 0.14.0
+git checkout {{ latest_released_version }} # replace variable by latest version number in releases
 
-# Deploy new AWX Operator
+$ Deploy new AWX Operator
 export NAMESPACE=<Name of the namespace where your AWX instanse exists>
 make deploy
 ```
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ An [Ansible AWX](https://github.com/ansible/awx) operator for Kubernetes built w
 * [Table of Contents](#table-of-contents)
    * [Purpose](#purpose)
    * [Usage](#usage)
-      * [Basic Install](#basic-install)
+      * [Basic Install on minikube for beginner or testing](#basic-install-on-minikube-for-beginner-or-testing)
       * [Basic Install on K8S Cluster](#basic-install-on-k8s-cluster)
       * [Admin user account configuration](#admin-user-account-configuration)
       * [Network and TLS Configuration](#network-and-tls-configuration)
@@ -49,7 +49,7 @@ This operator is meant to provide a more Kubernetes-native installation method f
 
 ## Usage
 
-### Basic Install
+### Basic Install on minikube for beginner or testing
 
 This Kubernetes Operator is meant to be deployed in your Kubernetes cluster(s) and can manage one or more AWX instances in any namespace.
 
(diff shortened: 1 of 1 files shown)


kurokobo:
Thanks for referring my comment.

In my opinion, if there are both "Basic Install" and "Basic Install on K8S Cluster", users will be confused because minikube is also a kind of "K8S Cluster".

The need to clone and checkout the repo are [already described in "Basic Install" section](https://github.com/ansible/awx-operator/blob/ee2980b8bbfac8c68b217e51d9e676743ecd57b3/README.md?plain=1#L104-L108):
````md
Now you need to deploy AWX Operator into your cluster. Clone this repo and `git checkout` the latest version from https://github.com/ansible/awx-operator/releases, and then run the following command:

```
$ export NAMESPACE=my-namespace
$ make deploy
...
````

Therefore, if you want to make the procedure more detailed, isn't it enough to modify above description and commands, instead of adding a new section?

**Comment by mrwiora** [comment]
@kurokobo thanks for the hint with the possible confusion - I declared the first as beginners (option 1) and the other one should be obvious for those who know what they do.
As the first one is for beginners it's okay, that the git clone command is explained, for the second it should be fair enough to have the git clone instructions in a code segment.

File: README.md [hunk 1]
```diff
@@ -199,6 +200,20 @@ For an example using the Nginx Controller in Minukube, don't miss our [demo vide
 
 [![asciicast](https://raw.githubusercontent.com/ansible/awx-operator/devel/docs/awx-demo.svg)](https://asciinema.org/a/416946)
 
+### Basic Install on K8S Cluster
+
+For those running a whole K8S Cluster the steps to set up the awx-operator are:
+```bash
```
shanemcd: For consistent formatting with the rest of the readme, please:

- Remove the `bash` here. 
- Update to look like the other code blocks in this readme (include the `$ ` shell prompts). 
- Please also add a blank newline before the code block.

README.md
```diff
@@ -48,7 +49,7 @@ This operator is meant to provide a more Kubernetes-native installation method f
 
 ## Usage
 
-### Basic Install
+### Basic Install on minikube for beginner or testing
```
shanemcd:
```suggestion
### Basic Install on minikube (beginner or testing)
```

README.md
```diff
@@ -199,6 +200,20 @@ For an example using the Nginx Controller in Minukube, don't miss our [demo vide
 
 [![asciicast](https://raw.githubusercontent.com/ansible/awx-operator/devel/docs/awx-demo.svg)](https://asciinema.org/a/416946)
 
+### Basic Install on K8S Cluster
```
shanemcd:
```suggestion
### Basic Install on existing cluster
```

README.md
```diff
@@ -199,6 +200,20 @@ For an example using the Nginx Controller in Minukube, don't miss our [demo vide
...
+
+For those running a whole K8S Cluster the steps to set up the awx-operator are:
+```bash
+# Prepare required files
+git clone https://github.com/ansible/awx-operator.git
+cd awx-operator
+git checkout 0.14.0
```
shanemcd: Please remove the hardcoded version and add some kind of placeholder.

shanemcd (changes requested):

README.md
mrwiora: thanks man for the review and the suggestions - all applied :)
https://github.com/mrwiora/awx-operator/commit/97a97ded0df2f3c921016aa71e55c145b9ed2a48

README.md
mrwiora: same here
https://github.com/mrwiora/awx-operator/commit/97a97ded0df2f3c921016aa71e55c145b9ed2a48

shanemcd (approved):

PR #600, opened 2021-10-10
allow extra config postgres with args

### LINKED ISSUE #538: change postgres settings
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
need to change postgres settings

##### ENVIRONMENT
* AWX version: 19.3
* Operator version: 0.13
* AWX install method: minikube

##### STEPS TO REPRODUCE

I need to change postgres setting to avoid production failures
https://github.com/ansible/awx/issues/10996#issuecomment-917686258

##### EXPECTED RESULTS

postgres should be run with modified settings

##### ACTUAL RESULTS

postgres is running with default settings

##### ADDITIONAL INFORMATION

https://github.com/ansible/awx/issues/10996#issuecomment-917686258
here is the problem and the way how to resolve it. Need to know how to implement this via awx-operator

chris93111:
allow change settings for postgresql with extra args in deployment

fix https://github.com/ansible/awx-operator/issues/538

```
postgres_extra_args:
  - '-c'
  - 'max_connections=1000'
```


roles/installer/templates/postgres.yaml.j2
```diff
@@ -41,6 +41,10 @@ spec:
         - image: '{{ postgres_image }}:{{ postgres_image_version }}'
           imagePullPolicy: '{{ image_pull_policy }}'
           name: postgres
+{% if postgres_extra_args %}
+          command: postgres
```
shanemcd: I'm curious about this part. Does it not work without this line? The default entrypoint seems to do some stuff that we might not want to lose: https://github.com/docker-library/postgres/blob/ab940cbb923af99e2c7cf0e0ba5305bc6815aecc/12/bullseye/docker-entrypoint.sh

chris93111: yes it's work without, i delete

kdelee (approved):
This would be helpful

rooftopcellist:
@chris93111 I have tested out your changes and they look good, but it is gated by CI.  Do you mind amending your commit, then re-pushing to this branch to kick of CI again?  Then we can get this merged.  Sorry for the hassle.

rooftopcellist (approved):

chris93111:
@rooftopcellist no problem i do quickly

rooftopcellist:
@chris93111 thank you for the fast response!  I actually just squashed your commits and opened this PR: https://github.com/ansible/awx-operator/pull/753, so no need to do the force-push if you haven't already.  We just need to wait for CI to run, then I'll get this merged.  Thank you for your contribution!

rooftopcellist:
These changes merged as part of this PR: https://github.com/ansible/awx-operator/pull/753

PR #616, opened 2021-10-21
Use relatedImages to make disconnected deployments possible

rooftopcellist:
* Add ability to pass images in from the CSV for disconnected installs

### DIFF
--- a/roles/backup/tasks/init.yml
+++ b/roles/backup/tasks/init.yml
@@ -63,6 +63,17 @@
   when:
     - backup_pvc == '' or backup_pvc is not defined
 
+- name: Set user provided postgres image
+  set_fact:
+    _custom_postgres_image: "{{ postgres_image }}:{{ postgres_image_version }}"
+  when:
+    - postgres_image | default([]) | length
+    - postgres_image_version is defined and postgres_image_version != ''
+
+- name: Set Postgres image URL
+  set_fact:
+    _postgres_image: "{{ _custom_postgres_image | default(lookup('env', 'RELATED_IMAGE_POSTGRES')) }}"
+
 - name: Create management pod from templated deployment config
   k8s:
     name: "{{ ansible_operator_meta.name }}-db-management"
--- a/roles/installer/tasks/database_configuration.yml
+++ b/roles/installer/tasks/database_configuration.yml
@@ -52,6 +52,17 @@
     _pg_config: '{{ _custom_pg_config_resources["resources"] | default([]) | length | ternary(_custom_pg_config_resources, _default_pg_config_resources) }}'
   no_log: true
 
+- name: Set user provided postgres image
+  set_fact:
+    _custom_postgres_image: "{{ postgres_image }}:{{ postgres_image_version }}"
+  when:
+    - postgres_image | default([]) | length
+    - postgres_image_version is defined and postgres_image_version != ''
+
+- name: Set Postgres image URL
+  set_fact:
+    _postgres_image: "{{ _custom_postgres_image | default(lookup('env', 'RELATED_IMAGE_POSTGRES')) }}"
+
 - block:
     - name: Create Database configuration
       k8s:
--- a/roles/installer/tasks/resources_configuration.yml
+++ b/roles/installer/tasks/resources_configuration.yml
@@ -32,6 +32,49 @@
     - 'ingress'
   no_log: true
 
+- name: Set user provided awx app image
+  set_fact:
+    _custom_image: "{{ image }}:{{ image_version }}"
+  when:
+    - image | default([]) | length
+    - image_version is defined or image_version != ''
+
+- name: Set AWX App image URL
+  set_fact:
+    _image: "{{ _custom_image | default(lookup('env', 'RELATED_IMAGE_APP')) }}"
+
+- name: Set user provided controller init image
+  set_fact:
+    _custom_init_container_image: "{{ init_container_image }}:{{ init_container_image_version }}"
+  when:
+    - init_container_image | default([]) | length
+    - init_container_image_version is defined or init_container_image_version != ''
+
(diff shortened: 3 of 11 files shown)


roles/backup/tasks/init.yml
```diff
@@ -63,6 +63,17 @@
   when:
     - backup_pvc == '' or backup_pvc is not defined
 
+- name: Set user provided postgres image
+  set_fact:
+    _custom_postgres_image: "{{ postgres_image }}:{{ postgres_image_version }}"
+  when:
+    - postgres_image is defined
```
tchellomello: @rooftopcellist  how about using a defensive code here and ensure the `length` as well? Like what would happen if the user leave the `postgres_image:   ` with an empty space while editing the AWX spec? This way we can enforce the user added some information on it. 
```yaml
when: 
  - postgres_image | default([]) | length
  - postgres_image_version | default([]) | length
 ```

File: roles/installer/tasks/database_configuration.yml [hunk 2]
```diff
@@ -52,6 +52,17 @@
     _pg_config: '{{ _custom_pg_config_resources["resources"] | default([]) | length | ternary(_custom_pg_config_resources, _default_pg_config_resources) }}'
   no_log: true
 
+- name: Set user provided postgres image
+  set_fact:
+    _custom_postgres_image: "{{ postgres_image }}:{{ postgres_image_version }}"
+  when:
+    - postgres_image is defined
+    - postgres_image_version is defined
```
tchellomello: Same suggestion as above

roles/installer/tasks/resources_configuration.yml
```diff
@@ -32,6 +32,49 @@
     - 'ingress'
   no_log: true
 
+- name: Set user provided awx app image
+  set_fact:
+    _custom_image: "{{ image }}:{{ image_version }}"
+  when:
+    - image is defined
+    - image_version is defined
```
tchellomello: Same

roles/installer/tasks/resources_configuration.yml
```diff
@@ -32,6 +32,49 @@
...
+- name: Set AWX App image URL
+  set_fact:
+    _image: "{{ _custom_image | default(lookup('env', 'RELATED_IMAGE_APP')) }}"
+
+- name: Set user provided controller init image
+  set_fact:
+    _custom_init_container_image: "{{ init_container_image }}:{{ init_container_image_version }}"
+  when:
+    - init_container_image is defined
+    - init_container_image_version is defined
```
tchellomello: Same

roles/installer/tasks/resources_configuration.yml
```diff
@@ -32,6 +32,49 @@
...
+- name: Set Init image URL
+  set_fact:
+    _init_container_image: "{{ _custom_init_container_image | default(lookup('env', 'RELATED_IMAGE_INIT_CONTAINER')) }}"
+
+- name: Set user provided redis image
+  set_fact:
+    _custom_redis_image: "{{ redis_image }}:{{ redis_image_version }}"
+  when:
+    - redis_image is defined
+    - redis_image_version is defined
```
tchellomello: Same

roles/installer/tasks/resources_configuration.yml
```diff
@@ -32,6 +32,49 @@
...
+
+- name: Set Redis image URL
+  set_fact:
+    _redis_image: "{{ _custom_redis_image | default(lookup('env', 'RELATED_IMAGE_REDIS')) }}"
+
+- name: Set user provided control plane ee image
+  set_fact:
+    _custom_control_plane_ee_image: "{{ control_plane_ee_image}}"
+  when:
+    - control_plane_ee_image is defined
```
tchellomello: Same

roles/restore/tasks/init.yml
```diff
@@ -71,6 +71,17 @@
     force: true
     wait: true
 
+- name: Set user provided postgres image
+  set_fact:
+    _custom_postgres_image: "{{ postgres_image }}:{{ postgres_image_version }}"
+  when:
+    - postgres_image is defined
+    - postgres_image_version is defined
```
tchellomello: Same

tchellomello (approved):
@rooftopcellist  just added a small nit comment, but looks good to me! 

Thanks @rooftopcellist

rooftopcellist:
Thanks for the suggestion @tchellomello!  I just updated the PR.

shanemcd (approved):

roles/installer/tasks/database_configuration.yml
```diff
@@ -52,6 +52,17 @@
     _pg_config: '{{ _custom_pg_config_resources["resources"] | default([]) | length | ternary(_custom_pg_config_resources, _default_pg_config_resources) }}'
   no_log: true
 
+- name: Set user provided postgres image
+  set_fact:
+    _custom_postgres_image: "{{ postgres_image }}:{{ postgres_image_version }}"
+  when:
+    - postgres_image | default([]) | length
+    - postgres_image_version | default([]) | length
```
rooftopcellist: This task was failing because int doesn't have a length.


```
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'postgres_image_version | default([]) | length' failed. The error was: Unexpected templating type error occurred on ({% if postgres_image_version | default([]) | length %} True {% else %} False {% endif %}): object of type 'int' has no len()\n\nThe error appears to be in '/opt/ansible/roles/installer/tasks/database_configuration.yml': line 55, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Set user provided postgres image\n  ^ here\n"}
```

tchellomello (approved):

PR #619, opened 2021-10-26
Custom pod labels

### LINKED ISSUE #563: [Feature Request ] Add support for custom pod labels
We have several global network policies based on labels, currently the only way that I can make the awx accessible is edit the awx deployment after being created by the operator and set the correct label for allows access. Would it be possible to set the awx pod label directly on the operator ?

This request is similar to the https://github.com/ansible/awx-operator/issues/269 , but that one is about labeling the service, not the awx pod.

philipsd6:
This should resolve #563. If/when this passes muster, I'll squash so it can be merged.

### DIFF
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -83,6 +83,11 @@ topology_spread_constraints: ''
 
 affinity: {}
 
+# Custom labels for the AWX pods. Specify as literal block. E.g.:
+# pod_labels: |
+#   prometheus.io/scrape: 'true'
+pod_labels: ''
+
 # Add node tolerations for the AWX pods. Specify as literal block. E.g.:
 # tolerations: |
 #   - key: "dedicated"
--- a/README.md
+++ b/README.md
@@ -791,6 +791,24 @@ spec:
           topologyKey: topology.kubernetes.io/zone
 ```
 
+#### Adding custom labels to AWX pods
+
+You can add custom labels to the AWX pods.
+
+| Name                           | Description                 | Default |
+| -------------------------------| --------------------------- | ------- |
+| pod_labels                     | Add custom labels           | ''      |
+
+Example of customization could be:
+
+```yaml
+---
+spec:
+  ...
+  pod_labels: |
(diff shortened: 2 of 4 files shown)


roles/installer/defaults/main.yml
```diff
@@ -64,6 +64,11 @@ hostname: ''
 #   kubernetes.io/os: linux
 node_selector: ''
 
+# Custom labels for the AWX pods. Specify as literal block. E.g.:
+# pod_labels: |
+#   prometheus.io/scape: 'true'
```
shanemcd: Did you mean `scrape`?

philipsd6: Yes, sure did -- fixed, and PR updated.

shanemcd:
Hi, sorry for the slow review. Looks like this has conflicts now. Once you fix, we'll get this reviewed and merged.

philipsd6:
> Hi, sorry for the slow review. Looks like this has conflicts now. Once you fix, we'll get this reviewed and merged.

No problem @shanemcd -- I've gone ahead and rebased on devel without conflicts as requested, and squashed down to one commit.

EDIT: whoops, I forgot to pull upstream before rebasing, and my fork wasn't up-to-date, so I'll get that fixed quickly...

tchellomello (approved):
Very nice @philipsd6!! Thanks for your contribution!!

jsalatiel:
Any chance we can get this merged ?

timansky:
I am having the same problem. #1160 must resolve issue. waiting for merge

erz4 (approved):

fosterseth:
additional labels was implemented in this PR https://github.com/ansible/awx-operator/pull/1160

does that PR address the use case you had in mind @philipsd6

philipsd6:
@fosterseth I believe it does, as PR #1160 enables setting `additional_labels` which as I understand it would then apply (from the example in my README) update `prometheus.io/scrape-awx` to the pods, which was our use-case for this PR. However, the `additional_labels` solution will also apply it to the service as well, but I can't see why that would be a problem, so I would consider our use-case met by PR #1160.

PR #624, opened 2021-10-28
Set default images, but use them as a last resort

rooftopcellist:
* Only use them if user did not set an image, or RELATED_IMAGES_ var
    is not set

Part of the rationale for adding this in instead of just removing the defaults is that it it prevents someone from accidentally breaking disconnected deployment support by adding in default image values later on.


Spredzy:
Superseded by https://github.com/ansible/awx-operator/pull/630

PR #637, opened 2021-11-02
Fix kubectl commands in makefile

jul13nc:
As suggested in Issue#614, replacing kubectl by minikube kubectl --, for make deploy to work properly


shanemcd:
Hello. Sorry you ran into issues here, but not everyone uses minikube. Consider setting an alias in your shell, or just install kubectl. We're not going to accept this change.

PR #643, opened 2021-11-09
Add runasuser toggle

gamuniz:
add run_userid to allow uniform UID across task and controller-ee

### DIFF
--- a/roles/installer/templates/deployment.yaml.j2
+++ b/roles/installer/templates/deployment.yaml.j2
@@ -183,6 +183,11 @@ spec:
 {% if task_privileged == true %}
           securityContext:
             privileged: true
+{% elif run_userid %}
+          securityContext:
+            runAsUser: {{ run_userid }}
+{% else %}
+
 {% endif %}
 {% if task_command %}
           command: {{ task_command }}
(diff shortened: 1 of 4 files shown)


kdelee (approved):

roles/installer/templates/deployment.yaml.j2
```diff
@@ -183,6 +183,11 @@ spec:
 {% if task_privileged == true %}
           securityContext:
             privileged: true
+{% elif run_userid %}
+          securityContext:
+            runAsUser: {{ run_userid }}
+{% else %}
```
dsavineau: running a container with a specific user or with the `privileged` flag aren't exclusive (it's not a xor). You can have a privileged container with a non root user so the `elif` statement doesn't right to me.

dsavineau: What's the purpose of the `else` statement here ? there's nothing to do in this situation so we can drop that statement

dsavineau (changes requested):

roles/installer/templates/deployment.yaml.j2
gamuniz: I can fix that up, jinja2 isn't my favorite to work out. as for the xor, I didn't want to break existing functionality to fix this for my use-case. If you have a suggestion that works for both. I've never tried setting runasuser and privileged as it seems redundant to use runasuser and privileged: true

pabelanger:
Is this related to https://github.com/ansible/awx-operator/pull/645 ?

gamuniz:
closing in favor of #645

PR #645, opened 2021-11-10
Add support for arbitrary pod-level securityContext settings

shanemcd:
This allows for doing stuff like this on the custom resource:

```
security_context_settings:
  runAsUser: 1000
  runAsGroup: 0
```

I added `snakeCaseParameters: False` because without it, variables like `runAsUser` become `run_as_user`... and that doesnt work.


pabelanger:
https://github.com/ansible/awx-operator/pull/643 seems to be the same idea?

kdelee (approved):
Can we get this merged and an awx-operator release?

This is a blocker for some of us who are on the :cloud:

PR #661, opened 2021-11-19
Update README.md - fix status command

sooslaca:
one command is mentioned wrong in the readme and results this:
```
root@awx:~# kubectl logs -f deployments/awx-operator-controller-manager -c awx-manager
error: container awx-manager is not valid for pod awx-operator-controller-manager-68d787cfbd-lgrm8
```
proper container name is `manager`  (not `awx-manager`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ awx.awx.ansible.com/awx-demo created
 After a few minutes, the new AWX instance will be deployed. You can look at the operator pod logs in order to know where the installation process is at:
 
 ```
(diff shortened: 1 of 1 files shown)


shanemcd:
Were you using an old version of the operator? I just tested this on devel:

```
$ kubectl get pod awx-operator-controller-manager-6c96d9b446-h8ncc -o json | jq '.spec.containers[].name'
"kube-rbac-proxy"
"awx-manager"
```

Issue #668, opened 2021-11-22
Why image version in image name into the deployment


Davidffry:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
It seems, when we deploy awx that the image name with his version is inserting into the template.spec.containers[0].image : _ quay.io/ansible/awx-operator:0.15.0-11-g57aa585_

##### ENVIRONMENT
* AWX version: 0.15.0
* Operator version: 0.15.0
* Kubernetes version: 1.21.5
* AWX install method: WSL with docker desktop on Windows

##### STEPS TO REPRODUCE

Make deploy

##### EXPECTED RESULTS

```
...
** 32   template:
 33     metadata:
 34       creationTimestamp: null
 35       labels:
 36         control-plane: controller-manager
 37     spec:
 38       containers:
 39       - args:
 40         - --secure-listen-address=0.0.0.0:8443
 41         - --upstream=http://127.0.0.1:8080/
 42         - --logtostderr=true
 43         - --v=10
 44         image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
 45         imagePullPolicy: IfNotPresent
 46         name: kube-rbac-proxy
 47         ports:
 48         - containerPort: 8443
 49           name: https
 50           protocol: TCP
 51         resources: {}
 52         terminationMessagePath: /dev/termination-log
 53         terminationMessagePolicy: File
 54       - args:
 55         - --health-probe-bind-address=:6789
 56         - --metrics-bind-address=127.0.0.1:8080
 57         - --leader-elect
 58         - --leader-election-id=awx-operator
 59         env:
 60         - name: ANSIBLE_GATHERING
 61           value: explicit
 62         - name: ANSIBLE_DEBUG_LOGS
 63           value: "false"
 64         - name: WATCH_NAMESPACE
 65           valueFrom:
 66             fieldRef:
 67               apiVersion: v1
 68               fieldPath: metadata.namespace
 69         image: quay.io/ansible/awx-operator:0.15.0
 70         imagePullPolicy: IfNotPresent**
...
```

##### ACTUAL RESULTS

```
...
** 32   template:
 33     metadata:
 34       creationTimestamp: null
 35       labels:
 36         control-plane: controller-manager
 37     spec:
 38       containers:
 39       - args:
 40         - --secure-listen-address=0.0.0.0:8443
 41         - --upstream=http://127.0.0.1:8080/
 42         - --logtostderr=true
 43         - --v=10
[...]

PR #669, opened 2021-11-22
Need to get the real name of the tag for container image version

Davidffry:
please ref : #668


shanemcd:
This would break being able to do `make docker-build docker-push install deploy`.

Issue #242, opened 2021-04-23
Feature request: Ability to add custom labels to EE pods


RylandDeGregory:
I am attempting to use [AAD Pod Identity](https://azure.github.io/aad-pod-identity/docs/) to grant my AWX environment implicit access to Azure Key Vault using a user-assigned Managed Identity.  One of the requirements is that the pods be labeled with the name of the Managed Identity to use, like this:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: demo
  labels:
    aadpodidbinding: <MANAGED_IDENTITY_NAME>
```

The process works as expected, and I was able to validate by performing `kubectl label pod awx-pod aadpodidbinding=MANAGED_IDENTITY_NAME`, but *AWX generates a new pod for each job*. This prevents the authentication from working unless the EE pods can be created by the operator with that label.

I understand that there are already variables that set pod labels for use in node selection, and we already have variables that control annotations for ingress resource definitions. My question is, can there be another AWX spec variable added to set custom labels on EE pods? I know that my ask is specifically about AAD pod identity, but I'm sure there are other use cases for user-defined pod labels.

lukasertl:
I'd like to add to this request the ability to add custom annotations to the EE pod.  

With a custom annotation the EE could use the Hashicorp Vault agent injector to access secrets in Vault.

lukasertl:
Meanwhile I found that you can do this already by changing the pod definition in the tower instances Interface.

RylandDeGregory:
@lukasertl thank you so much, that worked perfectly!

tchellomello:
Just for the documentation, here is a screenshot of how you extend  custom labels to your `ee` pods; 

Instance Groups --> edit the instance group ---> `Customize pod specification` 

![image](https://user-images.githubusercontent.com/809840/116572985-6d003880-a8da-11eb-9869-3709d96fc631.png)

So I guess we already have a place for that in the AWX itself which allows more freedom than setting it on the spec itself. 

Are you good with this approach @RylandDeGregory?

RylandDeGregory:
Hey @tchellomello I understand that this is possible, and I've gotten it working using this approach. 

But, telling me to go make a pod spec customization in the GUI is not the answer I wanted. This is squarely a code change, not a DB or UI change, and should be possible to set in the code that I use to deploy AWX.

tchellomello:
[...]

PR #676, opened 2021-11-29
Add pod_labels for custom pod labels

locmai:
Adding `pod_labels` option for templating custom labels into the pods. 

This could simply solve the #242 since users could add the `aadpodbinding` label and other use cases.

Signed-off-by: Loc Mai <lmai@axon.com>

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -134,6 +134,9 @@ spec:
                 node_selector:
                   description: nodeSelector for the pods
                   type: string
+                pod_labels:
+                  description: Additional labels to apply to the pod
+                  type: string
                 service_labels:
                   description: Additional labels to apply to the service
                   type: string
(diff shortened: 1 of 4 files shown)


hungtran84:
great @locmai , exactly what i was looking for.
Related issue https://github.com/ansible/awx-operator/issues/242

kdelee:
Can you confirm that these pod labels are applied to the podspec for the `default` container group as well? if you could paste `${TOWER_URL}/api/v2/instance_groups/${ID OF DEFAULT CONTAINER GROUP}/` that would make it clear if the pod labels get applied to both the awx pods as well as the default container group

kdelee:
I'm 99% certain this won't fix applying labels to the default container group. I think we need something like https://github.com/ansible/awx/pull/11395 for it to filter through to the default container group, which is what is actually used for jobs (what you have would apply only the awx deployment itself). I'm not saying this is a bad change, just maybe won't do everything desired

locmai:
@kdelee Your PR looks better and makes more sense. I didn't look at this close enough, though it was just an easy config change. Closing this PR now.

Issue #667, opened 2021-11-22
Issue when 'make deploy' run 


Davidffry:
##### ISSUE TYPE
 - Bug Report

##### ENVIRONMENT
* AWX version: devel (after git checkout)
* Operator version: 0.15.0
* Kubernetes version: 1.21.5
* AWX install method: wsl with docker desktop for windows

##### STEPS TO REPRODUCE

```make deploy```

##### EXPECTED RESULTS

Execution

##### ACTUAL RESULTS
```
$  make deploy
make gen-resources | kubectl apply -f -
error: error validating "STDIN": error validating data: ValidationError(Namespace): unknown field "make[1]" in io.k8s.api.core.v1.Namespace; if you choose to ignore these errors, turn validation off with --validate=false
Makefile:87: recipe for target 'deploy' failed
make: *** [deploy] Error 1
```

##### ADDITIONAL INFORMATION

- One thing to change into the Makerfile : add `-k` to all curl because certificate issue.
- When rollback the commit : 782f97c42cc589419200bfccfd44c342ba010826
it seems to work, with other issue.

##### AWX-OPERATOR LOGS
No.

syswipe:
I have the same problem, but I use k3s instead of k8s cluster. Any suggestion?

jonathon2nd:
I get the same when I have devel checked-out. Try a tag, 0.15.0 works for me on a rancher managed k8s cluster, though I had to modify the makefile to include the kubeconfig.

Davidffry:
Ok, I ll recheck that. Thanks

syswipe:
i rolled back to 0.14.0 and all run well

Davidffry:
I retest it from scratch (wsl with windows) is ok
I'll test it with k3s...

Davidffry:
You're right, I close the issue. my bad ;-)

PR #686, opened 2021-12-07
Fix deploy target for the devel branch

rooftopcellist:
Related Issue: https://github.com/ansible/awx-operator/issues/667

  * piping a make target within another target causes issues

When running `make deploy` on the devel branch, I get the following error.  This fixes it, though it isn't DRY..
```
$ VERSION=0.15.0 NAMESPACE=ca-awx IMG=quay.io/chadams/awx-operator:dev make deploy
make gen-resources | kubectl apply -f -
error: error validating "STDIN": error validating data: ValidationError(Namespace): unknown field "make[1]" in io.k8s.api.core.v1.Namespace; if you choose to ignore these errors, turn validation off with --validate=false
make: *** [Makefile:87: deploy] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,9 @@ gen-resources: kustomize ## Generate resources for controller and print to stdou
 	@$(KUSTOMIZE) build config/default
 
 deploy: kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
-	$(MAKE) gen-resources | kubectl apply -f -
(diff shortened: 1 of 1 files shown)


rooftopcellist:
@kdelee and I just built an awx image from this branch and deployed it with the awx-operator and can now run the awx-manage command, which allows us to run migrations.

kdelee (approved):

PR #688, opened 2021-12-07
Wait for Postgres to initialize before starting containers

rooftopcellist:
To avoid a bad state that can occur if the app containers come online before the postgres pod is up, we should add the following task to wait for Postgres to initialize.  

This used to exist in the launch_awx.sh script, but was removed as ansible is no longer available in the task container at that point. 

The ascii video below demonstrates that the wait task is honored.  
postgres-wait


fosterseth (approved):

rebeccahhh (approved):

AlanCoding:
Yes, that was removed in https://github.com/ansible/awx/commit/ddcbc408b908626d42f8d5393bf61abaf5ecb310. Clearly, the solution here wouldn't be affected by not having ansible, which was the reason that was done.

Issue #618, opened 2021-10-26
Migrations failed with external database on upgrade


MichaelPak:
##### ENVIRONMENT
* AWX version: 19.4.0
* Operator version: 0.14.0
* Kubernetes version: 1.21.1
* AWX install method: kind

##### SUMMARY
I am trying to update AWX from version 14 to the latest. In production, the application runs in a `docker-compose`. Raised a clean install locally through an operator in [kind](https://kind.sigs.k8s.io/), everything works. I got the idea to try installing with an old database to check that the migrations were successful. I installed the Postgresql locally, uploaded the backup there. I made it possible to connect to the local database through the [dockerhost](https://github.com/qoomon/docker-host).

```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: dockerhost
  labels:
    k8s-app: dockerhost
spec:
  replicas: 1
  selector:
    matchLabels:
      k8s-app: dockerhost
  template:
    metadata:
      labels:
        k8s-app: dockerhost
    spec:
      containers:
      - name: dockerhost
        image: qoomon/docker-host
        securityContext:
          capabilities:
            add: ["NET_ADMIN", "NET_RAW"]

---
apiVersion: v1
kind: Service
metadata:
  name: dockerhost
spec:
  clusterIP: None
  selector:
    k8s-app: dockerhost
```

In PostgreSQL I created two databases: `awx` for the old backup and `awxnew` for the new installation.

```
---
apiVersion: v1
kind: Secret
metadata:
  name: awx-secret-key
  namespace: awx
stringData:
  secret_key: "awxsecret"
type: Opaque

---
apiVersion: v1
kind: Secret
metadata:
  name: awx-old-postgres-configuration
  namespace: awx
stringData:
  host: dockerhost
  port: "5432"
  database: awx
  username: awx
  password: awx
type: Opaque

---
apiVersion: v1
kind: Secret
metadata:
  name: awx-postgres-configuration
  namespace: awx
stringData:
  host: dockerhost
  port: "5432"
  database: awxnew
  username: awxnew
  password: awxnew
  sslmode: disable
  type: unmanaged
type: Opaque

---
apiVersion: awx.ansible.com/v1beta1
kind: AWX
metadata:
  name: awx
spec:
  service_type: NodePort
  old_postgres_configuration_secret: awx-old-postgres-configuration
  postgres_configuration_secret: awx-postgres-configuration
  secret_key_secret: awx-secret-key
```

[...]

PR #695, opened 2021-12-14
fix issue #618 with migration script when the remote database use the non public schema

absynth76:


rooftopcellist:
This was implemented and merged in this PR:
* https://github.com/ansible/awx-operator/pull/923

Could you please open a new issue if this doesn't cover your use case?

Thanks!

PR #701, opened 2021-12-16
Enable setting a list of image_pull_secrets

MaximilianMeister:
When there are e.g. multiple authenticated container registries used
we need to be able to add multiple imagePullSecrets to the k8s resource

NOTE: I haven't run any tests so far to confirm that it works

### DIFF
--- a/roles/installer/templates/deployment.yaml.j2
+++ b/roles/installer/templates/deployment.yaml.j2
@@ -29,9 +29,14 @@ spec:
         app.kubernetes.io/component: '{{ deployment_type }}'
     spec:
       serviceAccountName: '{{ ansible_operator_meta.name }}'
-{% if image_pull_secret %}
+{% if image_pull_secret is defined %}
       imagePullSecrets:
         - name: {{ image_pull_secret }}
+{% elif image_pull_secrets | length > 0 %}
+      imagePullSecrets:
+{% for secret in image_pull_secrets %}
+        - name: {{ secret }}
+{% endfor %}
 {% endif %}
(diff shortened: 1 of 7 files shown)


roles/installer/templates/deployment.yaml.j2
```diff
@@ -29,9 +29,11 @@ spec:
         app.kubernetes.io/component: '{{ deployment_type }}'
     spec:
       serviceAccountName: '{{ ansible_operator_meta.name }}'
-{% if image_pull_secret %}
+{% if image_pull_secrets | length > 0 %}
```
kdelee: can we do something with `if image_pull_secret is defined` and if singular `image_pull_secret` is defined (as would be the case with older specs) use that as a fall back?

MaximilianMeister: @kdelee thanks for the hint 👍 I changed it a bit to your suggestion, that should keep it backwards compatible

rooftopcellist:
Sorry for the long wait @MaximilianMeister, thank you for your patience.  There were a couple conflicts because of other changes recently merged.  I picked your commit into a [new PR](https://github.com/ansible/awx-operator/pull/860) and fixes the conflicts.  I will have this merged before EOD.  Thank you for contributing this and for making the suggested changes to make it backwards compatible!

rooftopcellist:
This change has now merged in [PR 860](https://github.com/ansible/awx-operator/pull/860).  Thank you!

MaximilianMeister:
@rooftopcellist thanks for picking this up 👍

PR #703, opened 2021-12-17
update README.md

besmirzanaj:
existing documentation is a bit confusing on the install part and does not work.
Updated to make it work but still not perfect as kustomize binary needs to be on its own path

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,9 +105,12 @@ Now you need to deploy AWX Operator into your cluster. Clone this repo and `git 
 
 ```
 $ export NAMESPACE=my-namespace
-$ make deploy
-cd config/manager && /home/user/awx-operator/bin/kustomize edit set image controller=quay.io/ansible/awx-operator:0.14.0
-/home/user/awx-operator/bin/kustomize build config/default | kubectl apply -f -
+$ #install kustomize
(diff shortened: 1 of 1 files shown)


tchellomello (changes requested):
Thanks for your contribution @besmirzanaj. 

It seems that we already have a target on the `Makefile` that covers this (see https://github.com/ansible/awx-operator/blob/devel/Makefile#L75-L76) 

That means whenever you do `make deploy` the target `kustomize` will be invoked when the binary is not present on the `PATH`.

Just to provide an example:

```bash
(python3.10) mdemello@storm ~/d/R/awx-operator (devel) [0|1]> ls -la bin                                                                                         15:20:54
ls: cannot access 'bin': No such file or directory
(python3.10) mdemello@storm ~/d/R/awx-operator (devel) [0|2]> make kustomize                                                                                     15:20:56
(python3.10) mdemello@storm ~/d/R/awx-operator (devel)> ls -la bin/kustomize                                                                                     15:20:58
-rwxr-xr-x. 1 mdemello mdemello 34041856 Nov 11  2020 bin/kustomize
```

Maybe you want to clarify the use of that target to explain to users that the `kustomize` binary will be installed automatically.  Does it make sense?

shanemcd:
Closing due to lack of response.

PR #717, opened 2022-01-04
Scale down app pod when database is unavailable

rooftopcellist:
In some upgrade scenarios, the postgres StatefulSet will be changed, resulting in the postgres pod being terminated and re-created.  As a result, the database becomes unavailable temporarily, which causes the application to throw 500 errors and often will not recover even when the database is avialable again.  

To avoid this, the app pod deployment should be scaled down to 0 in cases where postgresql is expected to be unavailable.  In this case, whenever the postgres StatefulSet it changed, we can assume the postgres pod will be unavailable temporarily.


tchellomello (approved):
Yes, it does make sense!!  great work @rooftopcellist
you++

PR #718, opened 2022-01-04
[Do Not Merge Yet] Use pgsql subPath directory inside volume to remain consistent

rooftopcellist:
Use pgsql subPath directory inside volume to remain consistent with past versions

  * data will not persist when upgrading if the subPath is different.
    The data is not lost, but the wrong directory in the volume is
mounted, which results in a new db getting initialized.

Now that the default value for postgres_data_dir is `/var/lib/pgsql/data/userdata`, this change is needed so that the subPath remains pgsql.  


rooftopcellist:
This would cause issues for upgrades from the current version forward... As such I am closing this.

PR #726, opened 2022-01-11
Add topology constraints to AWX CRD

Skaopap:
# Description 

Add the ability to specify topologySpreadConstraints for AWX pods.  
  
Usefull for AWS EKS multi-AZ deployment using karpenter autoscaler.  
  
# Changes

- Added the section in the awxs.awx.ansible.com CRD
- Added the section in the deployment template
- Added the section in the olm-parameters
- Added the variable in defaults along with some documentation
- Added documentation in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -537,16 +537,18 @@ spec:
 You can constrain the AWX pods created by the operator to run on a certain subset of nodes. `node_selector` and `postgres_selector` constrains
 the AWX pods to run only on the nodes that match all the specified key/value pairs. `tolerations` and `postgres_tolerations` allow the AWX
 pods to be scheduled onto nodes with matching taints.
+The ability to specify topologySpreadConstraints is also allowed through `topology_spread_constraints`  
 
 
-| Name                           | Description                 | Default |
-| -------------------------------| --------------------------- | ------- |
-| postgres_image                 | Path of the image to pull   | 12      |
-| postgres_image_version         | Image version to pull       | 12      |
-| node_selector                  | AWX pods' nodeSelector      | ''      |
-| tolerations                    | AWX pods' tolerations       | ''      |
-| postgres_selector              | Postgres pods' nodeSelector | ''      |
-| postgres_tolerations           | Postgres pods' tolerations  | ''      |
+| Name                           | Description                              | Default |
+| -------------------------------| ---------------------------------------- | ------- |
+| postgres_image                 | Path of the image to pull                | 12      |
+| postgres_image_version         | Image version to pull                    | 12      |
+| node_selector                  | AWX pods' nodeSelector                   | ''      |
+| topology_spread_constraints    | AWX pods' topologySpreadConstraints      | ''      |
+| tolerations                    | AWX pods' tolerations                    | ''      |
+| postgres_selector              | Postgres pods' nodeSelector              | ''      |
+| postgres_tolerations           | Postgres pods' tolerations               | ''      |
 
 Example of customization could be:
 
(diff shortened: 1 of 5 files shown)


kdelee:
@gamuniz @rooftopcellist I think this would be good to get in, as these topologySpreadConstraints can help us get the cluster to scale up and influence the scheduler to spread the controller pods out over different nodes in a way that may not happen automatically, given relatively low resource requests, we may end up with clumping of controller pods on a single node that could not actually handle all the pods operating at their max limits

roles/installer/defaults/main.yml
```diff
@@ -64,6 +64,17 @@ hostname: ''
 #   kubernetes.io/os: linux
 node_selector: ''
 
+# Add a topologySpreadConstraints for the AWX pods.
+# Specify as literal block. E.g.:
+# topology_spread_constraints: |
+#   - maxSkew: "100"
```
rooftopcellist: @Skaopap could you remove the quotationso on the value of the `maxSkew` value?  That will fail with the following error if configured when the deployment gets applied to the cluster:

```
TASK [Apply deployment resources] ******************************** 
fatal: [localhost]: FAILED! => {"changed": false, "error": 400, "msg": "Failed to apply object: b'{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"Deployment in version \\\\\"v1\\\\\" cannot be handled as a Deployment: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.TopologySpreadConstraints: []v1.TopologySpreadConstraint: v1.TopologySpreadConstraint.MaxSkew: readUint32: unexpected character: \\\\ufffd, error found in #10 byte of ...|maxSkew\\\\\":\\\\\"100\\\\\",\\\\\"topo|..., bigger context ...|p.kubernetes.io/name\\\\\":\\\\\"resourcename\\\\\"}},\\\\\"maxSkew\\\\\":\\\\\"100\\\\\",\\\\\"topologyKey\\\\\":\\\\\"topology.kubernetes.io/zone\\\\\",\\\\\"|...\",\"reason\":\"BadRequest\",\"code\":400}\\n'", "reason": "Bad Request", "status": 400}
```

README.md
```diff
@@ -558,6 +560,13 @@ spec:
     disktype: ssd
     kubernetes.io/arch: amd64
     kubernetes.io/os: linux
+  topology_spread_constraints: |
+    - maxSkew: "100"
```
rooftopcellist: Same here, this is an int, so we should remove the quotations on the maxSkew value.  See this example in the k8s doc: https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/#api

rooftopcellist (changes requested):
This looks great, thank you for your contribution @Skaopap !  Other than those two small changes to the examples, This PR is ready to get merged.  After those changes are made, I'll approve and as long as CI passes, we can merge this.  

I tried this out with the following spec:

```
spec:

...

  node_selector: |
    kubernetes.io/arch: amd64
    kubernetes.io/os: linux
  topology_spread_constraints: |
    - maxSkew: 100
      topologyKey: "topology.kubernetes.io/zoneA"
      whenUnsatisfiable: "ScheduleAnyway"
      labelSelector:
        matchLabels:
          kubernetes.io/hostname: "crc-x4qnm-master-0"
```


In my testing, the topology_spread_constraints parameter gets added to the spec as expected:

```
topology_spread_constraints: |
  - maxSkew: 100
    topologyKey: "topology.kubernetes.io/zone"
    whenUnsatisfiable: "ScheduleAnyway"
    labelSelector:
      matchLabels:
        kubernetes.io/hostname: "crc-x4qnm-master-0"
```

And the pods get scheduled correctly.  

Note that the AWX pod will not schedule if the specified node label is not matched. That error would manifest on the pod itself, for example:

```
kubectl describe pod awx-5dcb85f567-hxx4w

...

Events:
  Type     Reason            Age                  From               Message
  ----     ------            ----                 ----               -------
  Warning  FailedScheduling  19s (x7 over 6m23s)  default-scheduler  0/1 nodes are available: 1 node(s) didn't match Pod's node affinity/selector.
```

Skaopap:
Thanks a lot ! I just commited the changes you asked for. 

Btw thanks for all the work you've done on this operator, it's awesome.

rooftopcellist (approved):

rooftopcellist:
Thank you @Skaopap, and thank you for helping make it even better :smile: !

PR #753, opened 2022-01-11
Add ability to configure extra args for postgres

### LINKED ISSUE #538: change postgres settings

rooftopcellist:
## Summary

To get https://github.com/ansible/awx-operator/pull/600 merged, our CI needed a new commit hash in order to get retriggered.  I cherry-picked over @chris93111's commits, the squashed them in this PR.   

* add default extra args postgres
* add postgres_extra_args option to readme

## Info from original message

allow change settings for postgresql with extra args in deployment

fix https://github.com/ansible/awx-operator/issues/538

```
postgres_extra_args:
  - '-c'
  - 'max_connections=1000'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -422,6 +422,9 @@ spec:
     limits:
       storage: 50Gi
   postgres_storage_class: fast-ssd
+  postgres_extra_args:
+    - '-c'
+    - 'max_connections=1000'
 ```
 
 **Note**: If `postgres_storage_class` is not defined, Postgres will store it's data on a volume using the default storage class for your cluster.
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -362,6 +362,10 @@ spec:
                 postgres_data_path:
                   description: Path where the PostgreSQL data are located
                   type: string
+                postgres_extra_args:
+                  type: array
+                  items:
+                    type: string
                 ca_trust_bundle:
                   description: Path where the trusted CA bundle is available
                   type: string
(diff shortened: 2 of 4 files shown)


kdelee (approved):

rooftopcellist:
The `postgres_extra_args` are getting added to the Postgres StatefulSet correctly:

```
$ oc get sts awx-postgres -o yaml | grep ' max_connections=1000' -A 2 -B 4
    spec:
      containers:
      - args:
        - -c
        - max_connections=1000
        env:
        - name: POSTGRESQL_DATABASE

```

But it looks there if `postgres_extra_args` is not set, it will cause the installer to fail. See operator logs below:

```
fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'postgres_extra_args' is undefined\n\nThe error appears to be in '/opt/ansible/roles/installer/tasks/database_configuration.yml': line 96, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n- block:\n    - name: Create Database if no database is specified\n      ^ here\n"}
```

_working on a fix now_

rooftopcellist:
We should add OLM parameters for this as well, but I will handle that in a follow-up PR.  Thank you for your contribution @chris93111 !

PR #772, opened 2022-01-31
Always run database-check initContainer

rooftopcellist:
Follow-up for https://github.com/ansible/awx-operator/pull/755

The init container was only being called in certain cases, this change will ensure that it happens every time.

### DIFF
--- a/roles/installer/tasks/database_configuration.yml
+++ b/roles/installer/tasks/database_configuration.yml
@@ -157,6 +157,10 @@
   retries: 60
   when: pg_config['resources'][0]['data']['type'] | default('') | b64decode == 'managed'
 
+- name: Set database as managed
+  set_fact:
+    managed_database: "{{ pg_config['resources'][0]['data']['type'] | default('') | b64decode == 'managed' }}"
+
 - name: Look up details for this deployment
   k8s_info:
     api_version: "{{ api_version }}"
--- a/roles/installer/templates/deployment.yaml.j2
+++ b/roles/installer/templates/deployment.yaml.j2
@@ -34,6 +34,7 @@ spec:
         - name: {{ image_pull_secret }}
 {% endif %}
       initContainers:
+{% if managed_database %}
         - name: database-check
           image: '{{ _init_container_image }}'
           imagePullPolicy: '{{ image_pull_policy }}'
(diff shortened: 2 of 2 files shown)


Zokormazo:
I've been thinking... we only want to run db-checker when we go with managed db right? We should condition on that.

Anyhow, I tried with this PR, and I'm getting a templating error:
```
[0;31mfatal: [localhost]: FAILED! => {"changed": false, "module_stderr": "Traceback (most recent call last):\n  File \"/opt/ansible/.ansible/tmp/ansible-tmp-1643706559.327041-675-84201192897653/AnsiballZ_k8s.py\", line 102, in <module>\n    _ansiballz_main()\n  File \"/opt/ansible/.ansible/tmp/ansible-tmp-1643706559.327041-675-84201192897653/AnsiballZ_k8s.py\", line 94, in _ansiballz_main\n    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)\n  File \"/opt/ansible/.ansible/tmp/ansible-tmp-1643706559.327041-675-84201192897653/AnsiballZ_k8s.py\", line 40, in invoke_module\n    runpy.run_module(mod_name='ansible_collections.kubernetes.core.plugins.modules.k8s', init_globals=None, run_name='__main__', alter_sys=True)\n  File \"/usr/lib64/python3.6/runpy.py\", line 205, in run_module\n    return _run_module_code(code, init_globals, run_name, mod_spec)\n  File \"/usr/lib64/python3.6/runpy.py\", line 96, in _run_module_code\n    mod_name, mod_spec, pkg_name, script_name)\n  File \"/usr/lib64/python3.6/runp...
```
but I would say the error was there since https://github.com/ansible/awx-operator/pull/755 on db-checker block and now I'm hiting this because I was skipping that block on my previous attempts.

`volumes` and `claimName` are not properly idented here? https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/deployment.yaml.j2#L50-L53

tchellomello (approved):

rooftopcellist:
Thank you.  The syntax issue with the deployment.yml.j2 template (mount indentation) is now fixed. I also fixed an issue where if /check-db/pgsql/data/ doesn't exist (on fresh installs), the initContainer gets in a fail backoff loop.

I'll merge this shortly after a couple more sanity checks.

PR #791, opened 2022-02-11
set memory setting based on resource settings

kdelee:
This way we can start using this setting in AWX again to help fix
https://github.com/ansible/awx/issues/11640

This requires https://github.com/ansible/awx/pull/11725 to merge first

### DIFF
--- a/roles/installer/templates/config.yaml.j2
+++ b/roles/installer/templates/config.yaml.j2
@@ -32,6 +32,16 @@ data:
 
     IS_K8S = True
 
+    # Set memory available based off of resource request/limit for the task pod
+    memory_limit = '{{ task_resource_requirements["limits"]["memory"] if "limits" in task_resource_requirements and "memory" in task_resource_requirements["limits"] }}'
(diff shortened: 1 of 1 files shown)


roles/installer/templates/config.yaml.j2
```diff
@@ -30,6 +30,13 @@ data:
     JOBOUTPUT_ROOT = '/var/lib/awx/job_status'
 
     IS_K8S = True
+    SYSTEM_TASK_ABS_MEM = ''
+    task_resource_requirements = '{{ task_resource_requirements }}'
```
kdelee: this probably need some filter to get the `task_resource_requirements` to render to valid python dictionary

shanemcd:
I think these tests will fail until we release a version of AWX that contains https://github.com/ansible/awx/pull/11725

kdelee:
sanity check: the rendered settings.py
```
# Set memory available based off of resource request/limit for the task pod
memory_limit = ''
memory_request = '1Gi'
SYSTEM_TASK_ABS_MEM = memory_limit if memory_limit else memory_request

# Set cpu available based off of resource request/limit for the task pod
cpu_limit = ''
cpu_request = '500m'
SYSTEM_TASK_ABS_CPU = cpu_limit if cpu_limit else cpu_request
```

Issue #802, opened 2022-02-21
Multi-Attach error for `check-db-pvc` Volume used by `database-check` initContainer


stanislav-zaprudskiy:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
AWX installation with managed DB hangs, as it can't attach `check-db-pvc` Volume required by `database-check` initContainer mounts. The Volume in question is coming from the managed DB's PVC and is attached to its StatefulSet, thus it can't be attached to other Pods (`accessModes: [ReadWriteOnce]`)

##### ENVIRONMENT
* AWX version: 20.0.0
* Operator version: 0.17.0
* Kubernetes version: 1.20.8
* AWX install method: normal Kubernetes

##### STEPS TO REPRODUCE

Just follow normal installation guidelines using the versions provided.

##### EXPECTED RESULTS

Installation comes to completion successfully.

##### ACTUAL RESULTS

Installation hangs and never completes.

##### ADDITIONAL INFORMATION

Kubernetes emits the following events:

```
Warning   FailedAttachVolume       pod/awx-test-pr-43-5cc8f75ff9-g976f                        Multi-Attach error for volume "pv-awx-qa-9ead74f3f626427c8e22ca4d522794f8-b9be9476-9e1a-4fdd-994d-46e46af188c3" Volume is already used by pod(s) awx-test-pr-43-postgres-0
```

```
Warning   FailedMount              pod/awx-test-pr-43-5cc8f75ff9-g976f                        Unable to attach or mount volumes: unmounted volumes=[check-db-pvc], unattached volumes=[receptor-socket awx-test-pr-43-redis-data rsyslog-socket awx-test-pr-43-projects rsyslog-dir awx-test-pr-43-application-credentials supervisor-socket awx-test-pr-43-bundle-cacert awx-test-pr-43-nginx-conf ca-trust-extracted awx-test-pr-43-receptor-config awx-test-pr-43-secret-key check-db-pvc awx-test-pr-43-token-rpmvh awx-test-pr-43-redis-config awx-test-pr-43-settings awx-test-pr-43-redis-socket]: timed out waiting for the condition
```

Similar installations worked fine with version of operator 0.14.0.

And interestingly, the same installation comes to completion when deploying into Docker for Mac (3.6.0; engine 20.10.8) with Kubernetes v1.21.3. And just for the sake of completeness, installations using external PostgreSQL using the referenced versions of Kubernetes and AWX/operator work fine.

[...]

PR #803, opened 2022-02-21
Run database-check initContainer on postgres sts instead

rooftopcellist:
Fixes Issue: https://github.com/ansible/awx-operator/issues/802

This avoids issues with multple initContainers trying to mount the postgres pvc at once, as is the case when there are multiple replicas.  Instead of mounting the initContainer and running the check on the deployment config, it is now run on the postgres StatefulSet, and therefore will only be run once per deployment.

### DIFF
--- a/roles/installer/templates/postgres.yaml.j2
+++ b/roles/installer/templates/postgres.yaml.j2
@@ -40,13 +40,7 @@ spec:
       initContainers:
         - name: database-check
           image: '{{ _init_container_image }}'
-          resources:
-            requests:
-              memory: "64Mi"
-              cpu: "250m"
-            limits:
-              memory: "256Mi"
-              cpu: "1000m"
+          resources: {{ postgres_init_container_resource_requirements }}
           imagePullPolicy: '{{ image_pull_policy }}'
           command:
             - /bin/sh
--- a/roles/installer/templates/postgres.yaml.j2
+++ b/roles/installer/templates/postgres.yaml.j2
@@ -37,6 +37,26 @@ spec:
       imagePullSecrets:
         - name: {{ image_pull_secret }}
 {% endif %}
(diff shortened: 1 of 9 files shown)


shanemcd (approved):

shanemcd:
For some reason, the CI checks are hanging when trying to launch the Demo Job Template. I dont see anything here that _looks_ suspect, but other PRs are running fine.

roles/installer/templates/postgres.yaml.j2
```diff
@@ -37,6 +37,26 @@ spec:
       imagePullSecrets:
         - name: {{ image_pull_secret }}
 {% endif %}
+      initContainers:
+        - name: database-check
+          image: '{{ _init_container_image }}'
+          resources:
+            requests:
+              memory: "64Mi"
+              cpu: "250m"
```
shanemcd: This seems to be the reason.

```
$ kubectl get pod automation-job-1-lj2kf -o json | jq '.status.conditions[0].message'
"0/1 nodes are available: 1 Insufficient cpu."
```

I'm going to play around with tweaking the requests in [`awx_cr_molecule.yml.j2`](https://github.com/ansible/awx-operator/blob/3c2405f304e3ef4a987b81e9e1c4c8cf50ac0e6f/molecule/default/templates/awx_cr_molecule.yml.j2), but they are already pretty low.

roles/installer/templates/postgres.yaml.j2
```diff
@@ -37,6 +37,26 @@ spec:
...
+            limits:
+              memory: "256Mi"
```
shanemcd: After looking at this a bit more closely, I dont see why we would want hardcoded limits on this container when we dont have them on any of the others. I would just remove these.

PR #780, opened 2022-02-04
Add awx pods annotations

dr-duke:

PR #807, opened 2022-02-25
added support for pod annotations to awx deployment

MrBones757:
Hello,

I recently came across a use case where I needed pod annotations on my awx deployment.
I made the changes in this PR, and then found the MR here which appears to have gone silent: https://github.com/ansible/awx-operator/pull/780

I modeled by change here after the tolerations parameter, if needed I can change this to pod_annotations as requested in the other PR, though initially this made more sense to be to align with the tolerations param. Let me know.

Cheers :)


shanemcd:
Thank you!

shanemcd (approved):

PR #501, opened 2021-08-14
Added nodeport port properties

BongoEADGC6:
This PR is in to help close issue #291 

The goal is to just allow the user to specify the nodeport port they're like to use instead of just auto-generating

```yaml
---
spec:
  ...
  service_type: NodePort
  nodeport_port: 30080
```

PR #812, opened 2022-03-08
nodeport_port should map to nodePort

fosterseth:
original PR https://github.com/ansible/awx-operator/pull/501

related: https://github.com/ansible/awx-operator/issues/291

The above PR sets `port` but that doesn't seem right. I think it should set `nodePort`


fosterseth:
closing in favor of https://github.com/ansible/awx-operator/pull/815

PR #825, opened 2022-03-11
Customize CSRF options

davidtluong:
Allow CSRF options to be customized, extra settings doesn't seem like it wouldn't work due to it being hard coded already in the configmap.  Attached to JIRA.


rooftopcellist:
@rh-dluong Before merge, this PR needs the following:

- [ ] These 2 parameters need to be added to the CRD ([example](https://github.com/ansible/awx-operator/blob/devel/config/crd/bases/awx.ansible.com_awxs.yaml#L435))
- [ ] They also needed to be added to the OLM parameters ([example](https://github.com/ansible/awx-operator/blob/devel/config/manifests/bases/awx-operator.clusterserviceversion.yaml#L557-L561))
- [ ] An addition to the README.md on usage

davidtluong:
I believe I've done all that was requested

rooftopcellist (approved):

rooftopcellist:
Looks great, thank you for the contribution @rh-dluong !

Issue #829, opened 2022-03-14
image_version can't be used alone


rguhr:
##### ISSUE TYPE
 - Feature request

##### SUMMARY

Because of current [bugs](https://github.com/ansible/awx/issues/11765) in upstream AWX we want to change the AWX version with `image_version`. But this will be ignored if `image` is not also set.

```
- name: Set user provided awx app image
  set_fact:
    _custom_image: "{{ image }}:{{ image_version }}"
  when:
    - image | default([]) | length
    - image_version is defined or image_version != ''
```
https://github.com/ansible/awx-operator/blob/devel/roles/installer/tasks/resources_configuration.yml#L42

##### ENVIRONMENT
* Operator version: 0.18.0

##### WORKAROUND
As workaround you can also defined `image` and point it to the default value:
```
[...]
spec:
  image: "{{ _image }}"
  image_version: 19.5.0
```
_(Yes, we know that only the awx version that comes bundled with the awx operator is supported)_

rguhr:
Duplicate of https://github.com/ansible/awx-operator/issues/805

Issue #805, opened 2022-02-22
Image version is ignored by operator


uznog:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
`image_version` spec is being ignored and operator is always installing 20.0.0 (or latest, perhaps).

##### ENVIRONMENT
* AWX version: 19.5.0 / 20.0.0
* Operator version: 0.17.0
* Kubernetes version: v1.21.1
* AWX install method: Kind

##### STEPS TO REPRODUCE

Kind cluster configuration:
```yaml
apiVersion: kind.x-k8s.io/v1alpha4
kind: Cluster
nodes:
- role: control-plane
  extraPortMappings:
  - containerPort: 30000
    hostPort: 30000
- role: worker
```

AWX object manifest:
```yaml
---
apiVersion: awx.ansible.com/v1beta1
kind: AWX
metadata:
  name: awx-demo
spec:
  service_type: NodePort
  nodeport_port: 30000
  image_pull_policy: Always
  image_version: 19.5.1
```

##### EXPECTED RESULTS

Operator installs version 19.5.1

##### ACTUAL RESULTS

20.0.0 is installed:
```
awx-app-74fb654997-xdzzr:       
docker.io/redis:latest
quay.io/ansible/awx:20.0.0
quay.io/ansible/awx:20.0.0
quay.io/ansible/awx-ee:latest
```

uznog:
Might be a duplicate of #794

MrBones757:
if you set image and image version it correctly accepts the values, just some re-work needed in the string generation by the looks :)
https://github.com/ansible/awx-operator/blob/devel/roles/installer/tasks/resources_configuration.yml#L39

nixocio:
hey @rooftopcellist, please, can you take a look at this when you have a minute?

kurokobo:
@uznog 
In current implementation, both `image` and `image_version` have to be specified to take effect. This behavior came with following `when` statement.
https://github.com/ansible/awx-operator/blob/fbc2d3475c36a44c3d699303882c436a455b43cb/roles/installer/tasks/resources_configuration.yml#L39-L44

So you should specify like this:

```yaml
spec:
  ...
  image: quay.io/ansible
  image_version: 19.5.1
```

But in the first place, you shoud note that changing AWX version from default is currently not supported:
> Note: The `image` and `image_version` are intended for local mirroring scenarios. Please note that using a version of AWX other than the one bundled with the `awx-operator` is not supported.
[...]

PR #833, opened 2022-03-16
Make image_version usable alone

rguhr:
Related to https://github.com/ansible/awx-operator/issues/805 & https://github.com/ansible/awx-operator/issues/829

With this PR you don't need to set `image:` anymore to manipulate the desired awx version with `image_version:`.

### DIFF
--- a/roles/installer/tasks/resources_configuration.yml
+++ b/roles/installer/tasks/resources_configuration.yml
@@ -38,9 +38,8 @@
 
 - name: Set user provided awx app image
   set_fact:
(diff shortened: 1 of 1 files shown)


rguhr:
hey @rooftopcellist the affected code was commited by you.
Do you see any dependencies that could be broken by this merge? :)

shanemcd:
I think this change would cause a bug in cases where you provide `image_version` but **not** `image`.

The `set_fact`s we're doing here always looked weird to me. I would love to revisit this approach and try to do something a little more idiomatic where we use lookups in the variables inside `defaults/main.yml`

PR #841, opened 2022-03-22
Bump dependencies stream

rooftopcellist:
We are upgrading Postgresql from 12 to 13 and as such need to handle upgrades and data migration for managed postgres deployments.  This is one potential approach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1164,11 +1164,9 @@ after a successful upgrade by setting the following variable on the AWX spec.
 
 ```yaml
   spec:
-    postgres_keep_pvc_after_upgrade: True
+    postgres_keep_pvc_after_upgrade: False
 ```
 
-# Specify whether or not to keep the old PVC after PostgreSQL upgrades
-postgres_keep_pvc_after_upgrade: True
 
 #### v0.14.0
 
--- a/README.md
+++ b/README.md
@@ -699,8 +699,8 @@ The ability to specify topologySpreadConstraints is also allowed through `topolo
 
 | Name                        | Description                         | Default |
 | --------------------------- | ----------------------------------- | ------- |
-| postgres_image              | Path of the image to pull           | 12      |
-| postgres_image_version      | Image version to pull               | 12      |
+| postgres_image              | Path of the image to pull           | postgres      |
+| postgres_image_version      | Image version to pull               | 13      |
 | node_selector               | AWX pods' nodeSelector              | ''      |
 | topology_spread_constraints | AWX pods' topologySpreadConstraints | ''      |
 | tolerations                 | AWX pods' tolerations               | ''      |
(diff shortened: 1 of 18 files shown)


roles/installer/tasks/upgrade_postgres.yml
```diff
@@ -0,0 +1,138 @@
...
+  no_log: true
+
+- name: Stream backup from pg_dump to the new postgresql container
+  k8s_exec:
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    pod: "{{ postgres_pod_name }}"
+    command: |
+      bash -c """
+      set -e -o pipefail
+      PGPASSWORD={{ awx_postgres_pass }} {{ pgdump }} | PGPASSWORD={{ awx_postgres_pass }} {{ pg_restore }}
```
dsavineau, line 96: I'm not sure if the k8s_exec module supports passing env variables but it would be great to define `PGPASSWORD` as an env var rather than on the cli (if possible)

rooftopcellist: @dsavineau unfortunately it does not support it.  If I add the following to that task, it sets it inside the operator container which is executing the role's tasks, not within the postgresql container which is where the env var is needed.  

```
  environment:
    PGPASSWORD: {{ awx_postgres_pass }}
```

README.md
```diff
@@ -1148,6 +1148,21 @@ To upgrade AWX, it is recommended to upgrade the awx-operator to the version tha
...
+#### PostgreSQL Upgrade Considerations
+
+If there is a PostgreSQL major version upgrade, after the data directory on the PVC is migrated to the new version, the old PVC is kept by default.
+This provides the ability to roll back if needed, but can take up extra storage space in your cluster unnecessarily. You can configure it to be deleted automatically
+after a successful upgrade by setting the following variable on the AWX spec. 
+
+
+```yaml
+  spec:
+    postgres_keep_pvc_after_upgrade: True
```
shanemcd: Shouldn't this example say `False`?

README.md
```diff
@@ -1148,6 +1148,21 @@ To upgrade AWX, it is recommended to upgrade the awx-operator to the version tha
...
+```
+
+# Specify whether or not to keep the old PVC after PostgreSQL upgrades
+postgres_keep_pvc_after_upgrade: True
```
shanemcd: Can these 2 line get deleted? Seems redundant.

roles/restore/tasks/secrets.yml
```diff
@@ -30,8 +30,8 @@
   block:
     - name: Set new database host
       set_fact:
-        database_host: "{{ deployment_name }}-postgres"
-      no_log: "{{ no_log }}"
+        database_host: "{{ deployment_name }}-postgres-{{ supported_pg_version }}"
+      no_log: true
```
shanemcd: Don't we want to keep `"{{ no_log }}"`?

shanemcd:
Shouldn't we replace all occurrences of `no_log: true` with `no_log: {{ no_log }}`?

rooftopcellist:
@shanemcd Yes, good catch.  I have replaced all of the no_log lines and removed the redundant doc lines.  I also changed the doc example of `postgres_keep_pvc_after_upgrade` to be the opposite of the default value as suggested.

dsavineau (approved):

PR #843, opened 2022-03-23
Add docs for proper kustomization installs

### LINKED ISSUE #840: Add docs for proper kustomization installs
Currently, the docs recommend cloning the repo and running `make deploy`, which runs the `kustomize build` step.

For users who want to provide their own kustomizations, that is possible by just creating a `kustomization.yaml` file like the following, which avoids the need to clone the repo:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
# Note the ref points to a specific git tag
- github.com/ansible/awx-operator/config/default?ref=0.18.0
- awx.yaml

namespace: mynamespace

# put patches here
```

and the file awx.yaml:

```yaml
kind: AWX
metadata:
  name: my-awx
spec:
  ...
```

then running `kustomize build . | kubectl apply -f -` (may need to run twice for the CRD to be applied).

The kustomize docs are light on details about this kind of setup, but I believe it's the preferred way to do things, almost equivalent to creating an umbrella helm chart that includes the subchart as a dependency in Chart.yaml with custom values or other resources.

This approach has been working for [gatekeeper-library](https://github.com/open-policy-agent/gatekeeper-library#kustomize) and [metallb](https://metallb.org/installation/#installation-with-kustomize).

I can provide a PR if interested.

mac-chaffee:
Closes #840 

This PR adds documentation for installing awx-operator natively with Kustomize, rather than requireing the user to clone the repo. This way, users can provide their own patches and additional resources on top of the base resources.

I also regenerated the table of contents and added a comment telling users how to regenerate it like #793 found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,14 +5,15 @@
 An [Ansible AWX](https://github.com/ansible/awx) operator for Kubernetes built with [Operator SDK](https://github.com/operator-framework/operator-sdk) and Ansible.
 
 # Table of Contents
-
+<!-- Regenerate this table of contents using https://github.com/ekalinin/github-markdown-toc -->
+<!-- gh-md-toc --insert README.md -->
 <!--ts-->
 * [AWX Operator](#awx-operator)
 * [Table of Contents](#table-of-contents)
    * [Purpose](#purpose)
    * [Usage](#usage)
-      * [Basic Install on minikube (beginner or testing)](#basic-install-on-minikube-beginner-or-testing)
-      * [Basic Install on existing cluster](#basic-install-on-existing-cluster)
+      * [Creating a minikube cluster for testing](#creating-a-minikube-cluster-for-testing)
+      * [Basic Install](#basic-install)
       * [Admin user account configuration](#admin-user-account-configuration)
       * [Network and TLS Configuration](#network-and-tls-configuration)
          * [Service Type](#service-type)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,46 +106,66 @@ Let's create an alias for easier usage:
...
+Once you have a running Kubernetes cluster, you can deploy AWX Operator into your cluster using [Kustomize](https://kubectl.docs.kubernetes.io/guides/introduction/kustomize/).
+
+First, create a file called `kustomization.yaml` with the following content:
+
+```yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  # Find the latest tag here: https://github.com/ansible/awx-operator/releases
+  - github.com/ansible/awx-operator/config/default?ref=0.18.0
```
shanemcd: Please remove the hardcoded versions and include a placeholder. This is already outdated and we've done a lot of work to avoid the need for bumping versions in files before doing a release.

shanemcd (changes requested):
Hello. Thanks for this, it's a great contribution. One small request, otherwise this all looks awesome.

kurokobo:
Thanks for PR, I have two comments.

> `kustomize build . | kubectl apply -f -`

This requires that the `kustomize` command have to be installed, but there is no instruction to install `kustomize`.
The same things can be achieved by `kubectl appky -k .` without `kustomize` command, in my opinion this is much easier than installing/using `kustomize` command. What do you think?

> Next, create a file named `awx-demo.yml` in the same folder
> ...
> ```yaml
>   # Add this extra line:
>   - awx-demo.yaml
> ```

Two filenames are not matched (`yaml`/`yml`).

mac-chaffee:
The reason I didn't recommend using the version of kustomize that comes with kubectl is because that version is typically quite outdated. Looks like the situation is improving since [kubectl v1.23 is catching up](https://github.com/kubernetes/kubectl/blob/release-1.23/go.mod#L44), but as recently as [kubectl v1.20](https://github.com/kubernetes/kubectl/blob/7873ea42e5df4038630d4251620cdb9978d132c7/go.mod#L47) is two major versions behind.

I find that the one thing harder than keeping my clusters up-to-date is making sure all my users have updated clients ;)

EDIT: oh and it's very tough to see what version of kustomize you have when using the version shipped with kubectl. You have to go to the source and look at go.mod. So quite frustrating if trying to use some newer feature of kustomize that just doesn't seem to be working.

shanemcd (approved):
Awesome contribution. Thanks a ton.

PR #862, opened 2022-04-07
Add priority class options to high priority pods

fosterseth:
Adds options for setting the priority class to the control plane and postgres pods
the priorityclasses are presumed to already exist

e.g.

```yaml
---
kind: AWX
metadata:
  name: awx-demo
spec:
  service_type: nodeport
  control_plane_priority_class: awx-demo-high-priority
  postgres_priority_class: awx-demo-medium-priority
```

also sets default requests for the postgres pod to ensure a Burstable QoS

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -187,7 +187,7 @@ spec:
                   type: array
                   items:
                     type: string
-                image_pull_secret: # deprecated
+                image_pull_secret:  # deprecated
                   description: (Deprecated) Image pull secret for app and database containers
                   type: string
                 task_resource_requirements:
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -165,6 +165,9 @@ spec:
                 control_plane_ee_image:
                   description: Registry path to the Execution Environment container image to use on control plane pods
                   type: string
+                control_plane_priority_class:
+                  description: Assign a preexisting priority class to the control plane pods
+                  type: string
                 ee_pull_credentials_secret:
                   description: Secret where pull credentials for registered ees can be found
                   type: string
@@ -391,6 +394,9 @@ spec:
                   type: object
                 postgres_storage_class:
                   description: Storage class to use for the PostgreSQL PVC
+                  type: string
+                postgres_priority_class:
+                  description: Assign a preexisting priority class to the postgres pod
                   type: string
                 postgres_data_path:
                   description: Path where the PostgreSQL data are located
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -169,18 +169,18 @@ web_command: []
 
 task_resource_requirements:
   requests:
-    cpu: 500m
-    memory: 1Gi
+    cpu: 100m
+    memory: 128Mi
 
 web_resource_requirements:
   requests:
-    cpu: 1000m
-    memory: 2Gi
+    cpu: 100m
+    memory: 128Mi
 
 ee_resource_requirements:
   requests:
-    cpu: 500m
-    memory: 1Gi
+    cpu: 100m
+    memory: 64Mi
 
 # Customize CSRF options
 csrf_cookie_secure: False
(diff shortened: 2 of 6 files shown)


roles/installer/defaults/main.yml
```diff
@@ -223,7 +226,12 @@ postgres_storage_requirements:
   requests:
     storage: 8Gi
 postgres_init_container_resource_requirements: {}
-postgres_resource_requirements: {}
+postgres_resource_requirements:
```
rooftopcellist, line 231: @fosterseth I noticed this PR after putting up this other PR that makes a similar change: https://github.com/ansible/awx-operator/pull/876

I think your default request values may be a little too high (requests are guaranteed resources, it doesn't prevent the container from using more when needed).

rooftopcellist: Let's sync up next week on how to merge these.  Proposal: if you fold in my changes in #876 and fix the conflicts, we can just merge this and close the other.

fosterseth: I pulled your commit into this PR

ikhan2010:
@shanemcd do you mind giving this a review when you get chance, please?

roles/installer/defaults/main.yml
rooftopcellist: Thank you @fosterseth 

I will pull this down and test it out again.

rooftopcellist:
Here are a couple relevant snippets from my deployment:

AWX Deployment

```
        name: awx-ee
        resources:
          requests:
            cpu: 100m
            memory: 64Mi
```

```
        name: awx-task
        resources:
          requests:
            cpu: 100m
            memory: 128Mi
```

```
        name: awx-web
        resources:
          requests:
            cpu: 100m
            memory: 128Mi
```

Postgres StatefulSet

```
        name: database-check # <------------- init container
        resources:
          requests:
            cpu: 10m
            memory: 64Mi
```

```
        name: postgres
        ports:
        - containerPort: 5432
          name: postgres
          protocol: TCP
        resources:
          requests:
            cpu: 10m
            memory: 64Mi
...
      resources:
        requests:
          storage: 8Gi

```

config/crd/bases/awx.ansible.com_awxs.yaml
```diff
@@ -392,6 +395,9 @@ spec:
                 postgres_storage_class:
                   description: Storage class to use for the PostgreSQL PVC
                   type: string
+                postgres_priority_class:
+                  description: Assign a preexisting priority class to the postgres pod
+                  type: string
```
rooftopcellist, line 400: This should be an integer - https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass

roles/installer/templates/deployment.yaml.j2
```diff
@@ -41,6 +41,9 @@ spec:
 {% for secret in image_pull_secrets %}
         - name: {{ secret }}
 {% endfor %}
+{% endif %}
+{% if control_plane_priorityclass is defined %}
```
rooftopcellist: There is a typo here: `control_plane_priority_class`

rooftopcellist:
It would be nice to have a section about setting these priority classes in the README.md.

rooftopcellist:
@fosterseth This commit adds some info to the readme and fixes the typo noted above if you want to just cherry-pick it in (https://github.com/ansible/awx-operator/commit/dd3aa99f509ebc20b4b57b251d60413985738d28, and here is the branch: https://github.com/ansible/awx-operator/compare/devel...rooftopcellist:add-to-priorityclass-option?expand=1).  
The priority classes are now properly added to the deployment and statefulset:

```
$ oc get deployment awx-demo -o yaml | grep priorityClassName:
            f:priorityClassName: {}
      priorityClassName: high-priority

$ oc get sts awx-demo-postgres -o yaml | grep priorityClassName:
            f:priorityClassName: {}
      priorityClassName: high-priority
```

rooftopcellist:
Hmm... looks like CI is failing because the pod to run the job template is timing out:

```
Error creating pod: timed out waiting for the condition
```

This may be because of that typo.  fwiw, with the commit I linked earlier, I am able to launch jobs on my deployment of awx without issue.

rooftopcellist:
Looks like CI still failed, it is because the environment that GitHub workflows runs ion does not have a lot of resources.  The new defaults the pg init container that were added were enough to make it so that the pods never got scheduled.  I also reduced the requests for the other containers to avoid running into something like this in the future.  Changes are in this commit - https://github.com/ansible/awx-operator/commit/4b6f0522e9dc32d5ac3d2b62d68a5bddaf2e37f9

rooftopcellist (approved):
Thank you for this contribution!

PR #815, opened 2022-03-09
fix: fix corrupted spec for the service with nodeport type (#814)

kurokobo:
Closes #814, closes #813, closes #504, closes #291

Tested in my K3s with values:

- `service_type: nodeport`, with/without `nodeport_port`
- `service_type: loadbalancer`
- `ingress_type: ingress`

by building new image for testing by:

```bash
# build
BUILD_ARGS="--build-arg DEFAULT_AWX_VERSION=20.0.1" \
IMAGE_TAG_BASE=registry.example.com/awx/awx-operator \
VERSION=0.19.0 make docker-build

# push
docker push registry.example.com/awx/awx-operator:0.19.0

# deploy operator
IMG=registry.example.com/awx/awx-operator:0.19.0 make deploy

# deploy awx
kubectl -n awx apply -f awx.yaml
```

PR #865, opened 2022-04-11
Check if image_pull_secrets variable is defined

rooftopcellist:
* Do not attempt to backup secret if none are defined

This is the PR that introduced the issue: https://github.com/ansible/awx-operator/pull/860


Zokormazo (approved):

rooftopcellist:
I just tested a backup without `image_pull_secrets` specified on the awx spec and it worked as expected.  Restores worked after removing the old awx instance.  It turns out there is a port conflict when using NodePort, this is an unrelated issue though due to https://github.com/ansible/awx-operator/pull/815.  It is not an issue when using Routes.

PR #868, opened 2022-04-12
Fix namespace name in readme

mac-chaffee:
I copy/pasted the command output for the kustomize install but forgot that I had been using a different namespace than then one in the kustomization.yaml file example.


fosterseth (approved):

rooftopcellist (approved):
Thank you for taking the time to put up a fix for this @mac-chaffee !

PR #871, opened 2022-04-14
Fix incorrect register reference in initialize django play

fosterseth:
minor bug, referencing the wrong register variable


rooftopcellist:
Good catch @fosterseth, but since [this PR](https://github.com/ansible/awx-operator/pull/874) merged which completely removes that task, this change is no longer needed.

PR #873, opened 2022-04-14
Fix control plane EE image parameter

### LINKED ISSUE #821: Misleading variable name: control_plane_ee_image
##### ISSUE TYPE
 - Documentation

##### SUMMARY
The name of the variable `control_plane_ee_image` suggests it would change the image for "Control Plane Execution Environment" within AWX. This seems not to be the case, it only changes the image for the bundled <resource-name>-ee container whose only purpose in life seems to be to download project updates (?).

Something is odd here. Either the updated image is not corretly displayed in AWX GUI or the variable name is misleading and should rename to something better to indicate that it has nothing to do with the "Control Plane Execution Environment" within AWX.
```
---
kind: AWX
metadata:
  name: awx-test
  namespace: awx-test
spec:
  image_pull_secret: awx-test-cp-pull-credentials
  control_plane_ee_image: <private-registry>/containers/awx-ee:latest
```
Control Plane Execution Environment within AWX

##### ENVIRONMENT
* AWX version: 20.0.0
* Operator version: 17.0.0
* Kubernetes version: v1.20.7
* AWX install method: k8s

kimbernator:
`control_plane_ee_image` is not read in until after the template that uses it has been rendered (this is done within `app_credentials.yaml.j2` which is called in "Apply Resources"). This change ensures that if said parameter is defined, it will be properly reflected in the app credentials secret.

Resolves https://github.com/ansible/awx-operator/issues/821

### DIFF
--- a/roles/installer/tasks/resources_configuration.yml
+++ b/roles/installer/tasks/resources_configuration.yml
@@ -17,6 +17,16 @@
   set_fact:
     tower_pod_name: "{{ tower_pods['resources'][0]['metadata']['name'] | default('') }}"
 
+- name: Set user provided control plane ee image
+  set_fact:
+    _custom_control_plane_ee_image: "{{ control_plane_ee_image }}"
+  when:
+    - control_plane_ee_image | default([]) | length
+
+- name: Set Control Plane EE image URL
+  set_fact:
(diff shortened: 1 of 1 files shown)


fosterseth (approved):

fosterseth:
this looks good to me if it looks good to you @rooftopcellist

rooftopcellist (approved):

rooftopcellist:
Thank you for your contribution @kimbernator !

PR #876, opened 2022-04-15
Add default resource requests for postgres containers

rooftopcellist:
### Background

When deploying the awx-operator on some cloud providers, we noticed that sometimes the UI doesn't load properly, evidenced by 404's for assets and a blank page in the UI.  What was happening is that the web container came up before the postgresql container was completely set up.  This was partially resolved by adding a more specific check for the ready status of the postgresql container in another PR.  But this problem was also seen as a result of the resource requests not being set for the postgresql containers.  

On resource constrained clusters, k8s will give priority to containers that have explicitly requested resources.  Containers that don't get squeezed, which means the container may not have enough resources to process requests, or may be terminated.  

### Solution

This PR also lowers the default requests for the other awx containers.  This is because if the cluster does not have enough resources to satisfy the requests of all of the containers in a deployment, not all of the pods will be scheduled.  Prior to this PR, our requests were unnecessarily high.

### Additional Info

These changes were based on these k8s docs: https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-resource-requests-and-limits

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,14 +434,14 @@ If you don't have access to an external PostgreSQL service, the AWX operator can
 
 The following variables are customizable for the managed PostgreSQL service
 
-| Name                                          | Description                                   | Default                           |
-| --------------------------------------------- | --------------------------------------------- | --------------------------------- |
-| postgres_image                                | Path of the image to pull                     | postgres:12                       |
-| postgres_init_container_resource_requirements | Database init container resource requirements | requests: {}                      |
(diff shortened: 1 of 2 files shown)


Zokormazo (approved):

rooftopcellist:
Closing this in favor of https://github.com/ansible/awx-operator/pull/862 which contains this commit.

PR #877, opened 2022-04-18
Restore not managed external postgresql

lutskevich:
Update postgres.yml for restore from backup not managed external postgresql db.
Current role version always failed on step [Get the postgres pod information], because pod does't exist. But backup role make it success with awx_postgres_type condition.

### DIFF
--- a/roles/restore/tasks/postgres.yml
+++ b/roles/restore/tasks/postgres.yml
@@ -27,24 +27,26 @@
     postgres_label_selector: "app.kubernetes.io/instance=postgres-{{ deployment_name }}"
   when: postgres_label_selector is not defined
 
-- name: Get the postgres pod information
-  k8s_info:
-    kind: Pod
-    namespace: '{{ ansible_operator_meta.namespace }}'
-    label_selectors:
-      - "{{ postgres_label_selector }}"
-  register: postgres_pod
-  until:
-    - "postgres_pod['resources'] | length"
-    - "postgres_pod['resources'][0]['status']['phase'] == 'Running'"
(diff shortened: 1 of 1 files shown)


rooftopcellist:
@lutskevich Thank you for your contribution!  The changes themselves look correct and will solve failing restores for external db users.  It looks like CI is complaining about some indentation and trailing whitespace. Do you mind fixing that? After that, I will happily merge this.

You can run the CI & linters locally for a faster feedback loop by running the following command in the root of your clone of the repo:

```
molecule test -s kind
```

lutskevich:
Hi @rooftopcellist, Thank you for your answer! I fixed trailing whitespaces.

roles/restore/tasks/postgres.yml
```diff
@@ -27,24 +27,25 @@
...
+        - "postgres_pod['resources'] | length"
+        - "postgres_pod['resources'][0]['status']['phase'] == 'Running'"
+        - "postgres_pod['resources'][0]['status']['containerStatuses'][0]['ready'] == true"
+      delay: 5
+      retries: 60
+    - name: Set the resource pod name as a variable.
+      set_fact:
+        postgres_pod_name: "{{ postgres_pod['resources'][0]['metadata']['name'] }}"
+  when: awx_postgres_type == 'managed'
+
```
lutskevich, line 48: no spaces here

lutskevich:
When i run the test locally it doesn't find any trailing spaces. 
I don't understand what I'm doing wrong.

lutskevich:
line 48 fixed

Issue #693, opened 2021-12-12
Nginx.conf template - IPv6 support


Noles:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Operator generated nginx.conf does not listen on IPV6.

##### ENVIRONMENT
* AWX version: -
* Operator version: devel
* Kubernetes version: -
* AWX install method: kubernetes

##### STEPS TO REPRODUCE
Kubernetes installation with IPv6 including ipv6 nginx ingress or `service_type:NodePort`
Open Webui via IPv6

##### EXPECTED RESULTS

Connect to Webui via IPv6

##### ACTUAL RESULTS

Webui not reachable

##### ADDITIONAL INFORMATION

Update the `listen` parameters in nginx.conf template does fix the problem as far as I can see.
For example `listen 8052 default_server;` to `listen [::]:8052 default_server;`

https://github.com/ansible/awx-operator/blob/35062157e0cf4d1abdc4ebf0e2cb2f941e4562c2/roles/installer/templates/config.yaml.j2#L135
https://github.com/ansible/awx-operator/blob/35062157e0cf4d1abdc4ebf0e2cb2f941e4562c2/roles/installer/templates/config.yaml.j2#L145
https://github.com/ansible/awx-operator/blob/35062157e0cf4d1abdc4ebf0e2cb2f941e4562c2/roles/installer/templates/config.yaml.j2#L155

But I'm not sure right now if this could break other things.

shanemcd:
Happy to review a PR and discuss this further.

PR #878, opened 2022-04-19
nginx now listens on ipv6

RogerSik:
See #693

I would like to to test this. How i can? With

```
kubectl -n awx exec -i -t awx-5bf89574d6-fwxwq  --container awx-web -- /bin/bash
```

i don't have root access to modify the nginx file.

### DIFF
--- a/roles/installer/templates/config.yaml.j2
+++ b/roles/installer/templates/config.yaml.j2
@@ -144,7 +144,7 @@ data:
 
         {% if route_tls_termination_mechanism | lower == 'passthrough' %}
         server {
-            listen 8052 default_server;
+            listen [::]:8052 default_server;
             server_name _;
 
             # Redirect all HTTP links to the matching HTTPS page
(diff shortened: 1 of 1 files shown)


shanemcd:
Try opening a PR on your own fork and iterating there.

RogerSik:
I created a new branch on my fork and in the kustomization.yml changed the url to my fork

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  # Find the latest tag here: https://github.com/ansible/awx-operator/releases
  - github.com/basecom/awx-operator/config/default?ref=feature/nginx-ipv6-support
  - awx.yml
```

Applying works but on my AWX instance nothing changed.

RogerSik:
@shanemcd can you help?

RogerSik:
It would be cool if we can merge this / discuss this. Because we maintain currently an IPv4 EKS cluster only for AWX for an instance which can be fixed with one line.

The line itself have we already on working state on our other Nginx projects.

roles/installer/templates/config.yaml.j2
```diff
@@ -144,7 +144,7 @@ data:
 
         {% if route_tls_termination_mechanism | lower == 'passthrough' %}
         server {
-            listen 8052 default_server;
```
RogerSik, line 147:
```
listen [::]:8052 default_server;
```

needs to be additionally not overwriting otherwise IPv4 gets disabled

astehlik:
See #950 for test instructions. This solution is working :)

PR #886, opened 2022-04-25
Render cookie settings as a boolean

fosterseth:
these values are being rendered as strings, and django is recognizing them as strings

```
>>> settings.SESSION_COOKIE_SECURE
'False'
>>> type(settings.SESSION_COOKIE_SECURE)
<class 'str'>
```

should be rendered as booleans

This was causing problems when trying to log into app, `Forbidden (CSRF cookie not set.)`


rooftopcellist (approved):
It may be worth adding in `{{ csrf_cookie_secure | bool }}`. Though this patch seems to work either way.

shanemcd:
I agree with @rooftopcellist's suggestion.

fosterseth:
thanks, included the bool filter

PR #889, opened 2022-04-26
Expect pull secret for some environments

rooftopcellist:
Some environments require pull secrets for operand containers.  The pull secret needs to be added to the Service Account of the operator pulling those operand container images.  

Unfortunately, adding the pull secret to service_account.yaml is not sufficient because the image pull secret gets lost when generating the CSV.  As a result, I have devised this hack as a work around.  The pull secrets will be patched to the SA in the reconciliation loop for the AWX CR before the images are pulled.  

With this change, it will be possible to pre-create a secret called `deployment-pull-secret` that will be automatically attached to the operator SA.


rooftopcellist:
Closing this as after more testing, it won't actually solve our problem in certain disconnected environments.

Issue #634, opened 2021-11-02
Add support for remote execution nodes


erezo9:
##### ISSUE TYPE
Question

##### SUMMARY
How do I add awx isolated nodes via the operator? is that even possible?

I want to create an instance group which will contain an isolated node that will run jobs on that node
i am able to create instance groups via the api of awx, but how do I create a node or add a node that will run the jobs?

chrismeyersfsu:
You can probably see how it is technically possible to add a remote execution node. However, we just haven't had time to bake it into the operator and make it a first-class feature.

erezo9:
@chrismeyersfsu Is there a way to edit the tower inventory file? from all the docs read 
https://docs.ansible.com/ansible-tower/latest/html/administration/external_execution_envs.html
it states i need to edit the inventory file of the ansible tower and then check if it is available, but im guessing it applies for linux envs and not container environment

chrismeyersfsu:
I see where you are coming from.

Isolated instance groups are a supported Tower/Controller only feature. I think the community has gotten them to work in the past in AWX. The code for them is all contained in AWX but we just don't have any code for standing them up.

Fast forward to today. Isolated nodes have been removed. The plan is to replace them with what we call Receptor or execution nodes. (execution nodes are nodes that only run Receptor). We've thought about how this might work in openshift/kubernetes but haven't had time to implement it. The TL;DR is that you would need to configmap the Receptor config in the operator for control plane nodes to reach out to the isolated Receptor nodes.

erezo9:
@chrismeyersfsu so do you have an example how to do it? that you think is going to be implemented in the future?

or should i move to vm installation and ditch the operator for now

chrismeyersfsu:
The operator is the only supported installation method so you'd be on your own if you try and install to a VM. I don't have an example of how to do it. I think it would involve a configuration map of the receptor config to reach out the isolated nodes running receptor. I also don't know the timeline on this feature.

erezo9:
@chrismeyersfsu do you have any docs on how i configure the receptor so i can modify the operator accordingly?

or if there is another solution for that matter

chrismeyersfsu:
@erezo9 yep! https://receptor.readthedocs.io/en/latest/connecting_nodes.html

rhessing:
@erezo9 did you find a way to get the receptors to work? We have deployed the isolated nodes with a previous version but it does not seem to work anymore with future versions.

[...]

PR #891, opened 2022-04-26
Adding option to enable a websocket listener

dale-mittleman:
This is a simple way to add a websocket listener to the AWX deployment.  Doing so will allow for the introduction of remote execution nodes.

related #634 

There are a couple of questions that need to be addressed...
- Should it be possible to expose any type of listener (TCP / UDP / websocket)?  Websocket is simplest because it can be handled most easily by an ingress controller.  Should this PR also handle the other two cases?
- How should the ingress for the websocket be provided?  In my testing, I created an additional Ingress object to handle connections.  Should an additional ingress template be added to the repo?  Or should the websocket port be available through the primary ingress?
- Do the field names follow convention?  i.e. I wasn't sure whether to name the field _enable or _enabled.

### DIFF
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -275,3 +275,8 @@ garbage_collect_secrets: false
 development_mode: false
 
 security_context_settings: {}
+
+# Enable a websocket listener on the awx-task container
+websocket_listener_enable: false
+# Websocket listener port
+websocket_listener_port: 8080
--- a/roles/installer/templates/config.yaml.j2
+++ b/roles/installer/templates/config.yaml.j2
@@ -249,7 +249,12 @@ data:
         filename: /var/run/receptor/receptor.sock
         permissions: 0660
 
+    {% if websocket_listener_enable %}    
+    - ws-listener:
+        port: {{ websocket_listener_port }}
+    {% else %}
     - local-only:
+    {% endif %}
 
     - work-command:
         worktype: local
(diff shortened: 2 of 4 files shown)


shanemcd:
Hello. Thanks for this PR. We are going to prioritize remote execution nodes pretty soon, but I'm not sure we're going to go the direction you've taken this PR. There are some use cases where we want to expose this functionality but the end user will not have direct access to the underlying infrastructure. We may end up adding the ability to connect to external execution nodes via the API, possibly the UI. We're about a month out from planning the next release of our product and this is one of the things we will be scoping out. I'll circle back here and share any documents that fall out of that effort.

sdktr:
> Hello. Thanks for this PR. We are going to prioritize remote execution nodes pretty soon, but I'm not sure we're going to go the direction you've taken this PR. There are some use cases where we want to expose this functionality but the end user will not have direct access to the underlying infrastructure. We may end up adding the ability to connect to external execution nodes via the API, possibly the UI. We're about a month out from planning the next release of our product and this is one of the things we will be scoping out. I'll circle back here and share any documents that fall out of that effort.

Any news on the 'remote execution nodes' you're referring to here? Does that envision a non receptor based approach of job delegation?

dale-mittleman:
Closing this, as the project has implemented [execution nodes](https://github.com/ansible/awx/blob/devel/docs/execution_nodes.md), preferring a mesh model where the tower node connects out to all remote execution instances

PR #720, opened 2022-01-06
added capability to set the redis container resources

Ghost0shell:
## Description
Added the capability to setup the resources requirements to the redis container (part of awx pod) 

## changes

- Added the section in the awxs.awx.ansible.com CRD
- Added the section in the appropriate location in deployment template
- Added the variable in defaults

## Tests
Tested in my own env

PR #899, opened 2022-05-03
added capability to set the redis container resources

rooftopcellist:
Direct cherry-pick of https://github.com/ansible/awx-operator/pull/720.  I just resolved the conflicts.

### DIFF
--- a/molecule/default/templates/awx_cr_molecule.yml.j2
+++ b/molecule/default/templates/awx_cr_molecule.yml.j2
@@ -27,3 +27,4 @@ spec:
       memory: 32M
   postgres_resource_requirements: {}
   postgres_init_container_resource_requirements: {}
+  redis_resource_requirements: {}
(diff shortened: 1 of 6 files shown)


roles/installer/defaults/main.yml
```diff
@@ -188,6 +188,11 @@ session_cookie_secure: False
 
 # Assign a preexisting priority class to the control plane pods
 control_plane_priority_class: ''
+
+redis_resource_requirements:
+  requests:
+    cpu: 500m
```
rooftopcellist: I lowered the minimum requested resources to schedule the redis pod to align with our other pods in a reasonable way.

molecule/default/templates/awx_cr_molecule.yml.j2
```diff
@@ -27,3 +27,4 @@ spec:
       memory: 32M
   postgres_resource_requirements: {}
   postgres_init_container_resource_requirements: {}
+  redis_resource_requirements: {}
```
rooftopcellist, line 30: I set our redis resource requirements to an empty dictionary here so that the pods can be scheduled on github workflows as there are few resources available and the pods were not getting scheduled.

daneov:
Nice, thanks for this!

PR #929, opened 2022-05-19
installer: fix undefined limit key in config

dsavineau:
When the `task_resource_requirements` variable has no "limits" key (which
is the default value) then the config template generation fails

----------------------------------
```yaml
looking for "config.yaml.j2" at "/opt/ansible/roles/installer/templates/config.yaml.j2"
File lookup using /opt/ansible/roles/installer/templates/config.yaml.j2 as file
"msg": "The task includes an option with an undefined variable. The error was: the inline if-expression on line 36 evaluated to false and no else section was defined.

The error appears to be in /opt/ansible/roles/installer/tasks/resources_configuration.yml: line 30, column 3, but may

- name: Apply Resources
 ^ here
```
----------------------------------

The current condition doesn't have a else statement so the template fails
when the "limits" key isn't present.
This rewrite the current if/else statement in jinja template.

Signed-off-by: Dimitri Savineau <dsavinea@redhat.com>

### DIFF
--- a/roles/installer/templates/config.yaml.j2
+++ b/roles/installer/templates/config.yaml.j2
@@ -32,15 +32,15 @@ data:
 
     IS_K8S = True
 
+{% if "limits" in task_resource_requirements and "memory" in task_resource_requirements["limits"] %}
     # Set memory available based off of resource request/limit for the task pod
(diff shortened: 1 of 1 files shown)


rooftopcellist (approved):

Zokormazo (approved):

Zokormazo:
@kdelee @jainnikhil30 . For that thing we talked in https://github.com/ansible/awx-operator/pull/976

kdelee (approved):
:ship:

PR #944, opened 2022-06-10
Option to delete backup dir on PV when AWXBackup object is deleted

viv-dev:
This allows external tools better management of the number of backups stored on a PV where the same volume is used for every export and the volume itself is periodically backed up e.g. to S3 using Rancher.

Mechanism was implemented by setting a Finalizer on the AWXBackup object which runs the backup role with `finalizer_run` variable set to true. This then checks if there is a property called `clean_backup_on_delete` on the AWXBackup object being deleted, and if true, then fires up the backup pod and removes the directory on the PV.

### DIFF
--- a/config/crd/bases/awxbackup.ansible.com_awxbackups.yaml
+++ b/config/crd/bases/awxbackup.ansible.com_awxbackups.yaml
@@ -43,6 +43,9 @@ spec:
                 backup_storage_class:
                   description: Storage class to use when creating PVC for backup
                   type: string
+                clean_backup_on_delete:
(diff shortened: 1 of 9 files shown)


shanemcd:
Thanks for this addition and nice refactoring. Please resolve the conflict and we'll get this merged soon.

viv-dev:
Hey, thanks!

I've now resolved the conflicts and rebased my branch so it should be good to go when the checks pass.

shanemcd (approved):
Amazing work, thank you! 🥇

Issue #937, opened 2022-06-03
ldap_password_secret fails to work on deployment


leasley199:
##### ISSUE TYPE
 - Unable to use secret to inject LDAP Bind Password.

##### SUMMARY
Unable to inject ldap password into awx deployment via secret, ldap authentication fails. 

##### ENVIRONMENT
* AWX version: 21.1.0
* Operator version: 0.22.0
* Kubernetes version: 1.21.0~
* AWX install method: helm awx operator

##### STEPS TO REPRODUCE

(https://github.com/ansible/awx-operator#enabling-ldap-integration-at-awx-bootstrap)
Create a secret for the LDAP Bind user prior to standing up the awx stack
```
apiVersion: v1
kind: Secret
metadata:
  name: awx-ldap-password
data:
  ldap-password: <VALUE_HERE>
```
Stand-up and source the created secret. 
```
apiVersion: awx.ansible.com/v1beta1
kind: AWX
...
spec:
  ldap_cacert_secret: awx-ldap-password
```

##### EXPECTED RESULTS

Expected LDAP Authentication to work.

##### ACTUAL RESULTS

LDAP login fails. 

##### ADDITIONAL INFORMATION

Can get the LDAP working if you add the value <code>AUTH_LDAP_BIND_PASSWORD</code>
to the awx.spec. value 

for example: 
```
apiVersion: awx.ansible.com/v1beta1
kind: AWX
...
spec:
  - setting: AUTH_LDAP_BIND_PASSWORD
    value: 'VALUE_HERE'
```

##### AWX-OPERATOR LOGS
```
2022-06-03 19:06:10,681 ERROR    [4f4822bf266e496fa3279f7ce63b9f81] django_auth_ldap search_s('DC=example,DC=com', 2, '(userprincipalname=username@lhi.com)') raised OPERATIONS_ERROR({'msgtype': 101, 'msgid': 2, 'result': 1, 'desc': 'Operations error', 'ctrls': [], 'info': '000004DC: LdapErr: DSID-0C0907E9, comment: In order to perform this operation a successful bind must be completed on the connection., data 0, v2580'})

2022-06-03 19:06:10,681 DEBUG    [4f4822bf266e496fa3279f7ce63b9f81] django_auth_ldap search_s('DC=example,DC=com', 2, '(userprincipalname=%(user)s)') returned 0 objects:
```

fosterseth:
[...]

PR #659, opened 2021-11-19
Load LDAP password from secret and update guideline

hungtran84:
This PR is created for 2 things:
- Load LDAP bind password from k8s secret instead of extra settings due to security concern
- Update README with the instructions and sample of LDAP integration via extra settings
- Also add missing LDAPSearch module

Already tested in Azure AKS with a self-hosted windows Active Directory.

The most important things is to have LDAP integration with AWX deployment without any human intervention require and have LDAP setting checked in code.

PR #955, opened 2022-06-20
Fix ldap bind password secret usage: ldap_password_secret

tu-doan:
## Overview
This PR is created to fixed the missing configuration of the previous feature by @hungtran84 : [PR-659](https://github.com/ansible/awx-operator/pull/659)

Related issue: https://github.com/ansible/awx-operator/issues/937  

## Description
The `ldap_password_secret` will not take effect unless it is added into settings.py. Previous owner of this configuration has developed to read the password from k8s secret, but sadly, it is not used anywhere in the code.

## Note
If this pr can be approved and merged, can you please add this to version v0.21.1. Otherwise, the configuration `ldap_password_secret` in v0.21.0 is useless.

### DIFF
--- a/roles/installer/templates/config.yaml.j2
+++ b/roles/installer/templates/config.yaml.j2
@@ -101,6 +101,11 @@ data:
     BROADCAST_WEBSOCKET_PORT = 8052
     BROADCAST_WEBSOCKET_PROTOCOL = 'http'
 
+    # Load LDAP BIND password from Kubernetes secret if define
+    {% if ldap_password_secret -%}
+    AUTH_LDAP_BIND_PASSWORD = "{{ ldap_bind_password }}"
+    {% endif %}
+
 {% for item in extra_settings | default([]) %}
     {{ item.setting }} = {{ item.value }}
 {% endfor %}
--- a/roles/installer/templates/config.yaml.j2
+++ b/roles/installer/templates/config.yaml.j2
@@ -101,11 +101,6 @@ data:
     BROADCAST_WEBSOCKET_PORT = 8052
     BROADCAST_WEBSOCKET_PROTOCOL = 'http'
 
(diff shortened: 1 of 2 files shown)


hungtran84:
@doanminhtu thanks for the fix

roles/installer/templates/config.yaml.j2
```diff
@@ -101,6 +101,11 @@ data:
     BROADCAST_WEBSOCKET_PORT = 8052
     BROADCAST_WEBSOCKET_PROTOCOL = 'http'
 
+    # Load LDAP BIND password from Kubernetes secret if define
+    {% if ldap_password_secret -%}
+    AUTH_LDAP_BIND_PASSWORD = "{{ ldap_bind_password }}"
```
shanemcd: Hello and thanks for your contribution. Unfortunately it won't be _quite_ this easy. ConfigMaps should not be used to store sensitive information because they are not encrypted.

Luckily we already have a pattern for this. Check out these files:

https://github.com/ansible/awx-operator/blob/ebe5d1e2f4afd7fd329c18420947ff5840670293/roles/installer/templates/app_credentials.yaml.j2

https://github.com/ansible/awx-operator/blob/ebe5d1e2f4afd7fd329c18420947ff5840670293/roles/installer/templates/credentials.py.j2

shanemcd (changes requested):

tu-doan:
Hi @shanemcd , Thanks for your fast reply and comment. I have updated the PR.
I got the concern about secret in settings. Your suggestion on `credentials.py` is very helpful. However, I think that this `AUTH_LDAP_BIND_PASSWORD` is suitable to be located in `ldap.py`, which is also secret and managed by `app_credentials.yaml`.

shanemcd (approved):
Great work here @doanminhtu!

Issue #936, opened 2022-06-03
Is there a way to protect an AWX instance from automatically upgrading when the operator gets upgraded?


Cl0udius:
##### ISSUE TYPE
 - Question
 
##### SUMMARY
Is there a way to protect an AWX instance from automatically upgrading when the operator gets upgraded?

##### ADDITIONAL INFORMATION

We currently have two AWX instances in our K8S cluster installed with the AWX operator. 
After the operator upgrade all instances all existing AWX instances PODs get restartet even when we specify all images versions inside the AWX object.
This means we have a short but unwanted downtime for all AWX instances. 
We would like to have the possibility to upgrade those instances step by step and not all at once.

Is there some kind of tag we can set at the pod or deployment object of the existing instance to "protect" it from automatically updates?
As i could see it in the code the deployment template will be applied in any case. 
Maybe a conditional check on this task could solve our problem here.

##### ENVIRONMENT
* AWX version: 19.4.0
* Operator version: 0.19.0
* Kubernetes version: 1.23.6

vrubiolo:
Hi,

Just a question here since AFAIK the operator is namespace scoped: could you just upgrade one operator and not the other? I would expect that only one AWX instance would get upgraded, not the other (but I might be wrong here, I have only one instance in my cluster).

Cl0udius:
Hi,

i also got such thing in mind but we may could run into problems when there is a CRD update with the new operator version, as the CRDs are not namespace bound AFAIK. So i am not sure if this could lead to side effects.

BR

fosterseth:
if you look at the operator logs, what are the results for the `Apply deployment resources` task https://github.com/ansible/awx-operator/blob/0983220fba279a76a6d39bc22cd4e86959c34c5a/roles/installer/tasks/resources_configuration.yml#L75

does it say changed?

Cl0udius:
Hi,

i could not really get it from the logs. I found some tasks which were changed with this


```
TASK [Apply Resources] ********************************
changed: [localhost] => (item=None) => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": true}
```

The task you mention had the following output:

```
TASK [installer : Apply deployment resources] **********************************
task path: /opt/ansible/roles/installer/tasks/resources_configuration.yml:75

-------------------------------------------------------------------------------
[...]

PR #956, opened 2022-06-21
added auto_upgrade flag

Cl0udius:
Adds the possibility to set an auto_upgrade flag (true|false) to optionally prevent automatic AWX instance upgrades when the operator gets upgraded.

AWX will be upgraded if a deployment exists and the flag is set to true or no deployment exits.
With this we can set instances to "auto_upgrade: false" and recreate the deployment of multiple AWX instances step by step instead of doing it all at once.

Related issue: https://github.com/ansible/awx-operator/issues/936

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ An [Ansible AWX](https://github.com/ansible/awx) operator for Kubernetes built w
          * [Session Cookie Secure Setting](#session-cookie-secure-setting)
          * [Extra Settings](#extra-settings)
          * [Configure no_log](#no-log)
+         * [Auto Upgrade](#auto-upgrade)
+            * [Upgrade of instances without auto upgrade](#upgrade-of-instances-without-auto-upgrade)
          * [Service Account](#service-account)
       * [Uninstall](#uninstall)
       * [Upgrading](#upgrading)
@@ -1033,6 +1035,42 @@ Example configuration of `no_log` parameter
   spec:
     no_log: 'true'
 ```
+
+#### Auto upgrade
+With this parameter you can influence the behaviour during an operator upgrade.  
+If set to `true`, the operator will upgrade the specific instance directly.  
+When the value is set to `false`, and we have a running deployment, the operator will not update the AWX instance.  
+This can be useful when you have multiple AWX instances which you want to upgrade step by step instead of all at once.  
+
+
+| Name         | Description                        | Default |
+| -------------| ---------------------------------- | ------- |
+| auto_upgrade | Automatic upgrade of AWX instances | true    |
+
+Example configuration of `auto_upgrade` parameter
+
+```yaml
+  spec:
+    auto_upgrade: true
+```
+
+##### Upgrade of instances without auto upgrade
+
+There are two ways to upgrade instances which are marked with the 'auto_upgrade: false' flag.  
+
+Changing flags:
+
+- change the auto_upgrade flag on your AWX object to true  
+- wait until the upgrade process of that instance is finished
+- change the auto_upgrade flag on your AWX object back to false  
+
+Delete the deployment:
+
+- delete the deployment object of your AWX instance  
+```
+$ kubectl -n awx delete deployment <yourInstanceName> 
+```
+- wait until the instance gets redeployed  
 
 
 #### Service Account
--- a/roles/installer/tasks/main.yml
+++ b/roles/installer/tasks/main.yml
@@ -1,88 +1,13 @@
 ---
-- name: Patching labels to AWX kind
-  k8s:
-    state: present
-    definition:
-      apiVersion: '{{ api_version }}'
-      kind: '{{ kind }}'
-      name: '{{ ansible_operator_meta.name }}'
-      namespace: '{{ ansible_operator_meta.namespace }}'
-      metadata:
-        name: '{{ ansible_operator_meta.name }}'
-        namespace: '{{ ansible_operator_meta.namespace }}'
-        labels:
-          app.kubernetes.io/name: '{{ ansible_operator_meta.name }}'
-          app.kubernetes.io/part-of: '{{ ansible_operator_meta.name }}'
-          app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
-          app.kubernetes.io/component: '{{ deployment_type }}'
-          app.kubernetes.io/operator-version: '{{ lookup("env", "OPERATOR_VERSION") }}'
-
-- name: Include secret key configuration tasks
-  include_tasks: secret_key_configuration.yml
-
-- name: Load LDAP CAcert certificate
-  include_tasks: load_ldap_cacert_secret.yml
-  when:
-    - ldap_cacert_secret != ''
-
-- name: Load ldap bind password
-  include_tasks: load_ldap_password_secret.yml
-  when:
-    - ldap_password_secret != ''
-
-- name: Load bundle certificate authority certificate
-  include_tasks: load_bundle_cacert_secret.yml
-  when:
-    - bundle_cacert_secret != ''
-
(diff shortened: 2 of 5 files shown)


roles/installer/tasks/main.yml
```diff
@@ -1,88 +1,101 @@
...
-      metadata:
+- name: Check for presence of Deployment
+  k8s_info:
+    api_version: v1
+    kind: Deployment
+    name: "{{ ansible_operator_meta.name }}"
+    namespace: "{{ ansible_operator_meta.namespace }}"
+  register: tower_deployment
+
+- block:
```
fosterseth: instead of this `block`, might be best to split this out to another task. That is, move everything that is in your block into a `install.yml` and have the `Check for presence of Deployment` still in main.yml. then include and run the install.yml task if your auto_update condition is true.

fosterseth:
@shanemcd @rooftopcellist how do we feel about this `auto_upgrade` boolean? I definitely see the use case. That is, not bumping the awx pods and leading to application downtime only because the operator changed.

auto_upgrade might not be the best term, maybe `auto_redeploy`? If the user hard-codes the AWX version to use, they aren't really upgrading per-se, rather just bumping the awx pods.

roles/installer/tasks/main.yml
```diff
@@ -1,88 +1,101 @@
...
+    include_tasks: initialize_django.yml
+
+  - name: Update status variables
+    include_tasks: update_status.yml
+
+  - name: Cleanup & Set garbage collection refs
+    include_tasks: cleanup.yml
+
+  # Just execute deployment steps when auto_upgrade is true or when no deployment exists
+  when: (tower_deployment['resources'] | length > 0 and auto_upgrade == true) or (tower_deployment['resources'] | length == 0)
```
shanemcd: Please add `auto_upgrade` to `roles/installer/defaults/main.yml`

shanemcd:
I think I'm fine with both the idea and the naming of the variable. Even if you hardcode the version of the image, other stuff in the yaml might have changed too.

rooftopcellist:
@Cl0udius With this PR, if you set auto_upgrade to `false`, what exactly would the "manual upgrade" steps be?  Could we add that to the README.md?  

My guess is that for each instance, you would need to modify the AWX spec to: 
1. Temporarily change auto_upgrade to `true`, which will trigger the reconciliation loop for just that instance.  
2. Allow the reconciliation loop to complete, which will run the install.yml, and update the deployment yaml to point at the new AWX image.
3. Containers will cycle picking up the new AWX version.
4. Change auto_upgrade back to `true` on the spec.  The reconciliation loop should one one more time, but will be a no-op.

^^ Is that what you had in mind?

Cl0udius:
> @Cl0udius With this PR, if you set auto_upgrade to `false`, what exactly would the "manual upgrade" steps be? Could we add that to the README.md?
> 
> My guess is that for each instance, you would need to modify the AWX spec to:
> 
>     1. Temporarily change auto_upgrade to `true`, which will trigger the reconciliation loop for just that instance.
> 
>     2. Allow the reconciliation loop to complete, which will run the install.yml, and update the deployment yaml to point at the new AWX image.
> 
>     3. Containers will cycle picking up the new AWX version.
> 
>     4. Change auto_upgrade back to `true` on the spec.  The reconciliation loop should one one more time, but will be a no-op.
> 
> 
> ^^ Is that what you had in mind?

Hi,

that would be one way. Another thing we could do is to leave the flag on value false and just remove the deployment from K8S which will also trigger the loop and will rollout the new version.

I can add both ways if this would help.

BR

roles/installer/tasks/main.yml
Cl0udius: moved to install.yml

roles/installer/tasks/main.yml
Cl0udius: added to defaults

README.md
```diff
@@ -1034,6 +1036,23 @@ Example configuration of `no_log` parameter
     no_log: 'true'
 ```
 
+#### Auto upgrade
+With this parameter you can influence the behaviour during an operator upgrade.  
+If set to true the operator will upgrade the specific instance directly.  
+When the value is set to false and we have a running deployment the operator will not update the AWX instance.  
+This can be usefull when you have multiple AWX instances which you want to upgrade step by step instead all at once.
```
shanemcd:
```suggestion
This can be useful when you have multiple AWX instances which you want to upgrade step by step instead all at once.  
```
```suggestion
This can be usefull when you have multiple AWX instances which you want to upgrade step by step instead all at once.  
```

README.md
```diff
@@ -1034,6 +1036,23 @@ Example configuration of `no_log` parameter
     no_log: 'true'
 ```
 
+#### Auto upgrade
+With this parameter you can influence the behaviour during an operator upgrade.  
+If set to true the operator will upgrade the specific instance directly.
```
shanemcd:
```suggestion
If set to `true`, the operator will upgrade the specific instance directly.  
```
```suggestion
If set to true the operator will upgrade the specific instance directly.  
```

README.md
```diff
@@ -1034,6 +1036,23 @@ Example configuration of `no_log` parameter
     no_log: 'true'
 ```
 
+#### Auto upgrade
+With this parameter you can influence the behaviour during an operator upgrade.  
+If set to true the operator will upgrade the specific instance directly.  
+When the value is set to false and we have a running deployment the operator will not update the AWX instance.
```
shanemcd:
```suggestion
When the value is set to `false` and we have a running deployment, the operator will not update the AWX instance.  
```

README.md
shanemcd:
```suggestion
This can be usefull when you have multiple AWX instances which you want to upgrade step by step instead of all at once.  
```

roles/installer/tasks/main.yml
```diff
@@ -1,88 +1,13 @@
...
-- name: Update status variables
-  include_tasks: update_status.yml
+  register: tower_deployment
 
-- name: Cleanup & Set garbage collection refs
-  include_tasks: cleanup.yml
+# Just execute deployment steps when auto_upgrade is true or when no deployment exists
+- name: Start installation
+  include_tasks: install.yml
+  when: (tower_deployment['resources'] | length > 0 and auto_upgrade == true) or (tower_deployment['resources'] | length == 0)
```
shanemcd: You probably need to cast this to a boolean:

```suggestion
  when: (tower_deployment['resources'] | length > 0 and auto_upgrade | bool) or (tower_deployment['resources'] | length == 0)
```

README.md
```diff
@@ -1034,6 +1036,23 @@ Example configuration of `no_log` parameter
...
+
+| Name         | Description                        | Default |
+| -------------| ---------------------------------- | ------- |
+| auto_upgrade | Automatic upgrade of AWX instances | 'true'  |
+
+Example configuration of `auto_upgrade` parameter
+
+```yaml
+  spec:
+    auto_upgrade: 'true'
```
shanemcd: Is this intentionally represented as a string here?

README.md
```diff
@@ -1095,6 +1114,25 @@ Then install the new AWX Operator by following the instructions in [Basic Instal
 
 Once the new AWX Operator is up and running, your AWX deployment will also be upgraded.
 
+##### Upgrade of instances with auto upgrade
```
shanemcd: I would move this up under the `#### Auto upgrade` section.

Cl0udius: Moved like suggested

README.md
Cl0udius: Nope. Changed now.

shanemcd (approved):
I think people will like this feature. Thank you! ⭐

PR #959, opened 2022-06-22
Use specific version of redis image

mac-chaffee:
While redis has good backwards-compatability guarantees, it would be risky not to pin at least the major version of redis to a specific number. That way a pod restart won't silently upgrade someone's redis installation.

Additionally, some cluster admins ban the use of the "latest" tag: https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/disallowedtags/samples/container-image-must-not-have-latest-tag/constraint.yaml

### DIFF
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -127,7 +127,7 @@ extra_volumes: ''
 _image: quay.io/ansible/awx
 _image_version: "{{ lookup('env', 'DEFAULT_AWX_VERSION') or 'latest' }}"
 _redis_image: docker.io/redis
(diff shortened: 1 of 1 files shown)


mac-chaffee:
I think we might want to explore using non-root images like https://github.com/bitnami/bitnami-docker-redis , and maybe pinning the SHA as well in the future.

shanemcd (approved):
I'm fine with the change you've made here but I'm against pinning to a specific SHA. We'd constantly be changing it as CVEs pop up in the base image and we're not staffed to take on that level of maintenance for this project.

PR #967, opened 2022-06-27
Add pg password encrypt functionality

davidtluong:
In the following KCS, we allow the passing of encrypted postgres password in credentials.py:
https://access.redhat.com/solutions/4309941

In the operator, this is not possible at the moment as any changes to that file is not persistent.  I have added a variable to toggle adding the `from awx.main.utils import decrypt_value, get_encryption_key` in order to enable the functionality.  By default, I have disabled it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -463,7 +463,9 @@ stringData:
 type: Opaque
 ```
 
-> Please ensure that the value for the variable `password` should _not_ contain single or double quotes (`'`, `"`) or backslashes (`\`) to avoid any issues during deployment, backup or restoration.
+> Please ensure that the value for the variable `password` should _not_ contain single or double quotes (`'`, `"`) or backslashes (`\`) to avoid any issues during deployment, backup or restoration.  
+
+> In order to use an encrypted value for the password, it needs to be formatted like the following: $encrypted$AESCBC$Z0FBQUFBQmNONU9BbGQ1VjJyNDJRVTRKaFRIR09Ib2U5TGdaYVRfcXFXRjlmdmpZNjdoZVpEZ21QRWViMmNDOGJaM0dPeHN2b194NUxvQ1M5X3dSc1gxQ29TdDBKRkljWHc9PQ==
 
 > It is possible to set a specific username, password, port, or database, but still have the database managed by the operator. In this case, when creating the postgres-configuration secret, the `type: managed` field should be added.
 
--- a/roles/installer/templates/credentials.py.j2
+++ b/roles/installer/templates/credentials.py.j2
@@ -1,10 +1,18 @@
+{% if awx_postgres_pass.startswith('$encrypted$') | bool %}
+from awx.main.utils import decrypt_value, get_encryption_key
+
+{% endif %}
 DATABASES = {
     'default': {
         'ATOMIC_REQUESTS': True,
         'ENGINE': 'awx.main.db.profiled_pg',
         'NAME': "{{ awx_postgres_database }}",
         'USER': "{{ awx_postgres_user }}",
+{% if awx_postgres_pass.startswith('$encrypted$') | bool %}
+        'PASSWORD': decrypt_value(get_encryption_key('value'),'"{{ awx_postgres_pass }}"'),
+{% else %}
         'PASSWORD': "{{ awx_postgres_pass }}",
+{% endif %}
         'HOST': '{{ awx_postgres_host }}',
         'PORT': "{{ awx_postgres_port }}",
         'OPTIONS': { 'sslmode': '{{ awx_postgres_sslmode }}',
--- a/roles/backup/tasks/postgres.yml
+++ b/roles/backup/tasks/postgres.yml
@@ -24,6 +24,39 @@
   no_log: "{{ no_log }}"
 
 - block:
+    - name: Get the controller pod information
+      k8s_info:
+        kind: Pod
+        namespace: '{{ ansible_operator_meta.namespace }}'
+        label_selectors:
+          - "app.kubernetes.io/name = {{ deployment_name }}"
+      register: controller_pod
+      until:
+        - "controller_pod['resources'] | length"
+        - "controller_pod['resources'][0]['status']['phase'] == 'Running'"
+        - "controller_pod['resources'][0]['status']['containerStatuses'][0]['ready'] == true"
+      delay: 5
+      retries: 60
+
+    - name: Get decrypted variable
+      k8s_exec:
+        namespace: '{{ ansible_operator_meta.namespace }}'
+        pod: "{{ controller_pod.resources[0].metadata.name }}"
+        container: "{{ deployment_name }}-task"
+        command: |
+          awx-manage shell_plus -c """
+          from awx.main.utils import decrypt_value, get_encryption_key
+          decrypted_value = decrypt_value(get_encryption_key('value'),'{{postgresql_password}}')
+          print(decrypted_value)
+          """
+      register: decrypt_postgres_pass
+
+    - name: Set decrypted postgres pass as awx_postgres_pass
+      set_fact: 
+        awx_postgres_pass: "{{ decrypt_postgres_pass.stdout_lines[1] }}"
+  when: "'encrypted' in awx_postgres_pass"
+
+- block:
     - name: Delete pod to reload a resource configuration
       set_fact:
         postgres_label_selector: "app.kubernetes.io/instance=postgres-{{ deployment_name }}"


Zokormazo:
I wonder if we should we enable this option or defer to `Secret` encryption in the underlying k8s cluster to make this happen?

davidtluong:
Secret encryption would still leave it unencrytped in the credentials.py when mounted wouldn't it?

Also, when using this, I believe it'll only work with external databases, otherwise that `decrypt_value` function will be inserted into the postgres password for our postgresql deployment

Zokormazo:
> Secret encryption would still leave it unencrytped in the credentials.py when mounted wouldn't it?
> 
> Also, when using this, I believe it'll only work with external databases, otherwise that `decrypt_value` function will be inserted into the postgres password for our postgresql deployment

Oh true, you are right, nvm

README.md
```diff
@@ -1038,6 +1038,20 @@ Example configuration of `no_log` parameter
     no_log: 'true'
 ```
 
+#### Postgres Secret Encrypt
+Configure to enable storing encrypted postgres password
+
+| Name              | Description                   | Default |
+| ----------------- | ----------------------------- | ------- |
+| pg_secret_encrypt | Enable PG password encryption | 'false' |
```
dsavineau:
```suggestion
| pg_secret_encrypt | Enable PG password encryption | false |
```

README.md
```diff
@@ -1038,6 +1038,20 @@ Example configuration of `no_log` parameter
...
+
+Example configuration of `pg_secret_encrypt` parameter
+
+```yaml
+  spec:
+    pg_secret_encrypt: 'true'
```
dsavineau:
```suggestion
    pg_secret_encrypt: true
```

config/crd/bases/awx.ansible.com_awxs.yaml
```diff
@@ -485,6 +485,9 @@ spec:
                 no_log:
                   description: Configure no_log for no_log tasks
                   type: string
+                pg_encrypt_secret:
+                  description: Configure postgres password encryption functionality
+                  type: string
```
dsavineau:
```suggestion
                  type: boolean
                  default: false
```

config/manifests/bases/awx-operator.clusterserviceversion.yaml
```diff
@@ -579,6 +579,11 @@ spec:
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:com.tectonic.ui:hidden
+      - displayName: Postgresql Password Encryption
+        path: pg_secret_encrypt
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:advanced
+        - urn:alm:descriptor:com.tectonic.ui:hidden
```
dsavineau: We probably want to use `urn:alm:descriptor:com.tectonic.ui:booleanSwitch` with the boolean type

roles/installer/defaults/main.yml
```diff
@@ -288,3 +288,6 @@ no_log: 'true'
 # Should AWX instances be automatically upgraded when operator gets upgraded
 #
 auto_upgrade: true
+
+# Set pg_secret_encrypt if planning to use encrypted pg password
+pg_secret_encrypt: 'false'
```
```suggestion
pg_secret_encrypt: false
```

roles/installer/templates/credentials.py.j2
```diff
@@ -1,3 +1,7 @@
+{% if pg_secret_encrypt == "true" %}
```
dsavineau:
```suggestion
{% if pg_secret_encrypt | bool %}
```

roles/installer/templates/credentials.py.j2
```diff
@@ -1,3 +1,7 @@
...
+from awx.main.utils import decrypt_value, get_encryption_key
+
```
dsavineau, line 3: We should avoid adding a blank line here so we don't modify the credentials.py for existing deployment.

dsavineau (changes requested):
Except the string vs boolean type thing, the change doesn't seem enough to support encrypted value.

We still need to use the `decrypt_value` and `get_encryption_key` functions in the credentials.py.j2 template to get the password value.
This means that we need a new condition in the template for the the `PASSWORD` key/value

```diff
+{% if pg_secret_encrypt | bool %}
+        'PASSWORD': decrypt_value(get_encryption_key('value'),"{{ awx_postgres_pass }}"),
+{% else %}
         'PASSWORD': "{{ awx_postgres_pass }}",
+{% endif %}
```

We could also add an assert to check if the encrypted password value has the right format

```yaml
    - ansible.builtin.assert:
        that:
          - awx_postgres_pass.startswith('$encrypted$')
        fail_msg: 'encrypted awx_postgres_pass value should start with $encrypted$'
      when: pg_secret_encrypt | bool
```

As an alternate implementation, we could also drop the new `pg_secret_encrypt` variable and only rely on the `awx_postgres_pass` value (ie: if the string starts with $encrypted$) to detect if it's encrypted or not.

```diff
--- a/credentials.py.j2
+++ b/credentials.py.j2
@@ -1,10 +1,17 @@
+{% if awx_postgres_pass.startswith('$encrypted$') | bool %}
+from awx.main.utils import decrypt_value, get_encryption_key
+{% endif %}
 DATABASES = {
     'default': {
         'ATOMIC_REQUESTS': True,
         'ENGINE': 'awx.main.db.profiled_pg',
         'NAME': "{{ awx_postgres_database }}",
         'USER': "{{ awx_postgres_user }}",
+{% if awx_postgres_pass.startswith('$encrypted$') | bool %}
+        'PASSWORD': decrypt_value(get_encryption_key('value'),"{{ awx_postgres_pass }}"),
+{% else %}
         'PASSWORD': "{{ awx_postgres_pass }}",
+{% endif %}
         'HOST': '{{ awx_postgres_host }}',
         'PORT': "{{ awx_postgres_port }}",
         'OPTIONS': { 'sslmode': '{{ awx_postgres_sslmode }}',
```

The template change would be the only required change in this case.

davidtluong:
For your concern here:
```
We still need to use the decrypt_value and get_encryption_key functions in the credentials.py.j2 template to get the password value.
```

I figured the customer would've copied the whole value of that into the PG_PASSWORD field, for example, I would imagine their secret looks like this:
```
apiVersion: v1
data:
...
  password: decrypt_value(get_encryption_key('value'),'$encrypted$AESCBC$Z0FBQUFBQmNONU9BbGQ1VjJyNDJRVTRKaFRIR09Ib2U5TGdaYVRfcXFXRjlmdmpZNjdoZVpEZ21QRWViMmNDOGJaM0dPeHN2b194NUxvQ1M5X3dSc1gxQ29TdDBKRkljWHc9PQ==')
```

As for the boolean, I can change it no problem, I just thought since no_log was configured the same way that this would carry over as well.  

As for the blank line, would that still show up if the condition is false?

dsavineau:
> I figured the customer would've copied the whole value of that into the PG_PASSWORD field, for example, I would imagine their secret looks like this:

This won't work because the `awx_postgres_pass` variable is quoted in the credentials.py.j2 [1] so the `decrypt_value` and `get_encryption_key` functions won't be executed.

> As for the blank line, would that still show up if the condition is false?

it won't show up, I don't remember why I put that comment, so you can dismiss it

[1] https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/credentials.py.j2#L7

roles/installer/templates/credentials.py.j2
davidtluong: I went with just doing a starts with as suggested instead of making a whole new variable and changing the readme to reflect that.

rooftopcellist:
So let me see if I have this right:
* The goal here is for the user to have a deployment that does not have the PG Password exposed in plaintext in the credentials.py file inside the task and web containers.  For VM installs, we have the KCS article, but that won't work in openshift. 

For operator deployments, to have this value encrypted in the resulting credentials.py file with this PR, the user needs to:
1. Create an AutomationController CR with a pg config secret using the unencrypted password
2. Exec into the container and use the symmetric key (SECRET_KEY) unique to that deployment to encrypt the PG password using awx-manage and the steps in the KCS article.
3. Modify the pg secret to include the encrypted value
4. Restart the deployment: oc rollout deployment <deployment-name>
5. It will re-create the credentials.py file mounted into the container with the encrypted PG password.

There are currently [3 other places](https://github.com/ansible/awx-operator/search?q=awx_postgres_pass) where the awx_postgres_pass is used, the backup and restore roles, and the migrate_data.yml logic.  

Once the pg 12 --> 13 upgrade logic lands that will be another.  The backup and restore ones are not a problem, but the upgrade and migrate data ones are because the PG_PASSWORD value must be decrypted by the operator before being used.  

So, while the awx-task container is still up, we will need to have a k8s_exec task that will decrypt it and register the output as a variable.  @rh-dluong

davidtluong:
So, I'm wracking my brain and it seems there's a bit of a hitch with this.  I'm assume the way the decrypt works is that we use the SECRET_KEY to decrypt, however, due to these migrate_data task(s), this kinda shoots down this whole thing.

I've tested this, and it does decrypt the var successfully as a mockup:
```
- name: Set old postgres password if encrypted
  kubernetes.core.k8s_exec:
    namespace: dluong-aap
    pod: dluong-controller-5c5b7477c5-hq27x
    container:  dluong-controller-task
    command: |
      awx-manage shell_plus -c """
      from awx.main.utils import decrypt_value, get_encryption_key
      decrypted_value = decrypt_value(get_encryption_key('value'),'{{postgresql_password}}')
      print(decrypted_value)
      """
  register: decrypt_old_postgres
  when: "'encrypted' in postgresql_password"

- name: debug
  debug:
    var: decrypt_old_postgres.stdout_lines[-1]
```

However, in order for this to work with migrating data, we would need to probably, when backing up, unencrypt the password if it is encrypted, and store it in the config file since there is no reference to the old tower/controller app environment in the postgresql secret.  Though, that kinda seems against the spirit of having this in the first place.  And then when restoring or migrating, they will need to manually encrypt it again.  This seems like the only feasible way this PR can move on functionally.

Also, you can ignore this next part, this is just for me to track where we're setting some things later depending on how we want to go about this:
```
tasks/creation.yml:    - include_tasks: postgres.yml
tasks/creation.yml:    - include_tasks: secrets.yml
tasks/secrets.yml:  include_tasks: dump_generated_secret.yml
tasks/secrets.yml:  include_tasks: dump_secret.yml
tasks/secrets.yml:  include_tasks: dump_secret.yml
tasks/secrets.yml:  include_tasks: dump_secret.yml
tasks/secrets.yml:      bash -c "echo '{{ secrets | to_yaml }}' > {{ backup_dir }}/secrets.yml"
tasks/main.yml:  include_tasks: creation.yml
```

PR #972, opened 2022-06-30
Add to Backup and Restore roles option to set Limits/Requests over {{ ansible_operator_meta.name }}-db-management Pod.

silvinux:
**ISSUE TYPE**

-  New or Enhanced Feature

Backup and Restore roles does not take into account set this options as we can see in the upstream code, the pods are created from a template[1][2] which doesn't have the parameters necessary to be applied in the resource.

Cheers,
Thanks in advance,

[1].-roles/backup/templates/management-pod.yml.j2
[2].-roles/restore/templates/management-pod.yml.j2

--------------------------
Edit so that CI passes:
* New or Enhanced Feature

### DIFF
--- a/roles/backup/defaults/main.yml
+++ b/roles/backup/defaults/main.yml
@@ -20,8 +20,17 @@ clean_backup_on_delete: false
 # Variable to signal that this role is being run as a finalizer
 finalizer_run: false
 
+# Default resource requirements
+backup_resource_requirements:
+  limits:
+    cpu: "500m"
+    memory: "512Mi"
+  requests:
+    cpu: "500m"
+    memory: "512Mi"
 # Allow additional parameters to be added to the pg_dump backup command
 pg_dump_suffix: ''
 
 # Maintain some of the recommended `app.kubernetes.io/*` labels on the resource (self)
 set_self_labels: true
+...
--- a/roles/restore/defaults/main.yml
+++ b/roles/restore/defaults/main.yml
@@ -14,5 +14,15 @@ backup_dir: ''
 # Set no_log settings on certain tasks
 no_log: 'true'
 
+# Default resource requirements
+restore_resource_requirements:
+  limits:
+    cpu: "500m"
+    memory: "512Mi"
+  requests:
+    cpu: "500m"
+    memory: "512Mi"
+
 # Maintain some of the recommended `app.kubernetes.io/*` labels on the resource (self)
 set_self_labels: true
+...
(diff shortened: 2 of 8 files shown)


roles/restore/defaults/main.yml
```diff
@@ -14,5 +14,15 @@ backup_dir: ''
...
+  limits:
+    cpu: "500m"
+    memory: "512Mi"
+  requests:
+    cpu: "500m"
+    memory: "512Mi"
+
 # Maintain some of the recommended `app.kubernetes.io/*` labels on the resource (self)
 set_self_labels: true
+...
```
rooftopcellist, line 28: Is this intentional? This looks like invalid yaml.

rooftopcellist: I take that back, it is valid.  Denotes the end of a yaml file, I just haven't used it yet.  TIL.

roles/backup/defaults/main.yml
```diff
@@ -20,8 +20,17 @@ clean_backup_on_delete: false
 # Variable to signal that this role is being run as a finalizer
 finalizer_run: false
 
+# Default resource requirements
+backup_resource_requirements:
+  limits:
+    cpu: "500m"
+    memory: "512Mi"
+  requests:
+    cpu: "500m"
```
rooftopcellist, line 29: Setting requests this high is likely to cause issues when scheduling the management pod for some folks.  If there aren't enough free resources, it won't be scheduled and the restore will just hang and eventually fail even though the task could be carried out with far fewer resources.  

Similarly, throttling with a limit like that as a default could be problematic for those with much larger databases.  I think it is better to set these defaults high and be on the safe side, and let cluster admins tune it as they see fit.  

I suggest something more like:
```
  backup_resource_requirements:
    limits:
      cpu: "1000m"
      memory: "4096Mi"
    requests:
      cpu: "25m"
      memory: "32Mi"
```

rooftopcellist:
Hi @silvinux, I recently merged a PR that conflicts with yours in a couple spots.  To save you some hassle, I cherry-picked your commits into a branch and fixed the conflicts.  There were a couple other things that were needed:
* Lower requets and higher limits - https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-resource-requests-and-limits - I set them pretty conservatively so that they should be scheduled in just about any environment.  Cluster admins can tune them now as they see fit with your changes.  
* I added entries for the CSV form (used by OKD/Openshift Operator installs)
* expanded your descriptions in the backup & restore manifest files.  

https://github.com/ansible/awx-operator/pull/1030

I hope you don't mind merging there ^^.  Do you mind reviewing my changes and commenting what you think there?

rooftopcellist:
I just merged the other PR.  I think we can safely close this issue.

PR #976, opened 2022-07-01
configure callback receiver workers based on CPU

jainnikhil30:
ISSUE TYPE

- New or Enhanced Feature

### DIFF
--- a/roles/installer/templates/configmaps/config.yaml.j2
+++ b/roles/installer/templates/configmaps/config.yaml.j2
@@ -42,10 +42,19 @@ data:
     SYSTEM_TASK_ABS_CPU = '{{ task_resource_requirements["limits"]["cpu"] }}'
 {% endif %}
 
+{%- set cpu_limit = task_resource_requirements["limits"]["cpu"] if "limits" in task_resource_requirements and "cpu" in task_resource_requirements["limits"] -%}
+{%- if cpu_limit is defined -%}
+{%- set callback_receiver_cpu = cpu_limit | cpu_string_to_decimal -%}
+{%- if callback_receiver_cpu |int > 4 -%}
+# Set callback receiver workers based off cpu limit, default workers are 4, but if we have more than 4 cpu we can set higher value for workers
+    JOB_EVENT_WORKERS = {{ callback_receiver_cpu }}
+{%- endif -%}
+{%- endif %}
+
     SECRET_KEY = get_secret()
-    
+
     ALLOWED_HOSTS = ['*']
-    
+
     INTERNAL_API_URL = 'http://127.0.0.1:8052'
 
     # Sets Ansible Collection path
--- a/roles/installer/filter_plugins/cpu_string_to_decimal.py
+++ b/roles/installer/filter_plugins/cpu_string_to_decimal.py
@@ -1,0 +1,27 @@
+# Copyright (c) 2017 Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from ansible.errors import AnsibleFilterError
+
+__ERROR_MSG = "Not a valid cpu value. Cannot process value"
+
+class FilterModule(object):
+    def filters(self):
+        return {
+            'cpu_string_to_decimal': self.cpu_string_to_decimal
+        }
+    def cpu_string_to_decimal(self, cpu_string):
+
+        # verify if task_output is a dict
+        if not isinstance(cpu_string, str):
+            raise AnsibleFilterError(__ERROR_MSG)
+
+        if cpu_string[-1] == 'm':
+            cpu = int(cpu_string[:-1])//1000
+        else:
+         cpu = int(cpu_string)
+
+        return cpu


roles/installer/templates/config.yaml.j2
```diff
@@ -42,6 +50,12 @@ data:
     if cpu_limit:
         SYSTEM_TASK_ABS_CPU = cpu_limit
 
+    # Set callback receiver workers based off cpu limit, default workers are 4, but if we have more than 4 cpu we can set higher value for workers
+    if cpu_limit:
+        callback_receiver_cpu = {{ cpu_string_to_decimal(cpu_limit) }}
```
kdelee: 1) I think the `{{ }}` need to be in quotes
2) `cpu_limit` is defined in the python file you are templating, but not in the context that jinja will be working. as far as jina is concerned this is just some random text (is not loaded in python). So I think you you have to do something like
```
'{{ cpu_string_to_decimal(task_resource_requirements["limits"]["cpu"] if "limits" in task_resource_requirements and "cpu" in task_resource_requirements["limits"] else "") }}'
```  
I'm not sure if you can do a inline if/else like that. may need to just put that in the macro

roles/installer/templates/config.yaml.j2
```diff
@@ -1,5 +1,13 @@
 # AWX Web ConfigMap.
 ---
+{% macro cpu_string_to_decimal(cpu_limit) -%}
```
shanemcd: This is neat but I'm wondering if it would make more sense as a filter plugin.

roles/installer/templates/config.yaml.j2
```diff
@@ -33,14 +41,23 @@ data:
...
+        SYSTEM_TASK_ABS_MEM = '{{ task_resource_requirements["limits"]["memory"] }}'
+    {% endif %}
 
     # Set cpu available based off of resource request/limit for the task pod
-    cpu_limit = '{{ task_resource_requirements["limits"]["cpu"] if "limits" in task_resource_requirements and "cpu" in task_resource_requirements["limits"] }}'
-    if cpu_limit:
-        SYSTEM_TASK_ABS_CPU = cpu_limit
+    {% if "limits" in task_resource_requirements and "cpu" in task_resource_requirements["limits"] %}
+        SYSTEM_TASK_ABS_CPU = '{{ task_resource_requirements["limits"]["cpu"] }}'
+    {% endif %}
```
kdelee: we may need to backport these two changes to 4.2 @Zokormazo, @jainnikhil30 was having trouble with these templating locally on his machine. I wonder if its a jija version issue. I know I tested this back when I did it for 4.2.0 but I don't think we set limits on nightly so it very well could have broken.

jainnikhil30: @Zokormazo if there is no limit defined, we run into issues, because there is no else on that if condition.
`"msg": "The task includes an option with an undefined variable. The error was: the inline if-expression on line 36 evaluated to false and no else section was defined.` The above should fix that.

roles/installer/templates/config.yaml.j2
jainnikhil30: I could try that. But I felt macro is easier way to do it since we only need to do it for CPU value.

shanemcd: I just don't think this approach is very intuitive or maintainable, and I don't want others to follow this pattern.

jainnikhil30: Sure @shanemcd. Let me put it as a filter plugin.

jainnikhil30: @shanemcd Removed macro and added a filter plugin. Please let me know your feedback.

roles/installer/templates/config.yaml.j2
kdelee: sounds like this is fixed downstream.

kdelee:
@jainnikhil30 should be able to rebase and get fixes for the other uses of `limit` from https://github.com/ansible/awx-operator/pull/929 that now merged

jainnikhil30:
@kdelee Rebased with Squashing. Can we merge now ?

roles/installer/templates/configmaps/config.yaml.j2
```diff
@@ -42,6 +42,15 @@ data:
     SYSTEM_TASK_ABS_CPU = '{{ task_resource_requirements["limits"]["cpu"] }}'
 {% endif %}
 
+    # Set callback receiver workers based off cpu limit, default workers are 4, but if we have more than 4 cpu we can set higher value for workers
+    {% set cpu_limit = task_resource_requirements["limits"]["cpu"] if "limits" in task_resource_requirements and "cpu" in task_resource_requirements["limits"] -%}
+    {%- if cpu_limit is defined -%}
+        {% set callback_receiver_cpu = cpu_limit | cpu_string_to_decimal %}
+            {%- if callback_receiver_cpu |int > 4 -%}
+                JOB_EVENT_WORKERS = {{ callback_receiver_cpu }}
```
shanemcd: Won't this line be incorrectly indented in the rendered file?

jainnikhil30: My bad. Corrected. Thanks @shanemcd

shanemcd (approved):

kdelee:
Merging! Thanks

Issue #716, opened 2022-01-01
Please add configuration capability for ingressClassName


danielmoeller:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
In ingress template the ingessClassName cannot be set. One can only work with ingress annotations, which -according to Nginx Ingress docs- seems to be legacy.

##### ENVIRONMENT
* AWX version: 19.5.0
* Operator version: 0.15.0
* Kubernetes version:  v1.23.1
* AWX install method: openshift, minishift, docker on linux, docker for mac, boot2docker

PR #977, opened 2022-07-01
Adding ingress class name to ingress template (#716)

somebadcode:
Added `ingress_class_name` variable to allow you to set `ingressClassName` in the ingress resource.


somebadcode:
I've been unable to figure out how to run the molecule tests and I don't get the desired results when I run it in our staging cluster.

somebadcode:
The molecule test is able to deploy both the operator and AWX. I can see the pod running fine but the test is unable to actually connect to AWX. I haven't changed any of the testing to include an ingress class so that's not it.

I'm running the tests using Rancher Desktop with the `containerd` runtime and k8s 1.20.x. I've been able to set up the operator and AWX in the cluster myself as well as do port forwarding to test it out but molecule just times out when it tries to do the same.

PR #979, opened 2022-07-07
Introduce `set_self_labels` in CRDs

### LINKED ISSUE #756: AWX operator should not set app.kubernetes.io/managed-by label on AWX resources
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

The AWX operator is incorrectly (IMHO) patching the labels of `AWX` resources to have `app.kubernetes.io/managed-by: awx-operator`. This label is supposed to indicate who is responsible for creating, updating and deleting the resource. The AWX operator does not do this - it reacts to changes in the `AWX` resource made by someone/something else, hence it should not change the `app.kubernetes.io/managed-by` label on the `AWX` resource.

It is correct for the AWX operator to set `app.kubernetes.io/managed-by: awx-operator` on the resources that it creates, e.g. `Deployment`s.

My use case, which is currently being prevented by this bug, is that I have an `AWX` resource as part of a Helm chart. Helm sets the `app.kubernetes.io/managed-by` label to `Helm`, and refuses to manage the resource once that label has been changed.

##### ENVIRONMENT
* AWX version: Default for 0.15.0
* Operator version: 0.15.0
* Kubernetes version: 1.22
* AWX install method: k3s

##### STEPS TO REPRODUCE

Make an AWX resource with the `app.kubernetes.io/managed-by` set to something other than `awx-operator`:

```
kind: AWX
metadata:
  name: my-awx
  labels:
    app.kubernetes.io/managed-by: manual
```

##### EXPECTED RESULTS

That the label `app.kubernetes.io/managed-by` would not change.

##### ACTUAL RESULTS

The label `app.kubernetes.io/managed-by` is set to `awx-operator`.

##### ADDITIONAL INFORMATION

N/A

##### AWX-OPERATOR LOGS

Not relevant.

stanislav-zaprudskiy:
To allow skipping labels maintenance on CRs processed by the operator.
Resolves https://github.com/ansible/awx-operator/issues/756

### DIFF
--- a/roles/backup/tasks/creation.yml
+++ b/roles/backup/tasks/creation.yml
@@ -16,6 +16,7 @@
           app.kubernetes.io/managed-by: "{{ deployment_type }}-operator"
           app.kubernetes.io/component: "{{ deployment_type }}"
           app.kubernetes.io/operator-version: '{{ lookup("env", "OPERATOR_VERSION") }}'
+  when: set_self_labels | bool
 
 - name: Look up details for this backup object
   k8s_info:
--- a/roles/installer/tasks/install.yml
+++ b/roles/installer/tasks/install.yml
@@ -16,6 +16,7 @@
           app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
           app.kubernetes.io/component: '{{ deployment_type }}'
           app.kubernetes.io/operator-version: '{{ lookup("env", "OPERATOR_VERSION") }}'
+  when: set_self_labels | bool
 
 - name: Include secret key configuration tasks
   include_tasks: secret_key_configuration.yml
--- a/roles/restore/tasks/main.yml
+++ b/roles/restore/tasks/main.yml
@@ -16,6 +16,7 @@
           app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
           app.kubernetes.io/component: '{{ deployment_type }}'
           app.kubernetes.io/operator-version: '{{ lookup("env", "OPERATOR_VERSION") }}'
+  when: set_self_labels | bool
 
 - name: Look up details for this restore object
   k8s_info:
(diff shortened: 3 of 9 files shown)


roles/installer/tasks/install.yml
```diff
@@ -16,6 +16,7 @@
           app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
           app.kubernetes.io/component: '{{ deployment_type }}'
           app.kubernetes.io/operator-version: '{{ lookup("env", "OPERATOR_VERSION") }}'
+  when: set_self_labels
```
shanemcd:
```suggestion
  when: set_self_labels | bool
```

roles/backup/tasks/creation.yml
```diff
@@ -16,6 +16,7 @@
           app.kubernetes.io/managed-by: "{{ deployment_type }}-operator"
           app.kubernetes.io/component: "{{ deployment_type }}"
           app.kubernetes.io/operator-version: '{{ lookup("env", "OPERATOR_VERSION") }}'
+  when: set_self_labels
```
shanemcd:
```suggestion
  when: set_self_labels | bool
```

roles/restore/tasks/main.yml
```diff
@@ -16,6 +16,7 @@
           app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
           app.kubernetes.io/component: '{{ deployment_type }}'
           app.kubernetes.io/operator-version: '{{ lookup("env", "OPERATOR_VERSION") }}'
+  when: set_self_labels
```
shanemcd:
```suggestion
  when: set_self_labels | bool
```

shanemcd (approved):
Turns out we need this for another use case. Thanks so much. 🥇

PR #984, opened 2022-07-14
fix helm-chart Makefile issue

bewing:
GNU make resolves the `$(wildcard )` macro when starting a block, and
caches it for the duration of the run.  In order to correctly remove
namespace references from the generated helm charts, we have to split
the generation and editing into two makefile targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,7 +269,10 @@ charts:
 	mkdir -p $@
 
 .PHONY: helm-chart
-helm-chart: kustomize helm kubectl-slice yq charts
+helm-chart: helm-chart-generate helm-chart-slice
+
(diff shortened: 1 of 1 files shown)


bewing:
Tested with
```
rm -rf charts
IMAGE_TAG_BASE=quay.io/ansible/awx-operator VERSION=0.24.0 make helm-chart helm-package
helm upgrade --install awx-operator --repo https://ansible.github.io/awx-operator/ --namespace arbitrary --create-namespace --version 0.24.0 .cr-release-packages/awx-operator-0.24.0.tgz
```

shanemcd (approved):
Thanks 🍰

PR #988, opened 2022-07-18
Deprecate the backup_pvc_namespace field

rooftopcellist:
- Removes backup_pvc_namespace field from OLM forms
  - This field has created confusion for users and there is no strong
    case for needing this functionality
  - Users will still be able to add it to the yaml for the CR if they
    want to use it with the cluster-scoped AWX operator

> Note: a cluster-scoped AWX operator does not exist OOTB at the moment, but it may be added in the future, which is why I mention it above.

### DIFF
--- a/config/crd/bases/awxbackup.ansible.com_awxbackups.yaml
+++ b/config/crd/bases/awxbackup.ansible.com_awxbackups.yaml
@@ -32,13 +32,13 @@ spec:
                   description: Name of the deployment to be backed up
                   type: string
                 backup_pvc:
-                  description: Name of the PVC to be used for storing the backup
+                  description: Name of the backup PVC
                   type: string
                 backup_pvc_namespace:
-                  description: Namespace the PVC is in
+                  description: (Deprecated) Namespace the PVC is in
                   type: string
                 backup_storage_requirements:
-                  description: Storage requirements for the PostgreSQL container
+                  description: Storage requirements for backup PVC (may be similar to existing postgres PVC backing up from)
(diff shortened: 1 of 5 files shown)


config/crd/bases/awxbackup.ansible.com_awxbackups.yaml
```diff
@@ -35,7 +35,7 @@ spec:
                   description: Name of the PVC to be used for storing the backup
```
rlopez133: When I first read this, I read it as "the name of the PVC I'm trying to backup from" (e.g. the source). Can we change this to "Name of the backup PVC"

config/crd/bases/awxbackup.ansible.com_awxbackups.yaml
```diff
@@ -35,7 +35,7 @@ spec:
...
                   type: string
                 backup_pvc_namespace:
-                  description: Namespace the PVC is in
+                  description: (Deprecated) Namespace the PVC is in
                   type: string
                 backup_storage_requirements:
                   description: Storage requirements for the PostgreSQL container
```
rlopez133: Even though we know this is for the PostgreSQL container that will use this PVC, user may not be aware. How about change this to "Storage requirements for backup PVC (may be similar to existing postgres PVC backing up from)"

Or something along those lines

rlopez133 (changes requested):

rooftopcellist:
I got the go ahead from Roger out of band.  _merging_

Issue #986, opened 2022-07-17
Document installation using an Ansible playbook


mhalano:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that AWX Operator is open source software provided for free and that I might not receive a timely response.

### Feature Summary

The README has documentation on how to install using Kubernetes's Kustomization and Helm Chart, but it would be a nice idea if document how to use an Ansible playbook, something like `ansible-playbook -i inventory install-awx.yml`.

PaulVerhoeven1:
There is some documentation about this: https://github.com/ansible/awx-operator/blob/devel/docs/debugging.md

mhalano:
I saw that, but it's a playbook to run the `installer` role used by awx-operator itself. My idea is how to implement the awx-operator using an Ansible playbook and the official installation methods, Kustomize or Helm Chart. Or create a third official way using an official Ansible playbook. My use case is I want to create an Ansible playbook to install awx-operator (and awx itself) in my Kubernetes cluster.

jbradberry:
We'd love to see a PR for this.

d-rupp:
We have a playbook that checks out the awx-operator git repository, creates a kustomization file from a template, build and deploys the kustomization to the your current cluster and then deploy AWX itself. 
It that what you are looking for?

PaulVerhoeven1:
> We have a playbook that checks out the awx-operator git repository, creates a kustomization file from a template, build and deploys the kustomization to the your current cluster and then deploy AWX itself. It that what you are looking for?

I am interested in that!

d-rupp:
I added an example file. Not sure if the location or anything else is correct, but hopefully it helps.

PR #990, opened 2022-07-22
added ansible-deployment-example.yml file

d-rupp:
This shows how an awx-operator deployment can be done via a ansible playbook.

##### SUMMARY
This shows how an awx-operator deployment can be done via a ansible playbook.

For #986

##### ADDITIONAL INFORMATION
This can be improved in a lot of ways with more variables and all that, but it is a simple example that worked for us.


mhalano:
Very cool. Thank you. I made one myself that have, I think, some improvements. Here it is: https://gist.github.com/mhalano/e1b3f05e47e428c5739ad87b46a36ddc

playbooks/ansible-deployment-example.yml
```diff
@@ -0,0 +1,104 @@
...
+    dest: "~/awx-operator/kustomization.yaml"
+    mode: 0640
+
+- name: Deploy controller to the K8s cluster specified in ~/.kube/config.
+  shell: kustomize build . | kubectl apply -f -
+  environment:
+    PATH: "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/root/bin:/snap/bin"
+  args:
+    chdir: "~/awx-operator"
+  changed_when: true
```
mhalano: You could use [Kustomize lookup plugin](https://docs.ansible.com/ansible/latest/collections/kubernetes/core/kustomize_lookup.html) here

playbooks/ansible-deployment-example.yml
```diff
@@ -0,0 +1,104 @@
...
+
+- name: deploy kustomization file
+  copy:
+    content: |
+      ---
+      apiVersion: kustomize.config.k8s.io/v1beta1
+      kind: Kustomization
+      resources:
+        # Find the latest tag here: https://github.com/ansible/awx-operator/releases
+        - github.com/ansible/awx-operator/config/default?ref={{ awx_operator_version }}
```
mhalano: These lines download the AWX Operator directly, so it's redundant to download with Git.

playbooks/ansible-deployment-example.yml
d-rupp: Thanks, implemented. Did not know that one.

playbooks/ansible-deployment-example.yml
d-rupp: Removed the git checkout part, thanks!

rooftopcellist:
Hi @d-rupp, I like your idea of contributing back a playbook to install the operator and deploy an example AWX instance.  

FYI, we already have a playbook [here](https://github.com/ansible/awx-operator/blob/devel/ansible/instantiate-awx-deployment.yml) that just deploys the AWX instance.  

In my mind, there are 2 general use cases:
* Build and Deploy Operator to a namespace & Deploy AWX instance as well
* Only deploy AWX instance (assumes the operator is already available)

Would you be open to having these be separate playbook, but have the one that does both just do an `include_tasks: instantiate-awx-deployment.yml` task?

rooftopcellist:
Could you also move this to `./ansible` instead of `./playbooks`?  The `./playbooks` directory is somewhat managed by the operator-sdk, and may be modified when doing operator-sdk upgrades.  To avoid any issues, we have been putting playbooks like this in the `./ansible` directory.

d-rupp:
@rooftopcellist suggestions implemented, please check if this looks like what you were thinking about.

ansible/ansible-deployment-example.yml
```diff
@@ -0,0 +1,19 @@
...
+    ingress_type: Ingress
+    admin_user: admin
+    admin_email: "admin@localhost"
+
+  tasks:
+    - name: Deploy AWX Operator
+      include_tasks: deploy-awx-operator-kustomization.yml
+
+    - name: Deploy AWX Instance
+      include_tasks: instantiate-awx-deployment.yml
```
d-rupp:
```suggestion
      include_tasks: instantiate-awx-deployment.yml
    
```

ansible/deploy-awx-operator-kustomization.yml
```diff
@@ -0,0 +1,35 @@
...
+              newTag: {{ release.json.tag_name }}
+
+          # Specify a custom namespace in which to install AWX
+          namespace: {{ awx_namespace }}
+        dest: "./kustomization.yaml"
+        mode: 0640
+
+    - name: Create kubernetes resources for lookup output
+      kubernetes.core.k8s:
+        definition: "{{ lookup('kubernetes.core.kustomize', dir='.') }}"
```
rooftopcellist: When testing this out locally, I get this error:

```
TASK [Create kubernetes resources for lookup output] ****************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "lookup plugin (kubernetes.core.kustomize) not found"}
```

Looks like that's because I have kubernetes.core 1.2.1.  (https://docs.ansible.com/ansible/latest/collections/kubernetes/core/kustomize_lookup.html)


I will see if it is possible to bump our requirements.yml file for `kubernetes.core>=2.3.2`. I think it would be weird to expect users to have a newer kubernetes.core version that what we use in the operator roles.

rooftopcellist:
@d-rupp 

```
# Deploy AWX
ansible-playbook ansible/instantiate-awx-deployment.yml -e namespace=$NAMESPACE -e image=quay.io/ansible/awx -e image_version=latest -e service_type=clusterip -e ingress_type=route

# Deploy operator using kustomization
ansible-playbook ansible/deploy-awx-operator-kustomization.yml -e awx_namespace=ca-awx

# Deploy both using wrapper playbook
ansible-playbook ansible/ansible-deployment-example.yml 
```

This last one didn't work.  I think it's because the included files have hosts specified..

```
TASK [Deploy AWX Operator] ******************************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"reason": "conflicting action statements: hosts, tasks\n\nThe error appears to be in '/home/chadams/awx-operator/ansible/deploy-awx-operator-kustomization.yml': line 2, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n---\n- hosts: localhost\n  ^ here\n"}
```

rooftopcellist:
@d-rupp did you get a chance to look at that last comment?

d-rupp:
> @d-rupp did you get a chance to look at that last comment?

Sorry, this got lost in the shuffle. You are correct, i removed that. Downside is you can not call them separately anymore of course.

ansible/ansible-deployment-example.yml
```diff
@@ -0,0 +1,19 @@
...
+  vars:
+    awx_namespace: awx
+    service_type: ClusterIP
+    ingress_type: Ingress
+    admin_user: admin
+    admin_email: "admin@localhost"
+
+  tasks:
+    - name: Deploy AWX Operator
+      include_tasks: deploy-awx-operator-kustomization.yml
```
rooftopcellist, line 16: We could add an extra variable called `operator_only=true`, and have a `when:` here.  

Same for the AWX CR below.

PR #994, opened 2022-07-25
Move custom resource definitions into a CRDs folder

### LINKED ISSUE #993: Unable to install Helm Chart on fresh Kubernetes cluster

### Bug Summary

I attempted to use the provided Helm Chart to install the awx-operator on a fresh Kubernetes cluster but ran into a couple issues.

Firstly, the AWX `customresourcedefinitions` are included in the template directory. This results in Helm trying to create the AWX custom resources _before_ importing the custom resource definitions:

`Error: unable to build kubernetes objects from release manifest: resource mapping not found for name: "awx" namespace: "awx" from "": no matches for kind "AWX" in version "awx.ansible.com/v1beta1"
ensure CRDs are installed first`

According to the Helm [documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/), custom resource definitions should be separated out into a separate "CRDs" folder so that they are installed first.

Secondly, a templating issue exists in the managed Postgres deployment here: https://github.com/ansible/awx-operator/blob/0be17476cd691c7727fd3d1c43c6d449fd89a862/.helm/starter/templates/postgres-config.yaml#L10 - the port number is not correctly stringified, causing installation to fail:

`Error: Secret in version "v1" cannot be handled as a Secret: v1.Secret.StringData: ReadString: expects " or n, but found 5, error found in #10 byte of ...|","port":5432 `.

I will put up a PR shortly which should address these issues. As the second problem is quite minor I've included it in this same issue, but please let me know if I ought to be splitting it out.

### AWX Operator version

0.25.0

### AWX version

0.25.0

### Kubernetes platform

kubernetes

1.21

### Modifications

no

### Steps to reproduce

1) Create a fresh cluster or purge all AWX CRDs from an existing cluster:
`k delete crd awxs.awx.ansible.com
k delete crd awxbackups.awx.ansible.com
k delete crd awxrestores.awx.ansible.com `

2) Attempt a fresh installation of the awx-operator Helm Chart:

`helm upgrade --install awx-operator --create-namespace --namespace awx -f values.yaml --version 0.25.0 awx-operator/awx-operator `

### Expected results

Custom resource definitions are imported before trying to create custom resources.

### Actual results

The chart attempts to create custom resources before the CRDs are imported, causing installation to fail.

### Additional information

_No response_

### Operator Logs

_No response_

FleetAdmiralButter:
##### SUMMARY

As described in the Helm documentation (https://helm.sh/docs/chart_best_practices/custom_resource_definitions/), custom resource definitions should be included in a separate "crds" directory to ensure that Helm imports them before trying to create any custom resources. I've added a couple steps to the Makefile to move all `customresourcedefinitions` out of the template directory and into the "crds" directory.

I also found an issue with the Postgres deployment where the Port value was not being stringified properly. I've added `| quote` to ensure the templated `Secret` is valid.

##### ISSUE TYPE

 - Bug or Docs Fix

fixes #993

### DIFF
--- a/.helm/starter/templates/postgres-config.yaml
+++ b/.helm/starter/templates/postgres-config.yaml
@@ -7,7 +7,7 @@ metadata:
 {{- with $.Values.AWX.postgres }}
 stringData:
   host: {{ .host }}
-  port: {{ .port }}
+  port: {{ .port | quote }}
   database: {{ .dbName }}
   username: {{ .username }}
   password: {{ .password }}
--- a/Makefile
+++ b/Makefile
@@ -295,6 +295,8 @@ helm-chart-generate: kustomize helm kubectl-slice yq charts
 			--output-dir=charts/$(CHART_NAME)/templates \
 			--sort-by-kind
 	@echo "AWX Operator installed with Helm Chart version $(VERSION)" > charts/$(CHART_NAME)/templates/NOTES.txt
+	mkdir charts/$(CHART_NAME)/crds
+	mv charts/$(CHART_NAME)/templates/customresourcedefinition* charts/$(CHART_NAME)/crds
 
 .PHONY: helm-chart-edit
 helm-chart-slice:


miles-w-3:
This looks great, apologies for the Postgres error and thanks for the fix

rooftopcellist (approved):

PR #1005, opened 2022-07-29
Use new postgres pod label when migrating from old instance

rooftopcellist:
##### SUMMARY

Follow-up for https://github.com/ansible/awx-operator/pull/841

##### ISSUE TYPE

 - Bug

##### ADDITIONAL INFORMATION

We can no longer hard-code the name of the postgres pod in the logic used to migrate data from an old AWX deployment (migrate_data.yml).

### DIFF
--- a/roles/installer/tasks/migrate_data.yml
+++ b/roles/installer/tasks/migrate_data.yml
@@ -21,17 +21,12 @@
 - name: Get the postgres pod information
   k8s_info:
     kind: Pod
-    namespace: '{{ ansible_operator_meta.namespace }}'
-    name: '{{ ansible_operator_meta.name }}-postgres-0'  # using name to keep compatibility
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    label_selectors:
+      - "{{ postgres_label_selector }}"
     field_selectors:
       - status.phase=Running
   register: postgres_pod
-  until:
-    - "postgres_pod['resources'] | length"
-    - "postgres_pod['resources'][0]['status']['phase'] == 'Running'"
-    - "postgres_pod['resources'][0]['status']['containerStatuses'][0]['ready'] == true"
-  delay: 5
-  retries: 60
 
 - name: Set the resource pod name as a variable.
   set_fact:


rooftopcellist:
Steps taken to test this:

1. Check out awx-operator 0.25.0 and add a kustomize patch to set the imagePullPolicy to Always for the manager 

```
#file: ./config/manager/kustomization.yaml

patches:
- path: pull_policy/Always.yaml
```

2. Deploy the operator with `make docker-build docker-push deploy`
3. Create AWX CR called `awx`, once the UI is available, make some dummy resources to test data migration with 
4. Check out the branch for this PR () and build and deploy the operator in a similar fashion.
5. Wait for the existing AWX CR to be reconciled by the new operator container.
6. Create a new AWX CR called `new-awx` with the secret key and the postgres configuration secrets of the first deployment set in the spec.

```
$ cat awx-migrate-cr.yml 
---
kind: AWX
metadata:
  name: new-awx
spec:
  no_log: 'false'
  service_type: nodeport
  nodeport_port: 30081    # so that it doesn't conflict with existing deployment
  image_pull_policy: Always
  ingress_type: route
  secret_key_secret: awx-secret-key
  old_postgres_configuration_secret: awx-postgres-configuration

```

7. Observe that the step to "Get the postgres_pod information" now passes successfully
8. Observe that the data is successfully migrated and available on the new instance.  In my case, this user named Rick.  

image



#### Reproducer

To reproduce the original issue, I did the same steps as above, but used the `devel` branch.  

That resulted in this error:

```
TASK [installer : Get the postgres pod information] ****************************
task path: /opt/ansible/roles/installer/tasks/migrate_data.yml:21

-------------------------------------------------------------------------------
{"level":"info","ts":1659122188.3679442,"logger":"logging_event_handler","msg":"[playbook task start]","name":"new-awx","namespace":"ca-test","gvk":"awx.ansible.com/v1beta1, Kind=AWX","event_type":"playbook_on_task_start","job":"7511219993076592699","EventData.Name":"installer : Get the postgres pod information"}
{"level":"info","ts":1659122189.2721891,"logger":"proxy","msg":"Cache miss: /v1, Kind=Pod, ca-test/new-awx-postgres-0"}
```

fosterseth (approved):

PR #1012, opened 2022-08-02
Allow runtime modification of receptor config without updating configmap

TheRealHaoLiu:
##### SUMMARY
As part of the feature work for enabling AWX controlplane deployed on Kubernetes to peer out to remote execution node we needed to add the ability for AWX to dynamic modify the receptor.conf.

This PR add the following 
- on AWX operator startup
  - if no receptor mesh CA cert/key exist generate a pair of CA cert/key and store in a secret
  - if no receptor work signing key pair exists, generate a keypair and store in a secret
- on receptor container creation receptor client cert using  receptor mesh CA cert
- change logic for how receptor.conf is written to allow receptor.conf to be editable at runtime

Not all containers need access to all receptor files.
The volume mount changes in this PR fulfill the following:
- `receptor.conf` need to be shared between task container and ee container
  - **task** container writes the `receptor.conf`
  - **ee** consume the `receptor.conf`
- receptor ca cert/key need to be mounted by both ee container and web container
  - **ee** container need the ca cert
  - **web** container will need the ca key to sign client cert for remote execution node
  - **web** container will need the ca cert to generate install bundle for remote execution node
- receptor work private/public key need to be mounted by both ee container and web container
  - **ee** container need to private key to sign the work
  - **web** container need the public key to generate install bundle  for remote execution node

##### ISSUE TYPE
 - New or Enhanced Feature

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,11 @@
 FROM quay.io/operator-framework/ansible-operator:v1.22.2
 
+USER 0
+
+RUN dnf install -y openssl
+
+USER 1001
+
 ARG DEFAULT_AWX_VERSION
 ARG OPERATOR_VERSION
 ENV DEFAULT_AWX_VERSION=${DEFAULT_AWX_VERSION}
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -131,8 +131,6 @@ _redis_image: docker.io/redis
 _redis_image_version: 7
 _postgres_image: postgres
 _postgres_image_version: 13
-_init_container_image: quay.io/centos/centos
-_init_container_image_version: stream8
 image_pull_policy: IfNotPresent
 image_pull_secrets: []
 
@@ -157,6 +155,9 @@ ee_images:
     image: quay.io/ansible/awx-ee:latest
 
 _control_plane_ee_image: quay.io/ansible/awx-ee:latest
+
+_init_container_image: "{{ _control_plane_ee_image.split(':')[0] }}"
+_init_container_image_version: "{{ _control_plane_ee_image.split(':')[1] }}"
 
 create_preload_data: true
 
--- a/roles/installer/tasks/resources_configuration.yml
+++ b/roles/installer/tasks/resources_configuration.yml
@@ -27,6 +27,95 @@
   set_fact:
     _control_plane_ee_image: "{{ _custom_control_plane_ee_image | default(lookup('env', 'RELATED_IMAGE_CONTROL_PLANE_EE')) | default(_control_plane_ee_image, true) }}"
 
+- name: Check for Receptor CA Secret
+  k8s_info:
+    kind: Secret
+    namespace: '{{ ansible_operator_meta.namespace }}'
+    name: '{{ ansible_operator_meta.name }}-receptor-ca'
+  register: _receptor_ca
+  no_log: "{{ no_log }}"
+
+- name: Create Receptor Mesh CA
+  block:
+    - name: Create tempfile for receptor-ca.key
+      tempfile:
+        state: file
+        suffix: .key
+      register: _receptor_ca_key_file
+    - name: Generate Receptor CA key
+      command: |
+        openssl genrsa -out {{ _receptor_ca_key_file.path }} 4096
+      no_log: "{{ no_log }}"
+    - name: Create tempfile for receptor-ca.crt
+      tempfile:
+        state: file
+        suffix: .crt
+      register: _receptor_ca_crt_file
+    - name: Generate Receptor CA cert
+      command: |
+        openssl req -x509 -new -nodes -key {{ _receptor_ca_key_file.path }} \
+          -subj "/CN={{ ansible_operator_meta.name }} Receptor Root CA" \
+          -sha256 -days 3650 -out {{ _receptor_ca_crt_file.path }}
+      no_log: "{{ no_log }}"
+    - name: Create Receptor CA secret
+      k8s:
+        apply: true
+        definition: "{{ lookup('template', 'secrets/receptor_ca_secret.yaml.j2') }}"
+      no_log: "{{ no_log }}"
+    - name: Remove tempfiles
+      file:
+        path: "{{ item }}"
+        state: absent
+      loop:
+        - "{{ _receptor_ca_key_file.path }}"
+        - "{{ _receptor_ca_crt_file.path }}"
+  when: not _receptor_ca['resources'] | default([]) | length
+
+- name: Check for Receptor work signing Secret
+  k8s_info:
+    kind: Secret
+    namespace: '{{ ansible_operator_meta.namespace }}'
+    name: '{{ ansible_operator_meta.name }}-receptor-work-signing'
+  register: _receptor_work_signing
+  no_log: "{{ no_log }}"
+
+- name: Generate Receptor work signing RSA key pair
+  block:
+    - name: Create tempfile for receptor work signing private key
+      tempfile:
+        state: file
+        suffix: .pem
+      register: _receptor_work_signing_private_key_file
+    - name: Generate Receptor work signing private key
+      command: |
+        openssl genrsa -out {{ _receptor_work_signing_private_key_file.path }} 4096
+      no_log: "{{ no_log }}"
+    - name: Create tempfile for receptor work signing public key
+      tempfile:
+        state: file
+        suffix: .pem
+      register: _receptor_work_signing_public_key_file
+    - name: Generate Receptor work signing public key
+      command: |
+        openssl rsa \
+          -in {{ _receptor_work_signing_private_key_file.path }} \
+          -out {{ _receptor_work_signing_public_key_file.path }} \
+          -outform PEM -pubout
+      no_log: "{{ no_log }}"
+    - name: Create Receptor work signing Secret
+      k8s:
+        apply: true
+        definition: "{{ lookup('template', 'secrets/receptor_work_signing_secret.yaml.j2') }}"
+      no_log: "{{ no_log }}"
+    - name: Remove tempfiles
+      file:
+        path: "{{ item }}"
+        state: absent
+      loop:
+        - "{{ _receptor_work_signing_private_key_file.path }}"
+        - "{{ _receptor_work_signing_public_key_file.path }}"
+  when: not _receptor_work_signing['resources'] | default([]) | length
+
 - name: Apply Resources
   k8s:
     apply: yes
--- a/roles/installer/templates/deployments/deployment.yaml.j2
+++ b/roles/installer/templates/deployments/deployment.yaml.j2
@@ -39,7 +39,6 @@ spec:
       priorityClassName: '{{ control_plane_priority_class }}'
 {% endif %}
       initContainers:
-{% if bundle_ca_crt or projects_persistence|bool or init_container_extra_commands %}
         - name: init
           image: '{{ _init_container_image }}'
           imagePullPolicy: '{{ image_pull_policy }}'
@@ -47,6 +46,9 @@ spec:
             - /bin/sh
             - -c
             - |
+              hostname=$MY_POD_NAME
+              receptor --cert-makereq bits=2048 commonname=$hostname dnsname=$hostname nodeid=$hostname outreq=/etc/receptor/tls/receptor.req outkey=/etc/receptor/tls/receptor.key
+              receptor --cert-signreq req=/etc/receptor/tls/receptor.req cacert=/etc/receptor/tls/ca/receptor-ca.crt cakey=/etc/receptor/tls/ca/receptor-ca.key outcert=/etc/receptor/tls/receptor.crt verify=yes
 {% if bundle_ca_crt  %}
               mkdir -p /etc/pki/ca-trust/extracted/{java,pem,openssl,edk2}
               update-ca-trust
@@ -58,7 +60,17 @@ spec:
 {% if init_container_extra_commands %}
               {{ init_container_extra_commands | indent(width=14) }}
 {% endif %}
+          env:
+            - name: MY_POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
           volumeMounts:
+            - name: "{{ ansible_operator_meta.name }}-receptor-ca"
+              mountPath: "/etc/receptor/tls/ca"
+              readOnly: true
+            - name: "{{ ansible_operator_meta.name }}-receptor-tls"
+              mountPath: "/etc/receptor/tls/"
 {% if bundle_ca_crt  %}
             - name: "ca-trust-extracted"
               mountPath: "/etc/pki/ca-trust/extracted"
@@ -73,7 +85,6 @@ spec:
 {% endif %}
 {% if init_container_extra_volume_mounts -%}
             {{ init_container_extra_volume_mounts | indent(width=12, first=True) }}
-{% endif %}
 {% endif %}
       containers:
         - image: '{{ _redis_image }}'
@@ -163,6 +174,13 @@ spec:
               mountPath: "/var/lib/awx/rsyslog"
             - name: "{{ ansible_operator_meta.name }}-projects"
               mountPath: "/var/lib/awx/projects"
+            - name: "{{ ansible_operator_meta.name }}-receptor-work-signing"
+              mountPath: "/etc/receptor/signing/work-public-key.pem"
+              subPath: "work-public-key.pem"
+              readOnly: true
+            - name: "{{ ansible_operator_meta.name }}-receptor-ca"
+              mountPath: "/etc/receptor/tls/ca"
+              readOnly: true
 {% if development_mode | bool %}
             - name: awx-devel
               mountPath: "/awx_devel"
@@ -236,8 +254,10 @@ spec:
             - name: rsyslog-dir
               mountPath: "/var/lib/awx/rsyslog"
             - name: "{{ ansible_operator_meta.name }}-receptor-config"
-              mountPath: "/etc/receptor/receptor.conf"
-              subPath: receptor.conf
+              mountPath: "/etc/receptor/"
+            - name: "{{ ansible_operator_meta.name }}-receptor-work-signing"
+              mountPath: "/etc/receptor/signing/work-private-key.pem"
+              subPath: "work-private-key.pem"
               readOnly: true
             - name: receptor-socket
               mountPath: "/var/run/receptor"
@@ -279,7 +299,15 @@ spec:
           name: '{{ ansible_operator_meta.name }}-ee'
           imagePullPolicy: '{{ image_pull_policy }}'
           resources: {{ ee_resource_requirements }}
-          args: ['receptor', '--config', '/etc/receptor/receptor.conf']
+          args:
+            - /bin/sh
+            - -c
+            - |
+              if [ ! -f /etc/receptor/receptor.conf ]; then
+                cp /etc/receptor/receptor-default.conf /etc/receptor/receptor.conf
+                sed -i "s/HOSTNAME/$HOSTNAME/g" /etc/receptor/receptor.conf
+              fi
+              exec receptor --config /etc/receptor/receptor.conf
           volumeMounts:
 {% if bundle_ca_crt %}
             - name: "ca-trust-extracted"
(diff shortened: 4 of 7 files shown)


TheRealHaoLiu:
@shanemcd @rooftopcellist can u give this PR a quick WIP review

TheRealHaoLiu:
tested env setup using the following instruction
- env: ocp v4.10.18
- custom awx-operator build/deploy using `make docker-build docker-push deploy IMAGE_TAG_BASE=quay.io/haoliu/awx-operator`
- deploy awx using operator with ```ansible-playbook ansible/instantiate-awx-deployment.yml -e namespace=`oc project -q` -e image=quay.io/ansible/awx -e image_version=latest -e service_type=nodeport -e ingress_type=route```

validated the basic functionality of receptor by doing the following
- project sync of the 'Demo Project'
- running job using 'Demo Job Template'
- validated the functionality of the PR by

awx-receptor-ca created as designed
- /etc/receptor/receptor-default.conf and /etc/receptor/receptor.conf file is correctly created
- /etc/receptor/tls directory is correctly created and contain cert sign by the CA as well as the ca cert
- /etc/receptor/receptor.conf is modifiable and persist through restart (tested by exec-ing into the container and `kill 1`)

TheRealHaoLiu:
feedback from @shanemcd 
need to address 
```
# TODO: this is only needed if we use receptor to generate the mesh CA.
# We can probably use openssl for that.
COPY --from=quay.io/ansible/receptor:devel /usr/bin/receptor /usr/bin/receptor
```

TheRealHaoLiu:
made the change for what @shanemcd was asking for

TheRealHaoLiu:
note that both the CA cert and the cert signed by the CA are only valid for 10 years 

in the future we might need to handle cert rotation

shanemcd (approved):

jbradberry (approved):
Nothing jumps out at me as objectionable.

TheRealHaoLiu:
i think we will continue to add onto this PR to add the receptor work signing certificate as well 
so i'm going to put this PR back in draft

roles/installer/tasks/resources_configuration.yml
```diff
@@ -27,6 +27,55 @@
...
+  no_log: "{{ no_log }}"
+
+- name: Create Receptor Mesh CA
+  block:
+    - name: Generate Receptor CA key
+      command: |
+        openssl genrsa -out /tmp/{{ ansible_operator_meta.name }}-receptor-ca.key 4096
+    - name: Generate Receptor CA cert
+      command: |
+        openssl req -x509 -new -nodes -key /tmp/{{ ansible_operator_meta.name }}-receptor-ca.key \
```
Zokormazo: where does this /tmp live? I wonder if we are adding a security issue here storing the keys in the operator pod. Should we use `tempfile` module to create a tmp dir and remove it after creating the secret? I don't like the fact we have stale ca keys

roles/installer/tasks/resources_configuration.yml
```diff
@@ -27,6 +27,55 @@
...
+  register: _receptor_work_signing
+  no_log: "{{ no_log }}"
+
+- name: Generate Receptor work signing RSA keypair
+  block:
+    - name: Generate Receptor work signing private key
+      command: openssl genrsa -out /tmp/{{ ansible_operator_meta.name }}-work-private-key.pem 4096
+      no_log: "{{ no_log }}"
+    - name: Generate Receptor work signing public key
+      command: openssl rsa -in /tmp/{{ ansible_operator_meta.name }}-work-private-key.pem \
```
Zokormazo: same concerns here

roles/installer/templates/deployments/deployment.yaml.j2
```diff
@@ -46,14 +46,16 @@ spec:
...
-{% if bundle_ca_crt or projects_persistence|bool or init_container_extra_commands %}
         - name: init
           image: '{{ _init_container_image }}'
           imagePullPolicy: '{{ image_pull_policy }}'
           command:
             - /bin/sh
             - -c
             - |
+              hostname=$MY_POD_NAME
+              receptor --cert-makereq bits=2048 commonname=$hostname dnsname=$hostname nodeid=$hostname outreq=/etc/receptor/tls/receptor.req outkey=/etc/receptor/tls/receptor.key
```
Zokormazo, line 50: why not use `${MY_POD_NAME}` directly?

roles/installer/templates/deployments/deployment.yaml.j2
```diff
@@ -286,7 +306,14 @@ spec:
           name: '{{ ansible_operator_meta.name }}-ee'
           imagePullPolicy: '{{ image_pull_policy }}'
           resources: {{ ee_resource_requirements }}
-          args: ['receptor', '--config', '/etc/receptor/receptor.conf']
+          args:
+            - /bin/sh
+            - -c
+            - |
+              if [ ! -f /etc/receptor/receptor.conf ]; then
+                cp /etc/receptor/receptor-default.conf /etc/receptor/receptor.conf
```
Zokormazo, line 307: where is this receptor-default.conf coming from? Is this something @rooftopcellist should know about for downstream?

roles/installer/defaults/main.yml
```diff
@@ -158,6 +156,9 @@ ee_images:
 
 _control_plane_ee_image: quay.io/ansible/awx-ee:latest
 
+_init_container_image: "{{ _control_plane_ee_image.split(':')[0] }}"
+_init_container_image_version: "{{ _control_plane_ee_image.split(':')[1] }}"
```
Zokormazo, line 160: I wonder if downstream the control plane ee has everything to run all the different init scripts we could run in this init container

roles/installer/templates/deployments/deployment.yaml.j2
Zokormazo: m, does always /etc/receptor/tls exists? another thing for @rooftopcellist to check

Zokormazo: should we do `bits` something customizable? or go with 4096?

roles/installer/templates/deployments/deployment.yaml.j2
```diff
@@ -414,6 +460,8 @@ spec:
         - name: rsyslog-dir
           emptyDir: {}
         - name: {{ ansible_operator_meta.name }}-receptor-config
+          emptyDir: {}
+        - name: {{ ansible_operator_meta.name }}-default-receptor-config
```
Zokormazo, line 458: oh, it comes from here. Found the answer :)

roles/installer/templates/deployments/deployment.yaml.j2
rooftopcellist: I think passing MY_POD_NAME as an env var based on the value of metadata.name should work just fine.  I'm not entirely sure if you can use metadata.name directly there.  I'm Ok with either approach.  

------------------
We should probably check if /etc/receptor/tls exists if there is any expectation that the operator will be backwards compatible with older awx images.  Ideally it should, but if it can't be, let's be clear about that in the README.md and changelog.

roles/installer/templates/deployments/deployment.yaml.j2
TheRealHaoLiu: good point!

roles/installer/tasks/resources_configuration.yml
TheRealHaoLiu: good point!

roles/installer/tasks/resources_configuration.yml
TheRealHaoLiu: another good point

roles/installer/templates/deployments/deployment.yaml.j2
rooftopcellist: @Zokormazo Thanks for thinking to ask!

I talked to @TheRealHaoLiu, he just renamed the receptor.conf that gets config-mapped.  So this file does not need to be updated in the container build repo, it's source is `roles/installer/templates/configmaps/config.yaml.j2` which will be automatically synced downstream.

roles/installer/templates/deployments/deployment.yaml.j2
TheRealHaoLiu: /etc/receptor/tls should always exist since https://github.com/ansible/awx-operator/pull/1012/files#diff-29de64e6284e07a8a16b351b132d812618c7f4e3485cec09987c2374257b94fcR76-R80

puts it there 

older awx image should be fine since the operator is generating the CA cert/key and the init container is what's generating the client cert/key

roles/installer/templates/deployments/deployment.yaml.j2
TheRealHaoLiu: receptor-default.conf comes from configmap that's templated from config.yaml.j2 

it's the original receptor config file so it's not a new file

rooftopcellist:
I built and deployed an operator image from this PR and created a 21.5.0 AWX deployment with it to check if there was any backwards compatibility issues.  I was able to run jobs from the UI without issue and didn't see any other breakage.

rooftopcellist (approved):

roles/installer/defaults/main.yml
dsavineau: looks like this change is only needed to run receptor command from the ee image and the default ee image for the control place downstream (ee-supported-rhel8) already have receptor included.

Dockerfile
```diff
@@ -1,5 +1,11 @@
 FROM quay.io/operator-framework/ansible-operator:v1.22.2
 
+USER 0
+
+RUN dnf install -y openssl
```
dsavineau, line 5: I'm not a big fan of using openssl command while we could rely on the ansible community.crypto collection for that (which requires the python3-cryptography package).

https://github.com/ansible-collections/community.crypto

roles/installer/tasks/resources_configuration.yml
```diff
@@ -27,6 +27,55 @@
...
+- name: Create Receptor Mesh CA
+  block:
+    - name: Generate Receptor CA key
+      command: |
+        openssl genrsa -out /tmp/{{ ansible_operator_meta.name }}-receptor-ca.key 4096
+    - name: Generate Receptor CA cert
+      command: |
+        openssl req -x509 -new -nodes -key /tmp/{{ ansible_operator_meta.name }}-receptor-ca.key \
+          -subj "/CN={{ ansible_operator_meta.name }} Receptor Root CA" \
+          -sha256 -days 3650 -out /tmp/{{ ansible_operator_meta.name }}-receptor-ca.crt
```
dsavineau: As mentioned earlier, if we use the community.crypto collection then we could have more flexibility on the parameters (validity, key size, etc..)  for the end user

roles/installer/templates/deployments/deployment.yaml.j2
```diff
@@ -46,14 +46,16 @@ spec:
...
         - name: init
           image: '{{ _init_container_image }}'
           imagePullPolicy: '{{ image_pull_policy }}'
           command:
             - /bin/sh
             - -c
             - |
+              hostname=$MY_POD_NAME
+              receptor --cert-makereq bits=2048 commonname=$hostname dnsname=$hostname nodeid=$hostname outreq=/etc/receptor/tls/receptor.req outkey=/etc/receptor/tls/receptor.key
+              receptor --cert-signreq req=/etc/receptor/tls/receptor.req cacert=/etc/receptor/tls/ca/receptor-ca.crt cakey=/etc/receptor/tls/ca/receptor-ca.key outcert=/etc/receptor/tls/receptor.crt verify=yes
```
dsavineau, line 51: What does receptor on the TLS client cert/key that openssl (or the community.crypto collection) coudn't do ?
I mean if it's only for creating and signing TLS certificates then we don't really need to have two different workflows for the CA and client certificates.
Also, any reason why the TLS client certificate and key aren't stored as a k8s Secret like for the CA ? This would mean that the TLS client certificate/key will be recreate each time the deployment recreate (like during upgrades) but I don't know if that's an issue or not.

roles/installer/tasks/resources_configuration.yml
TheRealHaoLiu: addressed the comment by using a tempfile

roles/installer/tasks/resources_configuration.yml
TheRealHaoLiu: addressed the comment and now use a tempfile

roles/installer/templates/deployments/deployment.yaml.j2
TheRealHaoLiu: it is intended for the client cert/key to be regenerated for each new instance so that the instance will have a unique identity

TheRealHaoLiu: this runs in the init container that's why we did not use ansible playbook 

using receptor to generate it is just for connivance and yes receptor does put 
```
# Red Hat has an OID namespace (RHANANA). Receptor has its own designation under that.
RECEPTOR_OID = "1.3.6.1.4.1.2312.19.1"
```
in the SAN

roles/installer/tasks/resources_configuration.yml
```diff
@@ -27,6 +27,95 @@
...
+- name: Create Receptor Mesh CA
+  block:
+    - name: Create tempfile for receptor-ca.key
+      tempfile:
+        state: file
+        suffix: .key
+      register: _receptor_ca_key_file
+    - name: Generate Receptor CA key
+      command: |
+        openssl genrsa -out {{ _receptor_ca_key_file }} 4096
```
dsavineau:
```suggestion
        openssl genrsa -out {{ _receptor_ca_key_file.path }} 4096
```

roles/installer/tasks/resources_configuration.yml
```diff
@@ -27,6 +27,95 @@
...
+    - name: Create tempfile for receptor-ca.crt
+      tempfile:
+        state: file
+        suffix: .crt
+      register: _receptor_ca_crt_file
+    - name: Generate Receptor CA cert
+      command: |
+        openssl req -x509 -new -nodes -key {{ _receptor_ca_key_file }} \
+          -subj "/CN={{ ansible_operator_meta.name }} Receptor Root CA" \
+          -sha256 -days 3650 -out {{ _receptor_ca_crt_file }}
```
dsavineau:
```suggestion
        openssl req -x509 -new -nodes -key {{ _receptor_ca_key_file.path }} \
          -subj "/CN={{ ansible_operator_meta.name }} Receptor Root CA" \
          -sha256 -days 3650 -out {{ _receptor_ca_crt_file.path }}
```

roles/installer/tasks/resources_configuration.yml
```diff
@@ -27,6 +27,95 @@
...
+        apply: true
+        definition: "{{ lookup('template', 'secrets/receptor_ca_secret.yaml.j2') }}"
+      no_log: "{{ no_log }}"
+    - name: Remove tempfiles
+      file:
+        path: "{{ item }}"
+        state: absent
+      loop:
+        - "{{ _receptor_ca_key_file }}"
+        - "{{ _receptor_ca_crt_file }}"
```
dsavineau:
```suggestion
        - "{{ _receptor_ca_key_file.path }}"
        - "{{ _receptor_ca_crt_file.path }}"
```

roles/installer/tasks/resources_configuration.yml
```diff
@@ -27,6 +27,95 @@
...
+- name: Generate Receptor work signing RSA key pair
+  block:
+    - name: Create tempfile for receptor work signing private key
+      tempfile:
+        state: file
+        suffix: .pem
+      register: _receptor_work_signing_private_key_file
+    - name: Generate Receptor work signing private key
+      command: |
+        openssl genrsa -out {{ _receptor_work_signing_private_key_file }} 4096
```
dsavineau:
```suggestion
        openssl genrsa -out {{ _receptor_work_signing_private_key_file.path }} 4096
```

roles/installer/tasks/resources_configuration.yml
```diff
@@ -27,6 +27,95 @@
...
+    - name: Create tempfile for receptor work signing public key
+      tempfile:
+        state: file
+        suffix: .pem
+      register: _receptor_work_signing_public_key_file
+    - name: Generate Receptor work signing public key
+      command: |
+        openssl rsa \
+          -in {{ _receptor_work_signing_private_key_file }} \
+          -out {{ _receptor_work_signing_public_key_file }} \
```
dsavineau:
```suggestion
          -in {{ _receptor_work_signing_private_key_file.path }} \
          -out {{ _receptor_work_signing_public_key_file.path }} \
```

roles/installer/tasks/resources_configuration.yml
```diff
@@ -27,6 +27,95 @@
...
+        apply: true
+        definition: "{{ lookup('template', 'secrets/receptor_work_signing_secret.yaml.j2') }}"
+      no_log: "{{ no_log }}"
+    - name: Remove tempfiles
+      file:
+        path: "{{ item }}"
+        state: absent
+      loop:
+        - "{{ _receptor_work_signing_private_key_file }}"
+        - "{{ _receptor_work_signing_public_key_file }}"
```
dsavineau:
```suggestion
        - "{{ _receptor_work_signing_private_key_file.path }}"
        - "{{ _receptor_work_signing_public_key_file.path }}"
```

roles/installer/templates/secrets/receptor_ca_secret.yaml.j2
```diff
@@ -0,0 +1,15 @@
...
+  namespace: '{{ ansible_operator_meta.namespace }}'
+  labels:
+    app.kubernetes.io/name: '{{ ansible_operator_meta.name }}'
+    app.kubernetes.io/part-of: '{{ ansible_operator_meta.name }}'
+    app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
+    app.kubernetes.io/component: '{{ deployment_type }}'
+    app.kubernetes.io/operator-version: '{{ lookup("env", "OPERATOR_VERSION") }}'
+data:
+  receptor-ca.crt: '{{ lookup('file', '{{ _receptor_ca_crt_file }}') | b64encode }}'
+  receptor-ca.key: '{{ lookup('file', '{{ _receptor_ca_key_file }}') | b64encode }}'
```
dsavineau:
```suggestion
  receptor-ca.crt: '{{ lookup('file', '{{ _receptor_ca_crt_file.path }}') | b64encode }}'
  receptor-ca.key: '{{ lookup('file', '{{ _receptor_ca_key_file.path }}') | b64encode }}'
```

roles/installer/templates/secrets/receptor_work_signing_secret.yaml.j2
```diff
@@ -0,0 +1,15 @@
...
+  namespace: '{{ ansible_operator_meta.namespace }}'
+  labels:
+    app.kubernetes.io/name: '{{ ansible_operator_meta.name }}'
+    app.kubernetes.io/part-of: '{{ ansible_operator_meta.name }}'
+    app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
+    app.kubernetes.io/component: '{{ deployment_type }}'
+    app.kubernetes.io/operator-version: '{{ lookup("env", "OPERATOR_VERSION") }}'
+data:
+  work-private-key.pem: '{{ lookup('file', '{{ _receptor_work_signing_private_key_file }}') | b64encode }}'
+  work-public-key.pem: '{{ lookup('file', '{{ _receptor_work_signing_public_key_file }}') | b64encode }}'
```
dsavineau:
```suggestion
  work-private-key.pem: '{{ lookup('file', '{{ _receptor_work_signing_private_key_file.path }}') | b64encode }}'
  work-public-key.pem: '{{ lookup('file', '{{ _receptor_work_signing_public_key_file.path }}') | b64encode }}'
```

dsavineau (changes requested):
when using the `tempfile` module then you need to use the `path` field from the registered variables.

https://docs.ansible.com/ansible/latest/collections/ansible/builtin/tempfile_module.html#return-values

Dockerfile
shanemcd: We talked about this in a call earlier - everyone agrees with you, but there's some urgency to get things merged and there's other outstanding items that need to be taken care of. If you want to submit a PR after this merges (or submit a PR to this branch) feel free.

dsavineau (approved):

PR #1015, opened 2022-08-05
Organize installer templates into subdirectories

shanemcd:
##### ISSUE TYPE

As the number of templates have grows the directory structure has become a little unwieldy. This should help newcomers (and myself) grasp things more easily.


rooftopcellist:
I definitely agree with nesting the templates to make it clear what resources are being created and easier to see at glance where the template you are looking for is.

PR #1016, opened 2022-08-06
Allow for disabling management of status fields on CR

shanemcd:
##### SUMMARY
I was playing around with using the installer role without running the full operator. Something like:

```
podman run -u0 -ti --entrypoint=ansible-runner \
    -v $PWD/env/:/opt/ansible/env:Z \
    -v $HOME/.kube/:/opt/ansible/.kube:Z \
    quay.io/shanemcd/awx-operator:test run -r installer /opt/ansible
```

Where `$PWD/env/extravars` contains:

```
ansible_operator_meta:
  name: local-operator-test
  namespace: local-operator-test
set_self_labels: no
set_self_status: no
ingress_type: route
ansible_python_interpreter: /usr/bin/python3.8
```

We recently added `set_self_labels` in https://github.com/ansible/awx-operator/pull/979, but I also needed the ability to skip `update_status.yml`, since the CR does not exist at all under this scenario.

##### ISSUE TYPE

 - New or Enhanced Feature

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -497,7 +497,11 @@ spec:
                   type: boolean
                   default: true
                 set_self_labels:
-                  description: Maintain some of the recommended `app.kubernetes.io/*` labels on the resource (self)
+                  description: Maintain some of the recommended `app.kubernetes.io/*` labels on the custom resource object
+                  type: boolean
+                  default: true
+                set_self_status:
+                  description: Set some helpful status indicators on the custom resource object
                   type: boolean
                   default: true
               type: object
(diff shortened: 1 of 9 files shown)


config/crd/bases/awx.ansible.com_awxs.yaml
```diff
@@ -497,7 +497,11 @@ spec:
                   type: boolean
                   default: true
                 set_self_labels:
-                  description: Maintain some of the recommended `app.kubernetes.io/*` labels on the resource (self)
+                  description: Maintain some of the recommended `app.kubernetes.io/*` labels on the custom resource object
+                  type: boolean
+                  default: true
+                set_self_status:
```
rooftopcellist, line 503: @shanemcd we might want to add a bit more of a warning here in the description.  If users set this to false and are using the operator via `make deploy` or OLM/OKD, the reconciliation loop will just keep on running. 

Also, can you add an entry in the CSV template?  By default, a new field on the CRD will be added to the form (and not in even in the Advanced section).  

Something like this should work (config/manifests/bases/awx-operator.clusterserviceversion.yaml)
```
      - displayName: Set AWX statuses
        path: set_self_status
        x-descriptors:
```

Similarly, for the AWXBackup and AWXRestore CR's.

Otherwise, I think this is a great addition.  Once the conflicts from the sdk upgrade are resolved, it has my :heavy_check_mark:

PR #1035, opened 2022-08-25
AWX.enable corrected to AWX.enabled

SweetGeneral:
##### ISSUE TYPE

##### SUMMARY

- AWX'S CR doesn't has any field for AWX.enable which mentioned in README.md . Updated the same to AWX.enabled

##### ADDITIONAL INFORMATION

- The README.md in helmchart for awx operator suggests enabling `AWX.enable` whereas the field in the values config is `AWX.enabled`. This PR contains the correction for the same.
Thank you.


rooftopcellist:
Thank you for the fix!  It should indeed be `.enabled`

[Ref: https://helm.sh/docs/topics/charts/]

SweetGeneral:
@rooftopcellist Thanks! Is this good to get merged?

rooftopcellist:
Yep, I just had to make CI happy.  FYI, we added some new workflow logic so the Issue Type (in the PR body) has to be one of the 3 defaults (enforced by a CI check).  We use those issue types when generating the changelog.  

Thanks for your contribution!

PR #413, opened 2021-06-19
Use fsGroup and handles NFS shares

tchellomello:
Fixes: https://github.com/ansible/awx-operator/issues/395
Fixes: https://github.com/ansible/awx-operator/issues/357

This PR accomplishes a couple of things based on different environments. 

When using the `projects_persistence` with a `projects_existing_claim`, depending on the storage provide a `block` device or `filesystem` device will present to the AWX instance writable only by the `root` user.  This introduces a problem as the `awx` user is unable to write under `/var/lib/awx/projects` therefore not being able to sync projects. 

Using the `fsGroup` address the issue for `block` and `fs` mounts so the volume will be mounted allowing the `awx` user to write on it. However, `fsGroup` does not work for `NFS` shares as noted here: https://github.com/kubernetes/examples/issues/260

To work around this issue, we can use the `initContainer` approach to change the permissions and privileges for the mounted volume. 

Please refer to https://github.com/ansible/awx-operator/issues/395#issuecomment-864453580 and https://github.com/ansible/awx-operator/issues/357#issuecomment-863732211 for a full write-up while troubleshooting this issue. 

Another important addition is the ability to the user specify extra volumes as mentioned at https://github.com/ansible/awx-operator/issues/395#issuecomment-864659530

For example:

```yaml

spec:
  replicas: 1 
  ingress_type: none
  service_type: loadbalancer
  projects_existing_claim: awx-testing-nfs-claim
  projects_persistence: true
  extra_volumes: |
    - name: shared-volume
      persistentVolumeClaim:
        claimName: awx-testing-nfs-claim2
  init_container_extra_volume_mounts: |
    - name: shared-volume
      mountPath: /shared
  init_container_extra_commands: |
    # this will ensure the awx user can write on this volume
    chmod 775 /shared
    chgrp 1000 /shared
    date >> /var/lib/awx/projects/been_here
  task_extra_volume_mounts: |
    - name: shared-volume
      mountPath: /shared
```

**TL;DR**

**Before with nfs**
```sh
kubectl iexec awx /bin/bash                                                               15:08:49
Namespace: default | Pod: ✔ awx-nfs-596f6cf8b7-lb4xq
Container: ✔ awx-nfs-web
bash-4.4$ ls -lahd /var/lib/awx/projects
d---------. 1 root root 14 Jun 19 18:43 /var/lib/awx/projects
bash-4.4$ df -h /var/lib/awx/projects
Filesystem                        Size  Used Avail Use% Mounted on
[...]

PR #1051, opened 2022-09-14
Add check for cluster type & improve projects_persistence templating

rooftopcellist:
##### SUMMARY

Project Persistence requires a PVC which can be written to by the awx user.  Certain types of PVC's will default to all of the files in the mount being owned by root.  More details on this are included in this PR which introduced this templating logic - https://github.com/ansible/awx-operator/pull/413

This logic seeks to dynamically check what cluster type awx is deployed on to appropriately template the fsGroup only when applicable.  

##### ISSUE TYPE

### DIFF
--- a/roles/common/defaults/main.yml
+++ b/roles/common/defaults/main.yml
@@ -1,0 +1,8 @@
+---
+deployment_type: awx
+kind: 'AWX'
+api_version: '{{ deployment_type }}.ansible.com/v1beta1'
+
+# Used to determine some cluster specific logic regarding projects_persistence pvc permissions
+is_k8s: false
+is_openshift: false
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -1,0 +1,20 @@
+---
+
+- name: Get information about the cluster
+  set_fact:
+    api_groups: "{{ lookup('k8s', cluster_info='api_groups') }}"
+  when:
+  - not is_openshift | bool
+  - not is_k8s | bool
+
+- name: Determine the cluster type
+  set_fact:
+    is_openshift: "{{ True if 'route.openshift.io' in api_groups else False }}"
+    is_k8s: "{{ False if 'route.openshift.io' in api_groups else True }}"
+  when:
+  - not is_openshift | bool
+  - not is_k8s | bool
+
+# Indicate what kind of cluster we are in (OpenShift or Kubernetes).
+- debug:
+    msg: "CLUSTER TYPE: is_openshift={{ is_openshift }}; is_k8s={{ is_k8s }}"
--- a/roles/installer/templates/deployments/deployment.yaml.j2
+++ b/roles/installer/templates/deployments/deployment.yaml.j2
@@ -53,7 +53,7 @@ spec:
               mkdir -p /etc/pki/ca-trust/extracted/{java,pem,openssl,edk2}
               update-ca-trust
 {% endif %}
-{% if projects_persistence|bool %}
+{% if projects_persistence|bool and is_k8s|bool %}
               chmod 775 /var/lib/awx/projects
               chgrp 1000 /var/lib/awx/projects
 {% endif %}
@@ -79,7 +79,7 @@ spec:
               subPath: bundle-ca.crt
               readOnly: true
 {% endif %}
-{% if projects_persistence|bool %}
+{% if projects_persistence|bool and is_k8s|bool %}
             - name: "{{ ansible_operator_meta.name }}-projects"
               mountPath: "/var/lib/awx/projects"
 {% endif %}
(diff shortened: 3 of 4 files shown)


roles/common/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+          reason: FailedToIdentifyClusterType
+          status: "False"
+          lastTransitionTime: "{{ lookup('pipe', 'date --rfc-3339 seconds') }}"
+
+  - fail:
+      msg: "Cannot determine what type of cluster we are in"
+
+  when:
+  - not is_openshift
+  - not is_k8s
```
dsavineau: this block will never be evaluated since `is_k8s` and `is_openshift` variables are always set with default values and even if this wasn't done then we have the previous task that set both as facts on the same condition.

roles/common/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
+---
+
+- name: Get information about the cluster
+  set_fact:
+    api_groups: "{{ lookup('k8s', cluster_info='api_groups') }}"
+  when:
+  - not is_openshift
+  - not is_k8s
```
dsavineau:
```suggestion
  - not is_openshift | bool
  - not is_k8s | bool
```

roles/common/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+
+- name: Determine the cluster type
+  set_fact:
+    is_openshift: "{{ True if 'route.openshift.io' in api_groups else False }}"
+    is_k8s: "{{ False if 'route.openshift.io' in api_groups else True }}"
+  when:
+  - not is_openshift
+  - not is_k8s
```
dsavineau:
```suggestion
  - not is_openshift | bool
  - not is_k8s | bool
```

roles/common/defaults/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+deployment_type: awx
+kind: '{{ deployment_type | upper }}'
+api_version: '{{ deployment_type }}.ansible.com/v1beta1'
```
dsavineau: If we move those variables from the installer role then we can remove them from that role to avoid duplicate declaration ?

roles/common/vars/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+postgres_initdb_args: '--auth-host=scram-sha-256'
+postgres_host_auth_method: 'scram-sha-256'
+ldap_cacert_ca_crt: ''
+bundle_ca_crt: ''
+projects_existing_claim: ''
+supported_pg_version: 13
```
dsavineau: Is this needed since we don't use any of those variables in the common role ?
If we really need them then we should also remove them from the installer role.

roles/installer/templates/deployments/deployment.yaml.j2
```diff
@@ -53,7 +53,7 @@ spec:
               mkdir -p /etc/pki/ca-trust/extracted/{java,pem,openssl,edk2}
               update-ca-trust
 {% endif %}
-{% if projects_persistence|bool %}
+{% if projects_persistence|bool and is_k8s|bool %}
```
dsavineau, line 56: I'm a bit confused by this modification. Does that mean we can't use projects_persistence with OpenShift ?

dsavineau (changes requested):

roles/common/defaults/main.yml
rooftopcellist: I left them in in case people wanted to run the installer role directly, like in this PR - https://github.com/ansible/awx-operator/pull/1016

roles/common/vars/main.yml
rooftopcellist: Nope, I will remove.  Thanks!

roles/installer/templates/deployments/deployment.yaml.j2
rooftopcellist: @dsavineau These permissions changes are needed when using NFS PVC's for projects_persistence.  We cannot do this in Openshift because of security restrictions (specifically the fsGroup below).  Setting them only for k8s deployments is a hold-over until we find a better solution for Openshift.

roles/common/tasks/main.yml
rooftopcellist: Looks like is was working as it was because I see this in the logs:

```
 TASK [debug] ******************************** 
ok: [localhost] => {
    "msg": "CLUSTER TYPE: is_openshift=False; is_k8s=True"
}
```

I think `is_k8s is not defined` would have that sort of behavior.  Either way, `| bool` is a good idea.  I have added it.

dsavineau (approved):

PR #1063, opened 2022-09-28
change receptor ca secret to tls secret... with migration

TheRealHaoLiu:
##### SUMMARY
change the type of secret use for receptor ca to kubernetes.io/tls secret, to be more "proper"
... with migration

it seems that https://github.com/ansible/awx-operator/pull/1012 was included in release 0.29.0 

this means that people upgrading from 0.29.0 to future release will experience issue

this PR adds migration code for changing the secret 

supersedes https://github.com/ansible/awx-operator/pull/1062 

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
TODO: 
- [x] upgrade from 0.29.0 need to be tested

migration test 
- deploy awx v0.29.0 with 
```
make deploy VERSION=0.29.0
```
- instantiate awx with 
```
ansible-playbook ansible/instantiate-awx-deployment.yml \
    -e image=quay.io/haoliu/awx \
    -e image_version=devel \
    -e image_pull_policy=Always \
    -e namespace=$NAMESPACE \
    -e ingress_type=route
```
- validate that receptor ca secret is created and the content is the old format
```
awx-receptor-ca                                   Opaque                                2      13s
```
- add a remote execution node
- validate that remote execution node is functional by running demo job
- build and upgrade to the code in the PR with 
```
make docker-build docker-push deploy IMAGE_TAG_BASE=quay.io/haoliu/awx-operator
```
- validated that the secret is recreated with the same content in the tls format
```
awx-receptor-ca                                   kubernetes.io/tls                     2      52s
```
- validated that the awx deployment is updated
- validated that remote execution node remain functional after upgrade

### DIFF
--- a/roles/installer/tasks/resources_configuration.yml
+++ b/roles/installer/tasks/resources_configuration.yml
@@ -35,6 +35,50 @@
   register: _receptor_ca
   no_log: "{{ no_log }}"
 
+- name: Migrate Receptor CA Secret
+  when:
+    - _receptor_ca['resources'] | default([]) | length
+    - _receptor_ca['resources'][0]['type'] != "kubernetes.io/tls"
+  block:
+    - name: Delete old Receptor CA Secret
+      k8s:
+        state: absent
+        kind: Secret
+        namespace: '{{ ansible_operator_meta.namespace }}'
+        name: '{{ ansible_operator_meta.name }}-receptor-ca'
+    - name: Create tempfile for receptor-ca.key
+      tempfile:
+        state: file
+        suffix: .key
+      register: _receptor_ca_key_file
+    - name: Copy Receptor CA key from old secret to tempfile
+      copy:
+        content: "{{ _receptor_ca['resources'][0]['data']['receptor-ca.key'] | b64decode }}"
+        dest: "{{ _receptor_ca_key_file.path }}"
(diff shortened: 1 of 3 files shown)


fosterseth (approved):

rooftopcellist:
This PR works and the code is sound.  I have a couple of notes to add though:
* We should back this secret up.  To do so, we will also need to set the name of the secret as a status on the AWX object (like `receptorCaSecret` or something like that).  Here is [an example](https://github.com/ansible/awx-operator/blob/devel/roles/installer/tasks/update_status.yml#L11) of setting the status.
  * Once that status is added, you can just add it to [this list ](https://github.com/ansible/awx-operator/blob/devel/roles/backup/tasks/secrets.yml#L9)and it will automatically be backed up and restored.
* I think it might be a good idea to split the receptor mesh stuff into a separate file who's tasks get included from install.yml (maybe [here](https://github.com/ansible/awx-operator/blob/devel/roles/installer/tasks/install.yml#L54))

That being said, the backup/restore bit can be a separate PR and the other comment is aesthetic.  I am happy with merging this.  Thanks!

rooftopcellist (approved):

PR #1065, opened 2022-09-28
Add workaround to populate index.yaml

janorn:
This will download all releases before cr index is being run. fixes #1053

##### ISSUE TYPE

 - Bug, Docs Fix or other nominal change

##### ADDITIONAL INFORMATION

```
# This step to workaround issues with old releases being dropped.
== CHART FETCH previous releases ==
# Download all old releases
cd .cr-release-packages;\
for tag in 0.6.0 0.7.0 0.8.0 0.9.0 0.10.0 0.11.0 0.12.0 0.13.0 0.14.0 0.15.0 0.16.0 0.16.1 0.17.0 0.18.0 0.19.0 0.20.0 0.20.1 0.20.2 0.21.0 0.22.0 0.23.0 0.24.0 0.25.0 0.26.0 0.27.0 0.28.0 0.29.0; do\
        dl_url="https://github.com/${CHART_OWNER}/${CHART_REPO}/releases/download/${tag}/${CHART_REPO}-${tag}.tgz";\
        curl -RLOs -z "${CHART_REPO}-${tag}.tgz" --fail ${dl_url};\
        result=$?;\
        if [ ${result} -eq 0 ]; then\
                echo "Downloaded $dl_url";\
        fi;\
done
Downloaded https://github.com/ansible/awx-operator/releases/download/0.21.0/awx-operator-0.21.0.tgz
Downloaded https://github.com/ansible/awx-operator/releases/download/0.22.0/awx-operator-0.22.0.tgz
Downloaded https://github.com/ansible/awx-operator/releases/download/0.23.0/awx-operator-0.23.0.tgz
Downloaded https://github.com/ansible/awx-operator/releases/download/0.24.0/awx-operator-0.24.0.tgz
Downloaded https://github.com/ansible/awx-operator/releases/download/0.25.0/awx-operator-0.25.0.tgz
Downloaded https://github.com/ansible/awx-operator/releases/download/0.26.0/awx-operator-0.26.0.tgz
Downloaded https://github.com/ansible/awx-operator/releases/download/0.27.0/awx-operator-0.27.0.tgz
Downloaded https://github.com/ansible/awx-operator/releases/download/0.28.0/awx-operator-0.28.0.tgz
Downloaded https://github.com/ansible/awx-operator/releases/download/0.29.0/awx-operator-0.29.0.tgz
== CHART RELEASER (index) ==
```

PR #1075, opened 2022-10-06
Get tags from remote. Local copy not complete.

janorn:
##### SUMMARY
Due to how the build process checks out the repository all tags are not available when the index is going to be updated.

##### ISSUE TYPE

##### ADDITIONAL INFORMATION
These are the only tags in the shallow checkout.
```bash
$ git tag -l  --sort=creatordate
0.6.0
0.7.0
0.8.0
0.9.0
0.10.0
0.11.0
0.12.0
0.13.0
0.14.0
0.15.0
0.16.0
0.16.1
0.30.0
```
This PR fixes #1065 by changing the way to obtain the tags. New metod queries the remote instead.
```bash
$ git ls-remote --tags --sort=version:refname --refs -q | cut -d/ -f3
0.6.0
0.7.0
0.8.0
0.9.0
0.10.0
0.11.0
0.12.0
0.13.0
0.14.0
0.15.0
0.16.0
0.16.1
0.17.0
0.18.0
0.19.0
0.20.0
0.20.1
0.20.2
0.21.0
0.22.0
0.23.0
0.24.0
0.25.0
0.26.0
0.27.0
0.28.0
0.29.0
0.30.0
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -355,7 +355,7 @@ helm-package: cr helm-chart
 	$(CR) package ./charts/awx-operator
 
 # List all tags oldest to newest.
-TAGS := $(shell git tag -l  --sort=creatordate)
+TAGS := $(shell git ls-remote --tags --sort=version:refname --refs -q | cut -d/ -f3)
 
 # The actual release happens in ansible/helm-release.yml
 # until https://github.com/helm/chart-releaser/issues/122 happens


janorn:
@rooftopcellist could you have a look at this?

miles-w-3:
I think this is an ok solution, however I'm tempted to try and redo how charts are packaged to gh pages. Other charts oh GitHub do not seem to have the limitations with chart releaser, and I've seen some use an automated plugin to keep gh pages up to date. I'm not sure how much free time I'll have in the coming days, but I'll experiment with a streamlined build (potentially without the ansible role) on my fork when I can, and see if there's a concrete reason we need to do it the way it is now. Until then, this change should provide the needed functionality

miles-w-3:
In particular, I think we can use this action to automaticaly package and push the chart to the github-workflows branch https://github.blog/changelog/2021-12-16-github-pages-using-github-actions-for-builds-and-deployments-for-public-repositories/

rooftopcellist (approved):

rooftopcellist:
CI isn't being triggered for some reason.  closing and re-opening to trigger CI.

miles-w-3:
@rooftopcellist do you have thoughts on the potential of either having the helm chart as a separate repo or maintaining a live build of the helm chart within this one? 
Essentially right now we can't use chart releaser because there isn't a living version of the chart for it to compare it to in git tags, it's just built on release

It's also possible that it's just not a good fit since this is an operator, it just means we need to have extra solutions like this

PR #1098, opened 2022-10-24
Enable configuration of route and ingress api versions

sivateja04:
##### SUMMARY

Current version of the operator restricts the installation only on the versions of Kubernetes that use "v1". The new variables "ingress_api_version" and "route_api_version" are introduced to enhance the AWX operator features that provides a better control over the Kubernetes ingress API resources which are running on older versions as well.

##### ISSUE TYPE

 - New or Enhanced Feature

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -103,6 +103,9 @@ spec:
                 - ingress
                 - Route
                 - route
+              ingress_api_version:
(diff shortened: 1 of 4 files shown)


rooftopcellist:
I tested that the route default works on Openshift v4.11.  Siva tested it on k8s, and we have CI coverage there too.  I am happy with merging this.  

```
$ oc get route  -o yaml
apiVersion: v1
items:
- apiVersion: route.openshift.io/v1
  kind: Route
```

PR #1101, opened 2022-10-26
molecule: fix lint configuration

### LINKED ISSUE #1092: Remove lint from molecule

### Bug Summary

Starting 4.0.2 , molecule enforced an schema that is failing to validate `lint` calls.

I talked with molecule developers to see how to go forward and the answer is:

> Remove lint from your molecule files, we will remove it in the future anyway
> running lint with molecule was a historical mistake
> we could fix schema, but in the end we will still remove it

I skiped 4.0.2 in requirements.txt , but we need to fix this sooner the better

### AWX Operator version

latest

### AWX version

latest

### Kubernetes platform

other (please specify in additional information)

latest

### Modifications

no

### Steps to reproduce

Run molecule tests with molecule > 4.0.2

### Expected results

Get a good test run

### Actual results

Schema validation fails in `lint` key, blocking all the test suite

### Additional information

_No response_

### Operator Logs

_No response_

dsavineau:
##### SUMMARY
The `lint` attribute in the provisioner and verifier configuration isn't supported anymore.
This also removes the molecule pin in the requirements.

https://molecule.readthedocs.io/en/latest/configuration.html#lint

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

Closes: #1092


molecule/default/molecule.yml
```diff
@@ -6,15 +6,13 @@ driver:
 lint: |
   set -e
   yamllint .
+  ansible-lint
```
rooftopcellist, line 9: Should this also be included on line 8 of the kind/molecule.yml file?

https://github.com/ansible/awx-operator/pull/1101/files#diff-0aa6fe9c9b445b24185faa9e2ecae5ed7a4829e04bc50cecf8e7fc9c0fb9dadeR8

rooftopcellist (approved):

molecule/default/molecule.yml
rooftopcellist: looks like it fails when put in the kind/molecule.yml.  I'm happy with merging as is.

dsavineau: What kind of failures have you seen ? IMHO we should have the same change in the `kind/molecule.yml` file too, I just missed to added it to that file

rooftopcellist: I ran into the failures when making the same changes on the awx-resource-operator.  But I just tested on the awx-operator locally with `molecule test -s kind` and see them there too:

```WARNING  Listing 2 violation(s) that are fatal
syntax-check: 'command' is not a valid attribute for a Play
molecule/default/kustomize.yml:2:3 [WARNING]: provided hosts list is empty, only localhost is available. Note that
the implicit localhost does not match 'all'
ERROR! 'command' is not a valid attribute for a Play

The error appears to be in '/home/chadams/awx-operator/molecule/default/kustomize.yml': line 2, column 3, but may

---
- name: Build kustomize testing overlay
  ^ here


syntax-check: couldn't resolve module/action 'docker_image'. This often indicates a misspelling, missing collection, or incorrect module path.
molecule/kind/converge.yml:8:7 [WARNING]: provided hosts list is empty, only localhost is available. Note that
the implicit localhost does not match 'all'
ERROR! couldn't resolve module/action 'docker_image'. This often indicates a misspelling, missing collection, or incorrect module path.

The error appears to be in '/home/chadams/awx-operator/molecule/kind/converge.yml': line 8, column 7, but may

  tasks:
    - name: Build operator image
      ^ here
```

We could work through those.

rooftopcellist: Using `community.docker.docker_image` didn't seem to fix it...

dsavineau: I don't see molecule using the kustomize.yml file so I'm wondering if it is required at all and we could remove it. But if it is required maybe we can exclude it from the ansible-lint configuration since ansible-lint wants that file to the be playbook rather than a tasks file. 

the second issue is probably because you didn't install the collection from https://github.com/ansible/awx-operator/blob/devel/molecule/requirements.yml

dsavineau: > I don't see molecule using the kustomize.yml file

```console
$ grep -r 'kustomize.yml' .
./molecule/default/converge.yml:    - import_tasks: kustomize.yml
./molecule/default/destroy.yml:    - import_tasks: kustomize.yml
```
So the file is used, we might want to skip it in the ansible-lint config or move it to another directory so ansible-lint won't consider it as a playbook.

dsavineau: wow after moving the kustomize.yml file to a dedicated tasks directory then ansible-lint is now reporting a lot of things...

```console
                              Rule Violation Summary                              
 count tag                    profile    rule associated tags                     
    26 key-order[task]        basic      formatting, experimental (warning)       
    21 jinja[spacing]         basic      formatting (warning)                     
     4 schema[meta]           basic      core, experimental (warning)             
     3 schema[vars]           basic      core, experimental (warning)             
    34 name[missing]          basic      idiom                                    
     3 name[play]             basic      idiom (warning)                          
     3 name[template]         moderate   idiom                                    
     2 risky-file-permissions safety     unpredictability, experimental (warning) 
    11 no-changed-when        shared     command-shell, idempotency               
   204 fqcn[action-core]      production formatting                               
   100 fqcn[action]           production formatting (warning)                     

Failed after min profile: 252 failure(s), 159 warning(s) on 191 files.
```

rooftopcellist:
@shanemcd do you have an opinion on whether tasks should be fully qualified or just rely on it being set in meta/main.yml for a given role?

PR #1107, opened 2022-11-02
Backup and restore receptor tls secret with expected generated name

rooftopcellist:
##### SUMMARY
Follow-up for https://github.com/ansible/awx-operator/pull/1012

The receptor TLS and Work Signing secrets get generated with expected names:
* `<deployment-name>-receptor-ca`
* `<deployment-name>-receptor-work-signing`

This logic backs up those secrets and restores them.

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
The last problem to solve here is how to get the name of the old deployment in the context of the restore role.  May need to add that to the backup info.

### DIFF
--- a/roles/backup/tasks/secrets.yml
+++ b/roles/backup/tasks/secrets.yml
@@ -1,11 +1,5 @@
 ---
 
-- name: Create Temporary secrets file
-  tempfile:
-    state: file
-    suffix: .json
-  register: tmp_secrets
-
 - name: Dump (generated) secret names from statuses and data into file
   include_tasks: dump_generated_secret.yml
   with_items:
@@ -22,6 +16,12 @@
     - ldap_cacert_secret
     - bundle_cacert_secret
     - ee_pull_credentials_secret
+
+- name: Dump receptor secret names and data into file
+  include_tasks: dump_receptor_secrets.yml
+  loop:
+    - '{{ deployment_name }}-receptor-ca'
+    - '{{ deployment_name }}-receptor-work-signing'
 
 # image_pull_secret is deprecated in favor of image_pull_secrets
 - name: Dump image_pull_secret into file
--- a/roles/installer/tasks/cleanup.yml
+++ b/roles/installer/tasks/cleanup.yml
@@ -23,6 +23,8 @@
         - '{{ _secret_key }}'
         - '{{ _postgres_configuration }}'
         - '{{ _broadcast_websocket_secret }}'
+        - '{{ ansible_operator_meta.name }}-receptor-ca'
+        - '{{ ansible_operator_meta.name }}-receptor-work-signing'
       no_log: "{{ no_log }}"
 
   when: not garbage_collect_secrets | bool
--- a/roles/restore/tasks/deploy_awx.yml
+++ b/roles/restore/tasks/deploy_awx.yml
@@ -1,27 +1,5 @@
 ---
 
-- name: Get AWX object definition from pvc
-  k8s_exec:
-    namespace: "{{ backup_pvc_namespace }}"
-    pod: "{{ ansible_operator_meta.name }}-db-management"
-    command: >-
-      bash -c "cat '{{ backup_dir }}/awx_object'"
-  register: awx_object
-
-- name: Create temp file for spec dict
-  tempfile:
-    state: file
-  register: tmp_spec
-
-- name: Write spec vars to temp file
-  copy:
-    content: "{{ awx_object.stdout }}"
-    dest: "{{ tmp_spec.path }}"
-    mode: '0644'
-
-- name: Include spec vars to save them as a dict
-  include_vars: "{{ tmp_spec.path }}"
-
 - name: Deploy AWX
   k8s:
     state: "{{ state | default('present') }}"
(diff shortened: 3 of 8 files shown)


roles/restore/tasks/secrets.yml
```diff
@@ -54,6 +54,42 @@
...
+
+    - name: Create a new dict of receptor secrets with updated names
+      set_fact:
+        secrets: "{{ secrets | combine({ previous_receptor_ca_name : _ca_secret_name }) }}"
+      no_log: "{{ no_log }}"
+
+    - name: Create a new dict of receptor secrets with updated names
+      set_fact:
+        secrets: "{{ secrets | combine({ previous_receptor_tls_name : _work_signing_secret_name }) }}"
+      no_log: "{{ no_log }}"
```
dsavineau: Can't we just combine both previous_receptor_ca_name and previous_receptor_tls_name in the same set_fact task ? something like (but untested)

```suggestion
    - name: Create a new dict of receptor secrets with updated names
      set_fact:
        secrets: "{{ secrets | combine({previous_receptor_ca_name: _ca_secret_name, previous_receptor_tls_name: _work_signing_secret_name}) }}"
      no_log: "{{ no_log }}"
```

rooftopcellist: @dsavineau Unfortunately not, I tried that and Molecule/ansible-lint complains because they are being set with the same variable (even though it's a combine, it can't tell the difference).

rooftopcellist: Oh, I had them as two separate lines on the same commit.  What you suggest works.

roles/restore/tasks/deploy_awx.yml
```diff
@@ -1,27 +1,5 @@
 ---
 
-- name: Get AWX object definition from pvc
```
rooftopcellist, line 3: These tasks were moved to a new `import_vars.yml` file, which is now called before `secrets.yml` and `awx-cro.yml`.  This is because this `previous_deployment_name` variable needs to be available for use in the secrets.yml file.

roles/restore/tasks/secrets.yml
```diff
@@ -54,6 +54,37 @@
       no_log: "{{ no_log }}"
   when: secrets['postgresConfigurationSecret']['data']['type'] | b64decode == 'managed'
 
+- name: Set new receptor secret names
+  set_fact:
+    previous_receptor_ca_name: "{{ previous_deployment_name }}-receptor-ca"
+    previous_receptor_tls_name: "{{ previous_deployment_name }}-receptor-work-signing"
+  no_log: "{{ no_log }}"
+
+- name: Set new name for receptor secrets using deployment_name
```
rooftopcellist, line 63: The purpose of this block is to edit the json dictionary of secrets so that the deployment name matches that of the new deployment (`deployment_name`), not that of the old deployment (`previous_deployment_name`).

This is necessary because the new deployment expect the secrets to be prepended with the name of the current deployment.

roles/backup/tasks/secrets.yml
```diff
@@ -1,11 +1,5 @@
 ---
 
-- name: Create Temporary secrets file
```
rooftopcellist, line 3: This tempfile was not being used in the backup role, it is probably vestigial from a previous iteration on this code, so I removed it.

shanemcd (approved):

roles/installer/tasks/cleanup.yml
```diff
@@ -23,6 +23,8 @@
         - '{{ _secret_key }}'
         - '{{ _postgres_configuration }}'
         - '{{ _broadcast_websocket_secret }}'
+        - '{{ ansible_operator_meta.name }}-receptor-ca'
+        - '{{ ansible_operator_meta.name }}-receptor-work-signing'
```
rooftopcellist, line 27: Adding in these ownerRef removals is intended to make the following scenario possible:

scenario:
* delete awx CR (without backup) - secrets and pvc stay behind
* recreate awx CR (without restore)

This is some Hao and I just noticed.  FYI @shanemcd

rooftopcellist: Works as expected, here's the output: https://pastebin.com/BTmwBN4P

rooftopcellist:
We discussed how in the future it would be good to improve upon this by doing some of the dictionary manipulation on the restore side with an ansible module instead so we could do it natively in python instead of through a series of set_facts, which would gain us some performance.  This is good enough to merge for now though.

Bluefries:
the variable **{{ previous_deployment_name }}** is not defined in the AWX restore role, therefore, when I try to follow the restore role documentation that says :

_Before Restoring from a backup, be sure to: delete the old existing AWX CR_  [here](https://github.com/ansible/awx-operator/tree/devel/roles/restore)

And when creating my restore object, I get the following :
_The task includes an option with an undefined variable. The error was: 'previous_deployment_name' is undefined_
_The error appears to be in '/opt/ansible/roles/restore/tasks/secrets.yml': line 69, column 3_

Have also tried to declare this variable in the AWXRestore object but it's not taken by the operator or by the role.

@rooftopcellist could you have a look since you introduced this new variable?

Issue #1094, opened 2022-10-20
Permission denied: '/etc/receptor/receptor.conf.lock'


a-dawg:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that the AWX Operator is open source software provided for free and that I might not receive a timely response.

### Bug Summary

using the awx operator  and awx version 21.7.0 I am getting the error below: 
`````
2022-10-20 08:58:25,384 ERROR    [-] awx.main.dispatch Encountered unhandled error in dispatcher main loop
Traceback (most recent call last):
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/dispatch/worker/base.py", line 172, in run
    self.worker.on_start()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/dispatch/worker/task.py", line 140, in on_start
    dispatch_startup()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/system.py", line 86, in dispatch_startup
    write_receptor_config()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/receptor.py", line 653, in write_receptor_config
    with lock:
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/filelock/_api.py", line 220, in __enter__
    self.acquire()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/filelock/_api.py", line 173, in acquire
    self._acquire()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/filelock/_unix.py", line 35, in _acquire
    fd = os.open(self._lock_file, open_mode)
PermissionError: [Errno 13] Permission denied: '/etc/receptor/receptor.conf.lock'
Traceback (most recent call last):
  File "/usr/bin/awx-manage", line 8, in <module>
    sys.exit(manage())
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/__init__.py", line 201, in manage
    execute_from_command_line(sys.argv)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
[...]

PR #1110, opened 2022-11-03
Give receptor.conf write permissions for awx

fosterseth:
##### SUMMARY

related https://github.com/ansible/awx-operator/issues/1094

##### ISSUE TYPE

### DIFF
--- a/roles/installer/templates/deployments/deployment.yaml.j2
+++ b/roles/installer/templates/deployments/deployment.yaml.j2
@@ -317,6 +317,7 @@ spec:
               if [ ! -f /etc/receptor/receptor.conf ]; then
                 cp /etc/receptor/receptor-default.conf /etc/receptor/receptor.conf
                 sed -i "s/HOSTNAME/$HOSTNAME/g" /etc/receptor/receptor.conf
+                chmod 666 /etc/receptor/receptor.conf
               fi
               exec receptor --config /etc/receptor/receptor.conf
           volumeMounts:


rooftopcellist (approved):

TheRealHaoLiu:
i dont think we can run chmod because we are not running this as root (also should not run it as root)

alternative suggestion: 
we can do the change in the dockerfile for the awx-ee image and change `cp` to `cat >` to make sure we preserve permission

roles/installer/templates/deployments/deployment.yaml.j2
```diff
@@ -317,6 +317,8 @@ spec:
               if [ ! -f /etc/receptor/receptor.conf ]; then
                 cp /etc/receptor/receptor-default.conf /etc/receptor/receptor.conf
                 sed -i "s/HOSTNAME/$HOSTNAME/g" /etc/receptor/receptor.conf
+                chown awx /etc/receptor/receptor.conf
+                chmod 664 /etc/receptor/receptor.conf
```
rooftopcellist: We should actually do this in the Dockerfile instead, the cat the receptor.conf into the "stub" receptor.conf file, like we do for rsyslog.conf.

TheRealHaoLiu: at least we can't do chown since awx is not a user that exist

TheRealHaoLiu: i think just the chmod will work but i cant reproduce this issue on my openshift cluster

TheRealHaoLiu: /etc/receptor is a emptyDir mount so when task container and web container mount the emptyDir it should be own by them so ownership shouldnt be a issue

rooftopcellist (approved):

shanemcd:
@fosterseth Have you reproduced the original error? I kinda doubt this will be sufficient.

fosterseth:
@shanemcd it might be a false flag -- we introduced some incompatibility with the execution capacity expansion feature 

awx 21.6+ requires awx operator 0.29.0+

I noticed the user who opened the issue is on 0.28. I asked them to try on later operator versions

roles/installer/templates/deployments/deployment.yaml.j2
timansky: The Pod Security Context supports setting an fsGroup, which allows you to set the group ID that owns the volume, and thus who can write to it.

[More](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/)

fosterseth:
I don't think this is needed anymore

PR #1114, opened 2022-11-08
remove helm chart version from labels

### LINKED ISSUE #1057: Awx-operator Helm upgrade fails due to version labels in selector

### Bug Summary

Hi, 
If we try to run `helm upgrade` for awx-operator for example from 0.28.0 to 0.29.0 version update is failing with the following error: 
```
Error: UPGRADE FAILED: release awx-operator failed, and has been rolled back due to atomic being set: cannot patch "awx-operator-controller-manager" with kind Deployment: Deployment.apps "awx-operator-controller-manager" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"control-plane":"controller-manager", "helm.sh/chart":"awx-operator-0.29.0"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```
The reason for this is a version label used in deployment selector:

```
spec:
  replicas: 1
  selector:
    matchLabels:
      control-plane: controller-manager
      helm.sh/chart: awx-operator-0.29.0
  template:
    metadata:
      annotations:
        kubectl.kubernetes.io/default-container: manager
      labels:
        control-plane: controller-manager
        helm.sh/chart: awx-operator-0.29.0
```

Deployment selector labels are immutable and can't be changed during upgrade [https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates)
In our case this selector changes from one version to another

### AWX Operator version

0.28.0

### AWX version

21.5.0

### Kubernetes platform

kubernetes

1.23.5

### Modifications

no

### Steps to reproduce

run helm upgrade for awx-operator

### Expected results

deployment upgraded

### Actual results

upgrade fails with 
```
Error: UPGRADE FAILED: release awx-operator failed, and has been rolled back due to atomic being set: cannot patch "awx-operator-controller-manager" with kind Deployment: Deployment.apps "awx-operator-controller-manager" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"control-plane":"controller-manager", "helm.sh/chart":"awx-operator-0.29.0"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

### Additional information

_No response_

### Operator Logs

_No response_

FlorianLaunay:
##### SUMMARY
Remove the Helm chart version from `helm.sh/chart` labels in all Helm templates.

Fixes #1057 

##### ISSUE TYPE

##### ADDITIONAL INFORMATION
Including chart version in the `matchLabel` selector of the AWX deployment template causes problem during the `helm upgrade` of the AWX operator. Indeed this field is now immutable for deployments in recent versions of Kubernetes (https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates). So running a `helm upgrade` with a chart containing a deployment template with a different content in the `spec.selector.matchLabels` field is not allowed and causes an error.

An acceptable solution when using `kustomize` is to remove chart version in labels. This will allow to continue to adhere to Helm best practices by having a relationship between `metadata.labels` and `selector.matchLabels` fields.

Tested on `devel` with `make helm-chart` and executing multiple `helm upgrade`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -296,8 +296,8 @@ helm-chart: helm-chart-generate
 helm-chart-generate: kustomize helm kubectl-slice yq charts
 	@echo "== KUSTOMIZE: Set image and chart label =="
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	cd config/manager && $(KUSTOMIZE) edit set label helm.sh/chart:$(CHART_NAME)-$(VERSION)
-	cd config/default && $(KUSTOMIZE) edit set label helm.sh/chart:$(CHART_NAME)-$(VERSION)
+	cd config/manager && $(KUSTOMIZE) edit set label helm.sh/chart:$(CHART_NAME)
+	cd config/default && $(KUSTOMIZE) edit set label helm.sh/chart:$(CHART_NAME)
 
 	@echo "== Gather Helm Chart Metadata =="
 	# remove the existing chart if it exists


rooftopcellist:
@miles-w-3 do you mind taking a look at this? If you are OK with this change, I can merge.

miles-w-3:
I would be in favor of changing the label to just be the chart name rather than including the version, but it is not a functionality blocker to remove it. Generally, the `helm.sh/chart` is a nice way to `get` all resources deployed from a helm chart through kubectl. However, this label isn't present on the custom resources so we're currently not supporting it anyway.

I'm fine with removing it altogether, but maybe the best compromise would be to put just
```
helm.sh/chart: awx-operator
```
on all chart resources.

rooftopcellist:
@FlorianLaunay what do you think about the approach of [changing the label](https://github.com/ansible/awx-operator/blob/devel/Makefile#L299-L300)?

FlorianLaunay:
@miles-w-3 @rooftopcellist TBH I hadn't thought about just removing the version in all labels :innocent: but this is also fine for me. This will allow to continue to adhere to Helm best practices by having a relationship between `metadata.labels` and `selector.matchLabels` fields.
Should I modify this PR or open a new one ?

janorn:
Do it in this PR.

FlorianLaunay:
Rebase and PR modification done

rooftopcellist (approved):
Thanks for the contribution @FlorianLaunay !

PR #1129, opened 2022-11-23
add default cluster name

dru90i:
##### SUMMARY
Added the ability to specify a cluster name for the restore role. Removed duplicate parameter in pg_restore

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxrestores.yaml
+++ b/config/crd/bases/awx.ansible.com_awxrestores.yaml
@@ -50,6 +50,9 @@ spec:
                 description: Name of the restored deployment. This should be different from the original deployment name
                   if the original deployment still exists.
                 type: string
+              cluster_name:
+                description: Cluster name
+                type: string
               backup_name:
                 description: AWXBackup object name
                 type: string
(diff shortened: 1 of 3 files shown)


rooftopcellist:
Thanks for the contribution!  This might be good change for the migrate.yaml as well.

rooftopcellist:
@dru90i Could you tell me more about your use case? 

Typically, you would take a backup before doing upgrade, and in the case that the upgrade had issues, you could restore from your backup.  So you would be on the same cluster, therefore no need to specify a different cluster.  

If you want to migrate to a different cluster, I would suggest taking a look at the migration doc: 
* https://github.com/ansible/awx-operator/blob/devel/docs/migration.md

dru90i:
Yes, migration is suitable for cluster change. In our case, it was necessary to restore from a backup in a test cluster. The database could not recover due to the fact that the name of the service could not resolved, since clusterDomain (cluster name) was different from cluster.local. In addition, you can make the same edits to the backup role, since with a clusterDomain other than cluster.local, a backup cannot be created. In most cases, this will not have to be done since cluster.local is usually used.

Issue #1132, opened 2022-11-25
BUG on Receptor configuration - No verifysignature param


tanganellilore:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that the AWX Operator is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Good Morning Team,

I dicover a bug on Receptor configuration ConfigMap during the deployment of AWX. 
This bug can be refferred to AWX commit: [84f2b91105c959c4d89a63063cca441f3d67fc0f](https://github.com/ansible/awx/commit/84f2b91105c959c4d89a63063cca441f3d67fc0f) this [PR](https://github.com/ansible/awx/pull/13200) where introduced the  `sign_work` variable and `work_signing_enabled(config_data)` fuction that check if `work-verification` section is preset in the `/etc/receptor/receptor.conf` .

To align `awx-operator` to this commint we need to add [awx-operator/roles/installer/templates/configmaps/config.yaml.j2](https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/configmaps/config.yaml.j2), in `receptor_conf ` section the following part:
Adding  on key `work-command`, `work-kubernetes`, `work-kubernetes`
```yaml
verifysignature: true
```
Adding this section
```yaml
- work-verification:
    publickey: /etc/receptor/work_public_key.pem
```

Moreover, another issue is realted the installation bundle of execution_node. Due to the fact that `sign_work` is not poulated, the respective `verifysignature` key in the bundle is empty, and this create issue on the receptor configuration.

I will try to propose you a PR to fix this issue in a short time and I will try tro introduce a variable to enable/disable the `sign_work`.

### AWX Operator version

1.1.0

### AWX version

21.9.0

### Kubernetes platform

kubernetes

### Kubernetes/Platform version

v1.23.2

### Modifications

no

### Steps to reproduce

Update or install 21.9.0 and add new execution node.

Download the instalaltion bundle and you will se that verifysignature is missing.
If you enable it, execution node not work, because controller node not have the same setting

### Expected results

verifysignature enabled and working

### Actual results

Remote error: ERROR: could not parse response: ERROR: could not verify signature: signature is empty

AlanCoding:
[...]

PR #1133, opened 2022-11-25
Add sign_work and fix Configmap

tanganellilore:
##### SUMMARY
Refer to issue #1132

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/config/manifests/bases/awx-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/awx-operator.clusterserviceversion.yaml
@@ -777,6 +777,7 @@ spec:
   - email: awx-project@googlegroups.com
     name: AWX Team
   maturity: alpha
+  MinKubeVersion: 1.22.15
   provider:
     name: Ansible
     url: github.com/ansible/awx-operator


rooftopcellist:
We think the issue is that this variable need quotes to be evaluated correctly:
* https://github.com/ansible/awx/pull/13200/files#diff-ff993804c2a942f24d9d1a982df299fb18dadf0a53237363a25aaa4a62d053feR10

As is, it is probably evaluating to an empty string, thus the error.  Let's try to fix that fix, if so, we may not need this PR.

tanganellilore:
Nope, this is not related to quote.
The problem here is related to the default receptor.conf that with actual operator not have the reference on `work-verification`.
This, with the last pull request on awx, where was introduce `sign_work` and some rework on awx api receptor side, created this bug.

With this implementation, we allow to create a receptor.conf deployed by awx-operator with correct keyword used by awx backend for receptor bundle.

rooftopcellist:
cc @fosterseth what do you think?

PR #1142, opened 2022-12-01
add permission to get openshift version

TheRealHaoLiu:
Signed-off-by: Hao Liu <haoli@redhat.com>

##### SUMMARY
add permission to get config.openshift.io/clusteroperator resource on a openshift cluster to determine openshift-apiserver version

openshift apiserver version will be used in https://github.com/ansible/receptor/pull/693 for determine if to use --timestamp to resume when log stream is disconnected

NOTE: since the ClusterRoe `config/rbac/cluster_operator_viewer_clusterrole.yaml` is added to the install bundle that means the name will be prefix by `namePrefix` in `config/default/kustomization.yaml`  we use the operator's pod name with some unfortunate string parsing to get this prefix

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -45,6 +45,10 @@ spec:
           value: explicit
         - name: ANSIBLE_DEBUG_LOGS
           value: 'false'
+        - name: MY_POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
         - name: WATCH_NAMESPACE
           valueFrom:
             fieldRef:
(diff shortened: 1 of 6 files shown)


TheRealHaoLiu:
ok this wont work...

ClusterOperator is a cluster scoped CRD in order to allow awx to get that resource awx-operator will need to be able to create a clusterrole and clusterrolebinding 

we can have OLM create the clusterrole as a part of the bundle but we can not get around the problem of having to create clusterrolebinding... we dont want to give the operator permission to create either clusterrole or clusterrolebinding because that will open up potential attack vectors for privilege escalation

PR #1144, opened 2022-12-02
Add sign_work and fix Configmap

tanganellilore:
##### SUMMARY
Refer to issue https://github.com/ansible/awx-operator/issues/1132

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

This is a replacement of #1133

### DIFF
--- a/roles/installer/templates/configmaps/config.yaml.j2
+++ b/roles/installer/templates/configmaps/config.yaml.j2
@@ -277,3 +277,5 @@ data:
     - work-signing:
         privatekey: /etc/receptor/signing/work-private-key.pem
         tokenexpiration: 1m
+    - work-verification:
+        publickey: /etc/receptor/signing/work-public-key.pem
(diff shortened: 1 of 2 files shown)


roles/installer/templates/configmaps/config.yaml.j2
```diff
@@ -277,3 +277,5 @@ data:
     - work-signing:
         privatekey: /etc/receptor/signing/work-private-key.pem
         tokenexpiration: 1m
+    - work-verification:
+        publickey: /etc/receptor/signing/work-public-key.pem
```
AlanCoding, line 281: I made a prior comment that I agreed that `- work-verification:` should be added to this config map. I was basing that on a few mistaken assumptions.

In https://github.com/ansible/awx/pull/13213, I add the work signing bits if "work-verification" is in the config.

I now believe that should have checked if "work-signing" is in the config. The control node is the work unit _submitter_ and the execution node is the _receiver_. The submitter needs to sign and the receiver needs to verify. It matters not whether the submitter lists _any work types at all_, and verification only matters if you allow other nodes to submit work to you.

In docker-compose, in the receptor configuration, we conditionally exclude "work-signing" if work signing is turned off (and work-verification by the same criteria). It would not be recommended, and probably shouldn't be allowed to turn that off in OCP.

tanganellilore: I agree with you. If in awx side you will change `work-verification` with `work-signing` in the if section, in not requiared any change.

tanganellilore:
not anymore required

PR #1154, opened 2022-12-12
Allow partial changes for managed PostgreSQL

### LINKED ISSUE #1052: Creating a managed postgres db with custom password

### Bug Summary

When creating a postgres secret with only a custom password the ansible database_configuration tasks fails. According to the [readme ](https://github.com/ansible/awx-operator#external-postgresql-service) this should be possible.
```
resource "kubernetes_secret" "main" {
  metadata {
    namespace = var.kubernetes_namespace
    name      = "awx-postgres-configuration"
  }
  data = {
    "password" = random_password.password.result
    "type"     = "managed"
  }
}
```
```
      spec:
        postgres_configuration_secret: awx-postgres-configuration
```

https://github.com/ansible/awx-operator/blob/e8096a5f33dae40d0fbcc2cd6ad789e51dacb2d9/roles/installer/tasks/database_configuration.yml#L95-L103
```
TASK [installer : Store Database Configuration] ********************************
task path: /opt/ansible/roles/installer/tasks/database_configuration.yml:95
fatal: [localhost]: FAILED! => {\"censored\": \"the output has been hidden due to the fact that 'no_log: true' was specified for this result\"}

localhost                  : ok=31   changed=0    unreachable=0    failed=1    skipped=18   rescued=0    ignored=0
```

### AWX Operator version

0.28.0

### AWX version

Never got that far...

### Kubernetes platform

kubernetes

v1.23.5

### Modifications

no

### Steps to reproduce

Create a custom postgres password.
```
resource "kubernetes_secret" "main" {
  metadata {
    namespace = var.kubernetes_namespace
    name      = "awx-postgres-configuration"
  }
  data = {
    "password" = random_password.password.result
    "type"     = "managed"
  }
}
```
```
      spec:
        postgres_configuration_secret: awx-postgres-configuration
```

### Expected results

A postgres pod running using my defined password.

### Actual results

Ansible fails configuring the database.

### Additional information

```
TASK [installer : Store Database Configuration] ********************************
task path: /opt/ansible/roles/installer/tasks/database_configuration.yml:95
fatal: [localhost]: FAILED! => {\"censored\": \"the output has been hidden due to the fact that 'no_log: true' was specified for this result\"}

localhost                  : ok=31   changed=0    unreachable=0    failed=1    skipped=18   rescued=0    ignored=0
```

### Operator Logs

```
TASK [installer : Store Database Configuration] ********************************
task path: /opt/ansible/roles/installer/tasks/database_configuration.yml:95
fatal: [localhost]: FAILED! => {\"censored\": \"the output has been hidden due to the fact that 'no_log: true' was specified for this result\"}

localhost                  : ok=31   changed=0    unreachable=0    failed=1    skipped=18   rescued=0    ignored=0
```

justsomescripts:
##### SUMMARY

The README.md implies that it's possible to change single values (like username or password only) while still using the managed PostgreSQL database. This does currently not work as expected due to how the install role is set up.

This commit splits the variables into single values to work around the issue.

fixes #1052

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

Ansible errors:

```
TASK [installer : Read Database Configuration] *********************************
task path: /opt/ansible/roles/installer/tasks/database_configuration.yml:77

-------------------------------------------------------------------------------
{"level":"error","ts":1670860073.1093168,"logger":"logging_event_handler","msg":"","name":"awx","namespace":"dg-awx-custom","gvk":"awx.ansible.com/v1beta1, Kind=AWX","event_type":"runner_on_failed","job":"4885564860874173450","EventData.Task":"Store Database Configuration","EventData.TaskArgs":"","EventData.FailedTaskPath":"/opt/ansible/roles/installer/tasks/database_configuration.yml:95","error":"[playbook task failed]"}

--------------------------- Ansible Task StdOut -------------------------------

 TASK [Store Database Configuration] ******************************** 
fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'host'\n\nThe error appears to be in '/opt/ansible/roles/installer/tasks/database_configuration.yml': line 95, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Store Database Configuration\n  ^ here\n"}
```


justsomescripts:
After some further tests, the issue isn't solved completely by this unfortunately.

justsomescripts:
Tested on a full AWX deployment now, everything seems to work.

rooftopcellist:
@justsomescripts Is there a reason why pre-creating a postgres_configuration_secret with `type: managed` wouldn't work?

If you need to change a value after the fact, you could delete the secret and created it again with a different value.  Here is some example yaml you can try it out with:

```
---
apiVersion: v1
kind: Secret
metadata:
  name: custom-pg-secret
stringData:
  database: 'awx'
  host: 'my-postgres'
  password: 'test'
  port: '5432'
  type: 'managed'
  username: 'awx'
```

justsomescripts:
Yes, this is exactly what I wanted to do, it doesn't work in the current state and produces the error I mentioned. Should have described that more clearly.

It only works if the pre-created secret contains all values that the default one would (except sslmode, that has a default defined), including hostname for example what we don't want in our deployment.

We want something like

```
---
apiVersion: v1
kind: Secret
metadata:
  name: custom-pg-secret
stringData:
  database: 'awx'
  password: 'test'
  type: 'managed'
  username: 'awx'
```

Where hostname and port are set to the current defaults, while the custom values override the generates secret, which is how it would work with my PR.

justsomescripts:
Is there a chance this is reviewed again? If you don't like this approach, maybe it would make sense to map certain fields using a ConfigMap and only provide the actual secrets in the PSQL configuration secret, so you end up with something like

```yaml
---
apiVersion: v1
kind: Secret
metadata:
  name: postgres-secret
data:
  password: dGVzdAo=
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: postgres-configuration
data:
  database: 'awx'
  host: 'my-postgres'
  port: '5432'
  type: 'managed'
  username: 'awx'
```

This would definitely improve secret handling in case you have external secret management systems in place and also is more in line with the initial [idea how secrets and configmaps](https://stackoverflow.com/a/36925553) should be used in K8s.

So I guess the best approach would be to leave the Ansible code untouched here, split the PSQL config into a ConfigMap and Secret resource and define default values on K8s side to allow the partial changes. What do you think?

PR #1158, opened 2022-12-15
use awx-ee tag that matches awx version

fosterseth:
##### SUMMARY

pairs with https://github.com/ansible/awx/pull/13338

when AWX is released, awx-ee:latest will be tagged with the AWX release tag
e.g. `quay.io/ansible/awx-ee:21.10.2`

operator will use the awx-ee image that corresponds to the specified AWX image, which set by the `DEFAULT_AWX_VERSION` environment variable

##### ISSUE TYPE

 - New or Enhanced Feature

### DIFF
--- a/molecule/default/molecule.yml
+++ b/molecule/default/molecule.yml
@@ -23,6 +23,7 @@ provisioner:
       localhost:
         awx_image: ${AWX_TEST_IMAGE:-""}
         awx_version: ${AWX_TEST_VERSION:-""}
+        awx_ee_image: ${AWX_EE_TEST_IMAGE:-""}
         default_awx_version: "{{ lookup('url', 'https://api.github.com/repos/ansible/awx/releases/latest') | from_json | json_query('tag_name') }}"
         ansible_python_interpreter: '{{ ansible_playbook_python }}'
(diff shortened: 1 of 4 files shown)


fosterseth:
@shanemcd CI fails because quay.io/ansible/awx-ee:21.10.2 doesn't exist. After we merge in the other associated PR and do an official AWX release (thus creating the awx-ee tag), then this CI should pass and we can merge it in.

To test this I changed the image to quay.io/fosterseth/awx-ee (after creating a tag there) and the CI passed fine.

TheRealHaoLiu:
we still publish 'latest' right?

i think the CI should still use 'latest'

TheRealHaoLiu (approved):

rooftopcellist (approved):

rooftopcellist:
For now, we have decided to leave things as they are, so the default awx-ee image used will be `quay.io/ansible/awx-ee:latest`.  Users have the ability to override this image on the spec if desired.

```
spec:
  ee_images:
    - name: Custom AWX EE
      image: {{ awx_ee_image }}
  control_plane_ee_image: {{ awx_ee_image }}
```

PR #1160, opened 2022-12-16
Add `additional_labels` parameter

stanislav-zaprudskiy:
##### SUMMARY

In certain situations labeling of Kubernetes objects managed by the operator might be desired (e.g. for owner identification purposes). For that `inherited_labels` parameter could be used

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

When existing installation is updated with `inherited_labels` not all objects are automatically labelled, but only those which got updated during the role execution. For instance, `broadcast-websocket`, `receptor-ca` and `receptor-work-signing` Secrets aren't labeled under normal circumstances - as they are accompanied by existence check, which makes their management to be skipped if a resource exists. While e.g. `ServiceAccount`, `Service` and `Deployment` seem to always get new labels.

Comparing template data with remote data won't be a solution, as templates are dynamically populated during the run and will always be different from the remote state. Another idea would be to compare and update labels specifically - thoughts?


fosterseth:
hi @stanislav-zaprudskiy thanks for the PR submission, we'll be taking a look at it in detail shortly and follow up with you if it needs adjustments

thanks

rooftopcellist:
@stanislav-zaprudskiy This PR looks great, thank you for taking the time to put it together.  A couple things:

* would you be OK with renaming the parameter from `inherited_labels` to `additional_labels`? (cc @TheRealHaoLiu)
* it looks like the molecule CI tests are failing, those need to be resolved before merge
* (optional) it might be good to add a dummy `additional_labels` to the AWX CR used by molecule ([link](https://github.com/ansible/awx-operator/blob/devel/molecule/default/templates/awx_cr_molecule.yml.j2#L13)) so that we have some CI coverage for this new parameter.

roles/installer/tasks/install.yml
```diff
@@ -10,9 +10,28 @@
       metadata:
         name: '{{ ansible_operator_meta.name }}'
         namespace: '{{ ansible_operator_meta.namespace }}'
-        labels: '{{ lookup("template", "labels/common.yaml.j2") | from_yaml }}'
+        labels: '{{ lookup("template", "roles/common/templates/labels/common.yaml.j2") | from_yaml }}'
```
rooftopcellist: It looks like CI is failing because it can't find this template.


```
[installer : Start installation] ******************************************\\r\\ntask path: /opt/ansible/roles/installer/tasks/main.yml:11\\nincluded: /opt/ansible/roles/installer/tasks/install.yml for localhost\\n\\r\\nTASK [installer : Patching labels to AWX kind] *********************************\\r\\ntask path: /opt/ansible/roles/installer/tasks/install.yml:2\\n[WARNING]: Unable to find 'roles/common/templates/labels/common.yaml.j2' in\\r\\nexpected paths (use -vvvvv to see paths)\\nfatal: [localhost]: FAILED! => {\\\"msg\\\": \\\"An unhandled exception occurred while running the lookup plugin 'template'. Error was a <class 'ansible.errors.AnsibleError'>, original message: the template file roles/common/templates/labels/common.yaml.j2 could not be found for the lookup\\\"}
```

You may have to use the relative path here.

stanislav-zaprudskiy:
I updated the PR based on the feedback provided in:
* https://github.com/ansible/awx-operator/pull/1160#pullrequestreview-1244902851
* https://github.com/ansible/awx-operator/pull/1160#issuecomment-1379479267

Would you have a look again?

molecule/default/tasks/awx_test.yml
```diff
@@ -83,3 +83,61 @@
...
+             | dict2items | selectattr('key', 'in', this_awx.resources[0].spec.additional_labels)
+             | list
+          }}
+
+    - name: AWX Pod contains additional_labels
+      ansible.builtin.assert:
+        that:
+          - pod_additional_labels == awx_additional_labels
+
+    - name: Exract labels which should not be propagated to AWX Pod
```
rooftopcellist: Small typo here:

```suggested
    - name: Extract labels which should not be propagated to AWX Pod
```

rooftopcellist: There are a couple other places above.

rooftopcellist:
@stanislav-zaprudskiy other than a couple small typos, this looks good to go.  Thank you for adding tests!   ++

I will wait to merge until tomorrow in case you want to push a change for the typos.  Thank you for your contribution!

molecule/default/tasks/awx_test.yml
stanislav-zaprudskiy: I didn't get this, but I tried to rephrase it to better reflect what's inside the task - https://github.com/ansible/awx-operator/pull/1160/commits/7bfb31cb8ef0709e2fdbfda0a3405a163589401a

stanislav-zaprudskiy:
> I will wait to merge until tomorrow in case you want to push a change for the typos

I changed the wording a bit, but still not sure that I got you right. Feel free to complement, elaborate or just process as is :). Thank you!

molecule/default/tasks/awx_test.yml
rooftopcellist: @stanislav-zaprudskiy sorry, I should have  been more specific.  It's just a minor spelling typo: `Exract` --> `Extract`

rooftopcellist:
Thanks for the contribution!

jsalatiel:
What's the correct syntax on this one?
i have tried:
```
  additional_labels:
  - role: "frontend"
```
But i get 
```
Error from server (BadRequest): error when creating "a": AWX in version "v1beta1" cannot be handled as a AWX: strict decoding error: unknown field "spec.additional_labels[0].role"
```

stanislav-zaprudskiy:
@jsalatiel, this would be:

```yaml
additional_labels:
- role
```

whereas the `role` label should also be set for AWX object itself - so its value is picked up and applied for managed objects.

See full example here - https://github.com/ansible/awx-operator#labeling-operator-managed-objects.

sangmin-lee-woowahan:
@stanislav-zaprudskiy 
Does this add labels to the task and web pod?
I'm using the AWX helm chart and I added the following, but nothing happened.
```
AWX:
  enabled: true
  name: awx
  labels:
    role: "awx"
  spec:
    additional_labels:
    - role
```

PR #1172, opened 2022-12-27
update Trusting a Custom Certificate Authority

orcema:
##### SUMMARY
ReadMe upate for adding custom certificate bundle using the Kustomize file instead of the CLI. The readme only referes to the CLI command option

##### ADDITIONAL INFORMATION
I had hard times to identify how to declare to include statement for a custom certifcate bundle within the Kustomize file.
The tricky part for me was to spot the option "disableNameSuffixHash: true" in order to avoid renaming the secret name with an has suffix

### DIFF
--- a/README.md
+++ b/README.md
@@ -813,7 +813,22 @@ spec:
   bundle_cacert_secret: <resourcename>-custom-certs
 ```
 
-To create the secrets, you can use the commands below:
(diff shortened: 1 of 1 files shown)


shanemcd (approved):
Nice addition. Thanks!

PR #1145, opened 2022-12-04
Add ability to configmap and include config files for nginx.conf

rooftopcellist:
##### SUMMARY

Implements: https://github.com/ansible/awx-operator/issues/1112

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
This allows a user to add custom configuration the the `nginx.conf` file used by the awx-web container.

1. Create a ConfigMap with the extra settings you want to include in the nginx.conf
2. Create an extra_volumes entry in the AWX spec for this ConfigMap
3. Create an web_extra_volume_mounts entry in the AWX spec to mount this volume

PR #1173, opened 2022-12-28
add uwsgi.ini in configmap and deployment awx

dru90i:
##### SUMMARY
The uwsgi configuration has been added to config map. Allows you to change the uwsgi settings without rebuilding the image. For example, to limit memory using limit-as or reload-on-as. Enable stats-server, etc.

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

### DIFF
--- a/roles/installer/templates/configmaps/config.yaml.j2
+++ b/roles/installer/templates/configmaps/config.yaml.j2
@@ -232,6 +232,27 @@ data:
             }
         }
     }
+  uwsgi.ini: |
+    [uwsgi]
+    socket = 127.0.0.1:8050
+    processes = 5
+    master = true
+    vacuum = true
+    no-orphans = true
+    lazy-apps = true
+    manage-script-name = true
+    master-fifo = /var/lib/awx/awxfifo
+    max-requests = 1000
+    buffer-size = 32768
+
+    if-env = UWSGI_MOUNT_PATH
+    mount = %(_)=awx.wsgi:application
+    endif =
+
+    if-not-env = UWSGI_MOUNT_PATH
+    mount = /=awx.wsgi:application
+    endif =
+
   redis_conf: |
     unixsocket /var/run/redis/redis.sock
     unixsocketperm 777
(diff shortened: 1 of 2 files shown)


shanemcd:
I don't see any configurable options here. If you modify the configmap directly, the operator will run its reconciliation loop and overwrite any changes you make.

dru90i:
Why does the operator overwrite configmap if an additional config is placed in the configmap template? Initially, I needed to enable stats-server for uwsgi (https://uwsgi-docs.readthedocs.io/en/latest/StatsServer.html ) to do this, I needed to make changes to uwsgi.ini. I added it to configmap and mounted it in a container. It may be necessary for someone else to make uwsgi settings directly in configmap.

shanemcd:
This is inherent to how operators work in general, not just this one. Resources managed by an operator are expected to match the "desired state". Customizations are typically made by setting variables on the custom resource. You can see dozens of examples of other configurable options by looking at the variables documented in the readme and reading the corresponding code.

rooftopcellist:
@dru90i As @shanemcd mentioned, none of the parameters in the configmap you added for uwsgi are configurable via the AWX spec.  You could add that in for params you would like to make configurable.  

But I wonder if using the same pattern we did for the nginx.conf would work here ([link to example](https://github.com/ansible/awx-operator/pull/1145/files#diff-a9e274ca586fa68515be7144d74b9233148ce5f761b2e25b27ab62739b73c3bfR221)).  We could add an include in the uwsgi config, then the user could configmap in an nginx.ini to be used.  

It should also be noted that these particular parameters would get overidden as is when the `launch_awx.sh` is run (entrypoint of the awx container).  
* https://github.com/ansible/awx/blob/devel/Makefile#L202-L218

kdelee:
@rooftopcellist I suggest we close this and @dru90i can follow up https://github.com/ansible/awx-operator/pull/1487 with adding any additional uwsgi configurables as variables exposed on the CR now that we do mount a uwsgi config at deploy time

PR #1189, opened 2023-01-16
[web/task split] split web and task deployment + a few supporting bits

### LINKED ISSUE #1182: [web/task split] AWX operator change to split web and task into separate deployment
## Description
starting point https://github.com/ansible/awx-operator/tree/web-task-split

## TODO:
- [x] split up web and task into separate deployment 
- [x] add rsyslog sidecar to both web and task deployment
- [x] add toleration, web_toleration and task_toleration to allow web and task to be schedule on specific nodes https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/
- [x] add affinity, web_affinity and task_affinity to 
- [x] add node_selector, web_node_selector, task_node_selector
- [x] add replicas, task_replicas, web_replicas in CRD to separately configure web and task if specified
- [x] https://github.com/ansible/awx-operator/blob/web-task-split/roles/installer/tasks/main.yml/#L2-L8 make this respect both web and task deployment
- [x] add topology_spread_constraint and task_toplogy_spread_constraints and web_toplogy_spread_constraints on the web deployment
- [x] figure out if project dir need to be on web container [yes/no]
- [ ] should we allow task container to have a larger resource request/limit? because it will likely use more 
- [x] https://github.com/ansible/awx-operator/blob/devel/config/manifests/bases/awx-operator.clusterserviceversion.yaml update CSV with all the new paramaters
- [x] Document new paramaters and change in behavior in README

OPTIONAL changes
- [ ] create new  serviceaccount/role/rolebinding  for web deployment

thedoubl3j:
##### SUMMARY

Fixes portions of #1182

Specifically:
- splitting web and task containers into 2 deployments
- adding rsyslog sidecar container to each deployment
- add toleration, web_toleration and task_toleration to allow web and task to be schedule on specific nodes
- add affinity, web_affinity and task_affinity 
- add node_selector, web_node_selector, task_node_selector

##### ISSUE TYPE

 - New or Enhanced Feature
##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -89,6 +89,22 @@ topology_spread_constraints: ''
 #     effect: "NoSchedule"
 tolerations: ''
 
+# Add node tolerations for the task pods. Specify as literal block. E.g.:
+# tolerations: |
+#   - key: "dedicated"
+#     operator: "Equal"
+#     value: "AWX"
+#     effect: "NoSchedule"
+task_tolerations: ''
+
+# Add node tolerations for the web pods. Specify as literal block. E.g.:
+# tolerations: |
+#   - key: "dedicated"
+#     operator: "Equal"
+#     value: "AWX"
+#     effect: "NoSchedule"
+web_tolerations: ''
+
 # Add annotations to awx pods. Specify as literal block. E.g.:
 # annotations: |
 #   my.annotation/1: value
@@ -177,6 +193,9 @@ task_command: []
 web_args:
   - /usr/bin/launch_awx.sh
 web_command: []
+ryslog_args:
+  - /usr/bin/launch_awx_rsyslog.sh
+rsyslog_command: []
 
 task_resource_requirements:
   requests:
@@ -192,6 +211,12 @@ ee_resource_requirements:
   requests:
     cpu: 100m
     memory: 64Mi
+
+# TODO: validate defaul resource requirements
+rsyslog_resource_requirements:
+  requests:
+    cpu: 100m
+    memory: 128Mi
 
 # Customize CSRF options
 csrf_cookie_secure: False
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -90,20 +90,56 @@ topology_spread_constraints: ''
 tolerations: ''
 
 # Add node tolerations for the task pods. Specify as literal block. E.g.:
-# tolerations: |
+# task_tolerations: |
 #   - key: "dedicated"
 #     operator: "Equal"
-#     value: "AWX"
+#     value: "AWXtask"
 #     effect: "NoSchedule"
 task_tolerations: ''
 
 # Add node tolerations for the web pods. Specify as literal block. E.g.:
-# tolerations: |
+# web_tolerations: |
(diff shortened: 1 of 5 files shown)


roles/installer/templates/deployments/task.yaml.j2
```diff
@@ -269,6 +265,43 @@ spec:
 {% endif %}
 {% if ee_extra_env -%}
             {{ ee_extra_env | indent(width=12, first=True) }}
+{% endif %}
+        - image: '{{ _image }}'
```
erz4, line 269: any way to add some probes?

TheRealHaoLiu: liveness and readiness probe?

roles/installer/defaults/main.yml
```diff
@@ -89,6 +89,22 @@ topology_spread_constraints: ''
...
+# Add node tolerations for the task pods. Specify as literal block. E.g.:
+# tolerations: |
+#   - key: "dedicated"
+#     operator: "Equal"
+#     value: "AWX"
+#     effect: "NoSchedule"
+task_tolerations: ''
+
+# Add node tolerations for the web pods. Specify as literal block. E.g.:
+# tolerations: |
```
TheRealHaoLiu: web_tolerations

roles/installer/defaults/main.yml
```diff
@@ -89,6 +89,22 @@ topology_spread_constraints: ''
...
+#   - key: "dedicated"
+#     operator: "Equal"
+#     value: "AWX"
```
TheRealHaoLiu: value: "AWXWeb"

roles/installer/templates/deployments/web.yaml.j2
```diff
@@ -190,6 +186,47 @@ spec:
             {{ web_extra_env | indent(width=12, first=True) }}
 {% endif %}
           resources: {{ web_resource_requirements }}
+        - image: '{{ _image }}'
+          name: '{{ ansible_operator_meta.name }}-rsyslog'
+{% if rsyslog_command %}
+          command: {{ rsyslog_command }}
```
Zokormazo, line 192: will we ever reach here without `rsyslog_command` or `rsyslog_args` ? what happens in that case? we launch a container with the default command/arg called `myawx-rsyslog` ? I perhaps lost the context here, but these two conditionals, either should not exist or should condition the full existence of this container in the list and not just `command` or `args` fields imo

roles/installer/tasks/resources_configuration.yml
```diff
@@ -250,6 +250,8 @@
         tower_pod_name: '{{ _new_pod["resources"][0]["metadata"]["name"] }}'
   when:
     - tower_resources_result.changed or this_deployment_result.changed
+- debug:
```
TheRealHaoLiu, line 253: should this be here

TheRealHaoLiu (approved):

roles/installer/tasks/resources_configuration.yml
thedoubl3j: no, will remove in follow up PR

PR #1197, opened 2023-01-19
Update README.md

rakesh561:
Added procedure to disable ipv6 during install or upgrade which was was not mentioned in the doc.

SUMMARY

ISSUE TYPE

##### SUMMARY

By Default deployment tries to enable ipv6 for new install and upgrade.Recently flag was introduced to disable ipv6 but its not mentioned in the README. This will mainly help for people who are trying to upgrade or install awx from AWX 1.1.1

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

Users are running into errors with new install and upgrade ever since ipv6 had been introduced to enable

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ An [Ansible AWX](https://github.com/ansible/awx) operator for Kubernetes built w
             * [Cluster-scope to Namespace-scope considerations](#cluster-scope-to-namespace-scope-considerations)
             * [Project is now based on v1.x of the operator-sdk project](#project-is-now-based-on-v1x-of-the-operator-sdk-project)
             * [Steps to upgrade](#steps-to-upgrade)
-       * [Ipv6](#ipv6-disable-new-installs)
+       * [Disable IPV6 for install and upgrade](#ipv6-disable-new-installs)
       * [Add Execution Nodes](#adding-execution-nodes)
           * [Custom Receptor CA](#custom-receptor-ca)
    * [Contributing](#contributing)
@@ -1286,7 +1286,7 @@ Then install the new AWX Operator by following the instructions in [Basic Instal
 
 Once the new AWX Operator is up and running, your AWX deployment will also be upgraded.
 
-# Disable IPV6 for install and upgrade
+#### Disable IPV6 for install and upgrade
 Starting with  AWX Operator release 0.24.0 ipv6 had been enabled in ngnix configuration which caused upgrades/new install's to fail.
 As a work around until AWX Operator release 1.1.1 and AWX v21.10.0 it is suggested to enable ipv6 on worker nodes from 1.1.1 release
 a flag to disable ipv6 had been added.Inorder to disable ipv6 on ngnix configuration(awx-web container) it is recommended to add following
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1285,6 +1286,24 @@ Then install the new AWX Operator by following the instructions in [Basic Instal
 
 Once the new AWX Operator is up and running, your AWX deployment will also be upgraded.
 
+#### Disable IPV6 for install and upgrade
+Starting with  AWX Operator release 0.24.0 ipv6 had been enabled in ngnix configuration which caused upgrades/new install's to fail.
```
rooftopcellist: @rakesh561 I revised this a bit, what do you think?

```
Starting with  AWX Operator release 0.24.0, [IPv6 was enabled in ngnix configuration](https://github.com/ansible/awx-operator/pull/950) which causes upgrades and installs to fail in environments where IPv6 is not allowed.  Starting in  1.1.1 release, you can set the `ipv6_disabled` flag on the AWX spec.  If you need to use an AWX Operator version between 0.24.0 and 1.1.1 in an IPv6 disabled environment, it is suggested to enable ipv6 on worker nodes.

In order to disable ipv6 on ngnix configuration (awx-web container), add following to the AWX spec.
```

README.md
```diff
@@ -1285,6 +1286,24 @@ Then install the new AWX Operator by following the instructions in [Basic Instal
 
 Once the new AWX Operator is up and running, your AWX deployment will also be upgraded.
 
+#### Disable IPV6 for install and upgrade
```
rooftopcellist: Could we change the name of this section to just `Disable IPv6`?

README.md
rakesh561: @rooftopcellist  Updated the requested changes it makes sense.Only thing is unable to make disable ipv6 section highlighted like other sections.

README.md
rakesh561: @rooftopcellist  Updated the requested changes it makes sense.Only thing is unable to make disable ipv6 section highlighted like other sections.

README.md
```diff
@@ -55,6 +55,7 @@ An [Ansible AWX](https://github.com/ansible/awx) operator for Kubernetes built w
             * [Cluster-scope to Namespace-scope considerations](#cluster-scope-to-namespace-scope-considerations)
             * [Project is now based on v1.x of the operator-sdk project](#project-is-now-based-on-v1x-of-the-operator-sdk-project)
             * [Steps to upgrade](#steps-to-upgrade)
+      * [Disable IPV6](#IPV6)
```
rooftopcellist: To make this link work properly, you'll need:


```suggestion
      * [Disable IPV6](#disable-ipv6)
```

README.md
```diff
@@ -1285,6 +1286,26 @@ Then install the new AWX Operator by following the instructions in [Basic Instal
 
 Once the new AWX Operator is up and running, your AWX deployment will also be upgraded.
 
+### Disable IPV6 ###
+Starting with AWX Operator release 0.24.0,[IPV6 was enabled in ngnix configuration](https://github.com/ansible/awx-operator/pull/950) which causes
+upgrades and installs to fail in environments where IPv6 is not allowed. Starting in 1.1.1 release, tou can set the `ipv6_disabled` flag on the AWX
```
rooftopcellist: There is a small typo here:

```suggestion
upgrades and installs to fail in environments where IPv6 is not allowed. Starting in 1.1.1 release, you can set the `ipv6_disabled` flag on the AWX
```

README.md
rakesh561: Thank you very much @rooftopcellist  can you please check now.

README.md
rakesh561: Thank you very much @rooftopcellist  can you please check now.

rooftopcellist:
@rakesh561 Thank you for taking the time to improve the docs!

PR #1218, opened 2023-02-06
Deploy web and task component in independent deployment

TheRealHaoLiu:
##### SUMMARY

Requires https://github.com/ansible/awx/pull/13423 to function

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

```

```


djyasin (commented):
Did not find any references to `tower_pod` in revised code

thedoubl3j (commented):
for now I think this looks good, I am not seeing anything immediately on my first pass that would be a cause for concern. I will continue going through current code base to make sure we are renaming everything to the new names etc.

thedoubl3j:
the CI/molecule failure is caused by this 
```
"status": {
              "conditions": [
                  {
                      "lastTransitionTime": "2023-03-28T15:03:37Z",
                      "reason": "",
                      "status": "False",
                      "type": "Successful"
                  },
                  {
                      "lastTransitionTime": "2023-03-28T15:13:12Z",
                      "reason": "Failed",
                      "status": "False",
                      "type": "Failure"
                  },
                  {
                      "lastTransitionTime": "2023-03-28T15:13:12Z",
                      "reason": "Running",
                      "status": "True",
```
which we were able to reproduce locally. 

This is caused by us not being able to override the default image and image_version for the molecule deployment. By default, those are set to devel which will fail molecule testing because the new launch script for the rsyslog sidecar is not found in the devel image, but rather the feature image for this new split feature. 

#1308 has been opened to fix this. we were able to fix this locally by just swapping the hardcoded image and image_version.

Makefile
```diff
@@ -67,6 +67,9 @@ CHART_REPO ?= awx-operator
 CHART_BRANCH ?= gh-pages
 CHART_INDEX ?= index.yaml
 
+# Development
```
TheRealHaoLiu: this shouldnt be in this PR lets make the PR against devel

Makefile
```diff
@@ -87,6 +90,11 @@ all: docker-build
 help: ## Display this help.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
+##@ Development
```
TheRealHaoLiu: this should be in a separate PR

roles/installer/defaults/main.yml
```diff
@@ -72,6 +72,22 @@ hostname: ''
 #   kubernetes.io/os: linux
 node_selector: ''
 
+# Add a nodeSelector for the AWX pods. It must match a node's labels for the pod
+# to be scheduled on that node. Specify as literal block. E.g.:
+# node_selector: |
+#   disktype: ssd
```
TheRealHaoLiu:
```suggestion
```

roles/installer/defaults/main.yml
```diff
@@ -72,6 +72,22 @@ hostname: ''
...
+#   kubernetes.io/arch: amd64
+#   kubernetes.io/os: linux
+web_node_selector: ''
+
+# Add a nodeSelector for the AWX pods. It must match a node's labels for the pod
+# to be scheduled on that node. Specify as literal block. E.g.:
+# node_selector: |
+#   disktype: ssd
```
TheRealHaoLiu:
```suggestion
```

roles/installer/defaults/main.yml
```diff
@@ -83,7 +99,27 @@ node_selector: ''
...
-affinity: {}
+# Add a topologySpreadConstraints for the task pods.
+# Specify as literal block. E.g.:
+# task_topology_spread_constraints: |
+#   - maxSkew: 100
+#     topologyKey: "topology.kubernetes.io/zone"
+#     whenUnsatisfiable: "ScheduleAnyway"
+#     labelSelector:
+#       matchLabels:
+#         app.kubernetes.io/name: "<resourcename>""
```
TheRealHaoLiu:
```suggestion
#         app.kubernetes.io/name: "<resourcename>"
```

roles/installer/templates/deployments/task.yaml.j2
```diff
@@ -340,6 +238,8 @@ spec:
                 - /var/lib/pre-stop/scripts/termination-master
 {% endif %}
           env:
+            - name: AWX_COMPONENT
```
TheRealHaoLiu, line 241: this is added so that awx get_task_qeuename() function can identify when in web container and will try to select a task instance's queue

roles/installer/templates/networking/service.yaml.j2
```diff
@@ -13,7 +13,7 @@ metadata:
 {% endif %}
 spec:
   ports:
-  
+
```
TheRealHaoLiu:
```suggestion
```

roles/installer/defaults/main.yml
djyasin: Resolved in https://github.com/ansible/awx-operator/pull/1311

roles/installer/templates/networking/service.yaml.j2
djyasin: Resolve in https://github.com/ansible/awx-operator/pull/1311

Makefile
djyasin: Just to confirm, do we want to cut out these commits and put them in a separate PR?

CFSNM (approved):

PR #1193, opened 2023-01-18
AWX: Add `termination_grace_period_seconds`

stanislav-zaprudskiy:
##### SUMMARY

During deployment restarts or new rollouts, when old ReplicaSet Pods are being terminated, the corresponding jobs which are managed (executed or controlled) by old AWX Pods may end up in `Error` state as there is no mechanism to transfer them to the newly spawned AWX Pods. To work around the problem one could set `termination_grace_period_seconds` in AWX spec proposed by this PR, which does the following:

* It sets the corresponding [`terminationGracePeriodSeconds`](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination) Pod spec of the AWX Deployment to the value provided

  > The grace period is the duration in seconds after the processes running in the pod are sent a termination signal and the time when the processes are forcibly halted with a kill signal

* It adds a [`PreStop`](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#hook-handler-execution) hook script, which will keep AWX Pods in terminating state until it finished, up to `terminationGracePeriodSeconds`.

  > This grace period applies to the total time it takes for both the PreStop hook to execute and for the Container to stop normally

  While the hook script just waits until the corresponding AWX Pod (instance) no longer has any managed jobs, in which case it finishes with success and hands over the overall Pod termination process to normal AWX processes.

related [#11526](https://github.com/ansible/awx/issues/11526)
related [#12326](https://github.com/ansible/awx/issues/12326)

requires https://github.com/ansible/awx/pull/13445

rebased onto https://github.com/ansible/awx-operator/pull/1222

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

There are two `PreStop` hook scripts utilized by AWX Pod containers, all of which communicate with each other through files in a shared volume:
* `termination-master` - runs a so called "handler" command (`awx-manage disable_instance --wait --retry=inf`) until it finishes, or until it receives a "signal" (through files) from other instances of hook scripts. In case of successful handler termination it creates a "termination marker", signalling other instances of hook scripts that they could finish and thus proceed to the remaining container processes termination.
* `termination-waiter` - continously runs in a loop and checks for presence of "signals" from other hook scripts. Apart from checking for "termination marker", it also checks if the corresponding "heartbeat" file was recently updated (by `master`) and fails otherwise.

Out of 4 current containers only `awx-task` has `termination-master` as its hook, while others are `waiter`s (including `redis`).

[...]

PR #1222, opened 2023-02-07
Refine deployment rollouts

stanislav-zaprudskiy:
##### SUMMARY

This is a groundwork which originated in and got split out from #1193 as being beneficial independently from that PR. It addresses some unintended behavior and edge cases during `Deployment` rollouts, making it more robust and reliable. It's better to be reviewed on a per commit basis, with details and rationale available in the corresponding commit messages.

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/roles/installer/tasks/resources_configuration.yml
+++ b/roles/installer/tasks/resources_configuration.yml
@@ -40,13 +40,13 @@
     kind: Secret
     namespace: '{{ ansible_operator_meta.namespace }}'
     name: '{{ ansible_operator_meta.name }}-receptor-ca'
-  register: _receptor_ca
+  register: receptor_ca
   no_log: "{{ no_log }}"
 
 - name: Migrate Receptor CA Secret
   when:
-    - _receptor_ca['resources'] | default([]) | length
-    - _receptor_ca['resources'][0]['type'] != "kubernetes.io/tls"
+    - receptor_ca['resources'] | default([]) | length
+    - receptor_ca['resources'][0]['type'] != "kubernetes.io/tls"
   block:
     - name: Delete old Receptor CA Secret
       k8s:
@@ -61,7 +61,7 @@
       register: _receptor_ca_key_file
     - name: Copy Receptor CA key from old secret to tempfile
       copy:
-        content: "{{ _receptor_ca['resources'][0]['data']['receptor-ca.key'] | b64decode }}"
+        content: "{{ receptor_ca['resources'][0]['data']['receptor-ca.key'] | b64decode }}"
         dest: "{{ _receptor_ca_key_file.path }}"
       no_log: "{{ no_log }}"
     - name: Create tempfile for receptor-ca.crt
@@ -71,13 +71,24 @@
       register: _receptor_ca_crt_file
     - name: Copy Receptor CA cert from old secret to tempfile
       copy:
-        content: "{{ _receptor_ca['resources'][0]['data']['receptor-ca.crt'] | b64decode }}"
+        content: "{{ receptor_ca['resources'][0]['data']['receptor-ca.crt'] | b64decode }}"
         dest: "{{ _receptor_ca_crt_file.path }}"
       no_log: "{{ no_log }}"
     - name: Create New Receptor CA secret
       k8s:
         apply: true
         definition: "{{ lookup('template', 'secrets/receptor_ca_secret.yaml.j2') }}"
+      no_log: "{{ no_log }}"
+    - name: Read New Receptor CA Secret
+      k8s_info:
+        kind: Secret
+        namespace: '{{ ansible_operator_meta.namespace }}'
+        name: '{{ ansible_operator_meta.name }}-receptor-ca'
+      register: _receptor_ca
+      no_log: "{{ no_log }}"
+    - name: Set receptor_ca variable
+      set_fact:
+        receptor_ca: '{{ _receptor_ca }}'
       no_log: "{{ no_log }}"
     - name: Remove tempfiles
       file:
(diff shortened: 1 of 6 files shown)


fosterseth:
@stanislav-zaprudskiy thanks for splitting out this work from the other PR. We'll get some eyes on this PR shortly

roles/installer/tasks/resources_configuration.yml
```diff
@@ -13,9 +13,17 @@
       - status.phase=Running
   register: tower_pod
 
+- name: Set the resource pod as a variable.
```
rooftopcellist, line 16: This will prevent us from grabbing a pod that is in terminating state.  It also ensures that only one pod is grabbed (the oldest).

roles/installer/tasks/resources_configuration.yml
```diff
@@ -210,21 +250,10 @@
     apply: yes
     definition: "{{ lookup('template', 'deployments/deployment.yaml.j2') }}"
     wait: yes
+    wait_timeout: "{{ 120 * replicas or 120 }}"
```
rooftopcellist, line 253: @stanislav-zaprudskiy could you expand on why you added this?

roles/installer/tasks/secret_key_configuration.yml
```diff
@@ -40,10 +40,10 @@
 
 - name: Set secret key secret
   set_fact:
-    __secret_key_secret: '{{ _generated_secret_key["resources"] | default([]) | length | ternary(_generated_secret_key, _secret_key_secret)  }}'
+    secret_key: '{{ _generated_secret_key["resources"] | default([]) | length | ternary(_generated_secret_key, _secret_key_secret)  }}'
```
rooftopcellist, line 43: I want to do some testing around this before merging. cc @TheRealHaoLiu

roles/installer/tasks/resources_configuration.yml
stanislav-zaprudskiy: Adding the corresponding commit message just in case it got missed

> commit e589ceb6610b139c6cd85ca4d0f569e4a7487b07
> 
>     When applying Deployment wait up to (timeout * replicas)
> 
>     There are cases when having a new Deployment may be taking above the
>     default timeout of 120s.
>     For instance, when a Deployment has multiple replicas, and each replica
>     starts on a separate node, and the Deployment specifies new images, then
>     just pulling these new images for each replica may be taking above the
>     default timeout of 120s.
> 
>     Having the default time multiplied by the number of replicas should
>     provide generally enough time for all replicas to start

The corresponding parameter `wait: yes` ([already provided](https://github.com/ansible/awx-operator/pull/1222/files#diff-d8ddd3baf20d7042360b2b1c2c284a8545d24149814ce705535e71ec203a028cR252)) causes the task to wait until Deployment pods are [ready](https://github.com/ansible-collections/kubernetes.core/blob/e6ac87409830b6c698b05dba875cca87d45ea761/plugins/module_utils/k8s/waiter.py#L37), and with the default `120` seconds it could fail depending on the overall k8s and AWX configurations, causing the operator run to fail and be started again. Starting a new run won't solve the problem however, as it would run into the same timeout issue again and again until the new pods are ready - so by increasing maximum waiting time originally it saves from further failures and restarts.

Just to add, also the Deployment's [rollout strategy](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#rollingupdatedeployment-v1-apps) configuration could increase the time for having the new pods ready.

There could be cases when scheduling new pods may not be possible (e.g. due to lack of resources, or wrong image configuration, etc) - but the operator would stuck (unnecessary) waiting. Having lower timeout value would make users aware of the problem earlier - but in multi-replica AWX configurations running in multi-node clusters where image caches aren't generally available, the lower timeout values provide too much false-positive failures of operator runs.

roles/installer/tasks/resources_configuration.yml
stanislav-zaprudskiy: Just adding the corresponding commit message in case it got missed

> commit b3a74362aff871f4eb9b9ba73ea2b4ddf1ba46ff
> 
>     Make AWX Pod variable to be calculated respecting `creationTimestamp` and `deletionTimestamp`
> 
>     Do not consider Pods marked for deletion when calculating tower_pod to
>     address replicas scale down case - where normally Pods spawned recently
>     are being taken for removal. As well as the case when operator kicked
>     off but some old replicas are still terminating.
> 
>     Respect `creationTimestamp` so to make sure that the newest Pod is taken
>     after Deployment application, in which case multiple RS Pods (from old
>     RS and new RS) could be running simultaneously while the rollout is
>     happening.

I've also encountered a couple of cases when `Terminating` pod was picked up, and running `awx-manage` command on it during the later tasks was not possible (the pod was gone already, etc).

Here indeed the oldest pod is taken, while [later](https://github.com/ansible/awx-operator/pull/1222/files#diff-d8ddd3baf20d7042360b2b1c2c284a8545d24149814ce705535e71ec203a028cR275) in the code the newest (the one most recently created) is used.

roles/installer/templates/deployments/deployment.yaml.j2
```diff
@@ -20,8 +20,25 @@ spec:
...
         {{ lookup("template", "../common/templates/labels/common.yaml.j2")  | indent(width=8) | trim }}
         {{ lookup("template", "../common/templates/labels/version.yaml.j2") | indent(width=8) | trim }}
-{% if annotations %}
       annotations:
+{% for template in [
+    "configmaps/config",
+    "secrets/app_credentials",
+    "storage/persistent",
+  ] %}
+        checksum-{{ template | replace('/', '-') }}: "{{ lookup('template', template + '.yaml.j2') | md5 }}"
```
rooftopcellist, line 29: :heart: This is a great trick!  This is an elegant solution to the issue of deployments not being cycled when changes are made to the ConfigMap.  @TheRealHaoLiu is impressed too.

rooftopcellist:
@stanislav-zaprudskiy Thank you for this PR, it is exemplary.  :trophy: 
I have learned a lot from your contributions, and hope to learn more!

PR #1223, opened 2023-02-07
Main.yml respect both web and task deployments

djyasin:
##### SUMMARY

Making rols/installer/main.yml respect both web and task deployment

##### ISSUE TYPE

 - New or Enhanced Feature

### DIFF
--- a/roles/installer/tasks/main.yml
+++ b/roles/installer/tasks/main.yml
@@ -1,13 +1,24 @@
 ---
+- name: Check for presence of task deployment
+  k8s_info:
+    api_version: v1
+    kind: Deployment
+    name: "{{ ansible_operator_meta.name }}-task"
+    namespace: "{{ ansible_operator_meta.namespace }}"
+  register: tower_task_deployment
+
 - name: Check for presence of web deployment
   k8s_info:
     api_version: v1
     kind: Deployment
-    name: "{{ ansible_operator_meta.name }}-web "
+    name: "{{ ansible_operator_meta.name }}-web"
     namespace: "{{ ansible_operator_meta.namespace }}"
-  register: tower_deployment
+  register: tower_web_deployment
 
-# Just execute deployment steps when auto_upgrade is true or when no deployment exists
+# if web and task deployments BOTH do not exist run install.yml
+# if EITHER web or task deployment exist and auto_upgrade is set to true run install.yml
 - name: Start installation
   include_tasks: install.yml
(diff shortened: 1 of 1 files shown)


TheRealHaoLiu (commented):
lgtm

roles/installer/tasks/main.yml
```diff
@@ -1,10 +1,13 @@
...
   k8s_info:
     api_version: v1
     kind: Deployment
-    name: "{{ ansible_operator_meta.name }}"
+    name: "{{ item }}"
     namespace: "{{ ansible_operator_meta.namespace }}"
+  loop: 
+     - "{{ ansible_operator_meta.namespace }}-web"
+     - "{{ ansible_operator_meta.namespace }}-task"
   register: tower_deployment
```
TheRealHaoLiu: this doesnt work here u will just be overriding "tower_deployment"

TheRealHaoLiu: try to see what "tower_deployment" is being used for now and how it maps to "web" and "task" deployments

djyasin: Thank you! I was concerned about that. I will go investigate "tower_deployment."

roles/installer/tasks/main.yml
```diff
@@ -1,11 +1,10 @@
 ---
-- name: Check for presence of Deployment
+- name: Check for presence of web deployment
   k8s_info:
     api_version: v1
     kind: Deployment
-    name: "{{ ansible_operator_meta.name }}"
+    name: "{{ ansible_operator_meta.name-web }}"
```
TheRealHaoLiu: variable ansible_operator_meta.name-web is not a exist 

ansible_operator_meta.name exist 

so this line should be `name: "{{ ansible_operator_meta.name }}-web"`

it appends -web to whatever `ansible_operator_meta.name` resolve to

roles/installer/tasks/main.yml
```diff
@@ -1,11 +1,10 @@
...
     namespace: "{{ ansible_operator_meta.namespace }}"
-  register: tower_deployment
```
TheRealHaoLiu, line 8: this register should still exist or else the following code wont have the `tower_deployment` variable

TheRealHaoLiu:
`Start installation task` should only run under the following condition

- if web and task deployment BOTH **does not exist** run install.yml
- if EITHER web or task deployment exist and auto_upgrade is set to true run install.yml

thedoubl3j (approved):

rooftopcellist (approved):

TheRealHaoLiu:
superseded by https://github.com/ansible/awx-operator/pull/1241

PR #1229, opened 2023-02-09
Specify sslrootcert in database options conditionally (fixes #1215)

### LINKED ISSUE #1215: External postgres database connection over SSL always verify the certificate against CA with sslmode: require

### Bug Summary

When PostgreSQL is using self-signed certificate, there is no straight-forward way to allow connectivity from AWX.

Setting the `ssmode` to `require`still validates certificate. According to [documentation](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLMODE) it should not, unless CA certificate is explicitly specified (that turns it to `verify-ca` mode).

Due to the fact that `sslrootcert` is unconditionally set in the [credentials.py](https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/settings/credentials.py.j2#L11) there is no way to tell AWX to avoid certificate validation.
Adjusting the Jinja template to conditionally define the `sslrootcert` should solve the issue.

In our particular case the PostgreSQL is provisioned by Zalando Postgres Operator that in turns don't allow non-SSL connections, so setting `sslmode` to `disable/allow/prefer` does not work.

### AWX Operator version

1.1.4

### AWX version

21.11.0

### Kubernetes platform

openshift

 4.11.0-0.okd-2023-01-14-152430

### Modifications

no

### Steps to reproduce

1. Create [postgres configuration secret](https://github.com/ansible/awx-operator#external-postgresql-service) with connection details to the external database with non-trusted certificate, specifying:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: test-awx-postgres-configuration
type: Opaque
stringData:
  ...
  sslmode: require
  type: unmanaged
```

Create AWX resource with:
```yaml
---
spec:
  postgres_configuration_secret: test-awx-postgres-configuration
```

### Expected results

Connection works, skipping certificate verification, as `ssmode: require` should not verify certificate unless the `sslrootcert` is explicitly provided.

### Actual results

AWX is not able to work with database, complaining about certificate verification.

### Additional information

_No response_

### Operator Logs

_No response_

manfuin:
##### SUMMARY
Conditionally set `sslrootcert` option only when `sslmode` is set to `veriify-ca` or `verify-full` mode.
Setting `sslmode` to `require` or `prefer` will not validate certificate against CA cert.
More information in issue #1215.
 
##### ISSUE TYPE

##### ADDITIONAL INFORMATION

When PostgreSQL is using self-signed certificate, there is no straight-forward way to allow connectivity from AWX.

Currently setting the `ssmode` to `require` still validates certificate. According to [documentation](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLMODE) it should not, unless CA certificate is explicitly specified (that turns it to `verify-ca` mode). This happens because `sslrootcert` is always added to the configuration file.

This small changeset adds `sslrootcert` conditionally.

Before:
```
        'OPTIONS': { 'sslmode': 'require',
                     'sslrootcert': '/etc/pki/tls/certs/ca-bundle.crt',
        },
```

After:
```
        'OPTIONS': { 'sslmode': 'require',
        },
```
```
        'OPTIONS': { 'sslmode': 'verify-ca',
                     'sslrootcert':  '/etc/pki/tls/certs/ca-bundle.crt',
        },

```

### DIFF
--- a/roles/installer/templates/settings/credentials.py.j2
+++ b/roles/installer/templates/settings/credentials.py.j2
@@ -8,7 +8,9 @@ DATABASES = {
         'HOST': '{{ awx_postgres_host }}',
         'PORT': "{{ awx_postgres_port }}",
         'OPTIONS': { 'sslmode': '{{ awx_postgres_sslmode }}',
+                     {%- if awx_postgres_sslmode == 'verify-ca' or awx_postgres_sslmode == 'verify-full' %}
                      'sslrootcert': '{{ ca_trust_bundle }}',
+                     {%- endif %}
         },
     }
 }
--- a/roles/installer/templates/settings/credentials.py.j2
+++ b/roles/installer/templates/settings/credentials.py.j2
@@ -8,9 +8,9 @@ DATABASES = {
         'HOST': '{{ awx_postgres_host }}',
         'PORT': "{{ awx_postgres_port }}",
         'OPTIONS': { 'sslmode': '{{ awx_postgres_sslmode }}',
-                     {%- if awx_postgres_sslmode == 'verify-ca' or awx_postgres_sslmode == 'verify-full' %}
+{% if awx_postgres_sslmode in ['verify-ca', 'verify-full'] %}
                      'sslrootcert': '{{ ca_trust_bundle }}',
-                     {%- endif %}
+{% endif %}
         },
     }
 }


roles/installer/templates/settings/credentials.py.j2
```diff
@@ -8,7 +8,9 @@ DATABASES = {
         'HOST': '{{ awx_postgres_host }}',
         'PORT': "{{ awx_postgres_port }}",
         'OPTIONS': { 'sslmode': '{{ awx_postgres_sslmode }}',
+                     {%- if awx_postgres_sslmode == 'verify-ca' or awx_postgres_sslmode == 'verify-full' %}
                      'sslrootcert': '{{ ca_trust_bundle }}',
+                     {%- endif %}
```
dsavineau:
```suggestion
{% if awx_postgres_sslmode in ['verify-ca', 'verify-full'] %}
                     'sslrootcert': '{{ ca_trust_bundle }}',
{% endif %}
```

dsavineau (commented):
Before:
```python
        'OPTIONS': { 'sslmode': 'verify-ca',
                     'sslrootcert': '/etc/pki/tls/certs/ca-bundle.crt',
        },
```
After:
with `prefer`
```python
        'OPTIONS': { 'sslmode': 'prefer',        },
```
with `verify-ca`
```python
        'OPTIONS': { 'sslmode': 'verify-ca',                     'sslrootcert': '/etc/pki/tls/certs/ca-bundle.crt',        },
```
so the current implementation changes the credentials.py file formatting.
With the suggestion:
with `prefer`
```python
        'OPTIONS': { 'sslmode': 'prefer',
        },
```
with `verify-ca`
```python
        'OPTIONS': { 'sslmode': 'verify-ca',
                     'sslrootcert': '/etc/pki/tls/certs/ca-bundle.crt',
        },
```

dsavineau (approved):

PR #1232, opened 2023-02-10
feat: add HostAliases in web/task containers (fixes #646)

### LINKED ISSUE #646: Unable to use k8s HostAliases
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
With the current version of the operator it is not possible to use HostAliases in k8s for the AWX deployment/pod.

##### ENVIRONMENT
* AWX version: 19.4.0
* Operator version: 0.14.0
* Kubernetes version: 1.20
* AWX install method: openshift

##### STEPS TO REPRODUCE

- Apply new AWX deployment
- Use FQDNs without DNS resolution or use NAT

##### EXPECTED RESULTS
Able to set HostAliases, to use "classic /etc/hosts" entries.
Used for NAT or additional host entires without DNS resolution 

##### ACTUAL RESULTS
Unable to set HostAliases for the pod/deployment. DNS will always be used.
Modification of /etc/hosts in the AWX container is not possible. 
There is no support for HostAliases in the currently CRDs for AWX deployments.

conanxyz:
##### SUMMARY
add HostAliases in web/task containers

fixes #646

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
```
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -656,6 +656,25 @@ $ oc adm policy add-scc-to-user privileged -z awx
 
 Again, this is the most relaxed SCC that is provided by OpenShift, so be sure to familiarize yourself with the security concerns that accompany this action.
 
+#### Containers HostAliases Requirements
+
+Sometimes you might need to use HostAliases in web/task containers.
+
+| Name         | Description           | Default |
+| ------------ | --------------------- | ------- |
+| host_aliases | A list of HostAliases | None    |
+
+Example of customization could be:
+
+```yaml
+---
+spec:
+  ...
+  host_aliases:
+    - ip: <name-of-your-ip>
+      hostnames:
+        - <name-of-your-domain>
+```
 
 #### Containers Resource Requirements
 
--- a/README.md
+++ b/README.md
@@ -659,7 +659,7 @@ Again, this is the most relaxed SCC that is provided by OpenShift, so be sure to
 
 #### Containers HostAliases Requirements
 
-Sometimes you might need to use HostAliases in web/task containers.
+Sometimes you might need to use [HostAliases](https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/) in web/task containers.
 
 | Name         | Description           | Default |
 | ------------ | --------------------- | ------- |
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -578,6 +578,18 @@ spec:
               image_pull_secret:  # deprecated
                 description: (Deprecated) Image pull secret for app and database containers
                 type: string
+              host_aliases:
+                description: HostAliases for app containers
+                type: array
+                items:
+                  type: object
+                  properties:
+                    ip:
+                      type: string
+                    hostnames:
+                      type: array
+                      items:
+                        type: string
               task_resource_requirements:
                 description: Resource requirements for the task container
                 properties:
--- a/config/manifests/bases/awx-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/awx-operator.clusterserviceversion.yaml
@@ -312,7 +312,7 @@ spec:
         path: host_aliases
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
-        - urn:alm:descriptor:com.tectonic.ui:hidden
+        - urn:alm:descriptor:com.tectonic.ui:text
       - displayName: Web Container Resource Requirements
         path: web_resource_requirements
         x-descriptors:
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -656,6 +656,25 @@ $ oc adm policy add-scc-to-user privileged -z awx
 
 Again, this is the most relaxed SCC that is provided by OpenShift, so be sure to familiarize yourself with the security concerns that accompany this action.
 
+#### Containers HostAliases Requirements
+
+Sometimes you might need to use HostAliases in web/task containers.
```
rooftopcellist: Could you add a link to https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/ in the docs here?

config/crd/bases/awx.ansible.com_awxs.yaml
```diff
@@ -578,6 +578,18 @@ spec:
...
+                items:
+                  type: object
+                  properties:
+                    ip:
+                      type: string
+                    hostnames:
+                      type: array
+                      items:
+                        type: string
               task_resource_requirements:
```
rooftopcellist, line 597: Could you also add an entry in awx-operator.clusterserviceversion.yaml for this so that the parameter shows up in the OLM UI form?

Here is an example:
* https://github.com/ansible/awx-operator/blob/devel/config/manifests/bases/awx-operator.clusterserviceversion.yaml#L25-L28

These docs might be useful:
* https://github.com/openshift/console/blob/447e90085090543e670adfe8e9eb935312346dfa/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#20-fielddependency

rooftopcellist:
After those comments are addressed, this should be OK to merge.

README.md
conanxyz: yes, add in [here](https://github.com/ansible/awx-operator/pull/1232/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5)

config/crd/bases/awx.ansible.com_awxs.yaml
conanxyz: I added this [entry](https://github.com/ansible/awx-operator/pull/1232/files#diff-2a4968792da0598da409c1e5bbc604542cdf7b2b52bbbf4179dd018fcf9baaf4), but I'm not sure if it's right.

config/manifests/bases/awx-operator.clusterserviceversion.yaml
```diff
@@ -308,6 +308,11 @@ spec:
         path: image_pull_secret
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:hidden
+      - displayName: HostAliases for app containers
+        path: host_aliases
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:advanced
+        - urn:alm:descriptor:com.tectonic.ui:hidden
```
dsavineau: if it's a `hidden` field type then this won't display in the UI. This needs to be a `text` type
```suggestion
        - urn:alm:descriptor:com.tectonic.ui:text
```

dsavineau (changes requested):
Except the change needed in the CSV file then everything else looks good.

config/manifests/bases/awx-operator.clusterserviceversion.yaml
conanxyz: host_aliases should be an array type，can it be represented by text type here？

dsavineau: let me double check that, I tried with text and this was working fine.
but now I'm thinking about it, it wasn't a text field per se (as described in https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#6-text) but a form representing the object from the crd.
Maybe we don't need it at all.

@rooftopcellist any thought on this ?

host_aliases

dsavineau: yeah looks like you don't need the text type at all.

https://github.com/openshift/console/tree/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors#auto-generated-form

rooftopcellist: yep, you could just remove this line and it will work.

```suggestion
```

rooftopcellist: I just verified that either works.  Probably best to be explicit and set 'text' though.

rooftopcellist:
The commits in this PR have already merged here:
* https://github.com/ansible/awx-operator/pull/1373

So I'll close this now.  Thanks @zonzpoo !

PR #1243, opened 2023-02-14
[web/task split] Replica readme updates

thedoubl3j:
##### SUMMARY

Update the readme with the new constraint settings added with the deployment split. 

Portion of #1182 

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

I am also looking at adding in a section that talks about the deployment split itself to talk briefly about functionality as well maybe talk about why we did this etc. Any ideas around placement or what folks would want to see there would be greatly appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -715,7 +715,12 @@ spec:
 ```
 #### Scaling the Web and Task Pods independently 
 
-You can scale replicas up or down for each deployment by using the `web_replicas` or `task_replicas` respectively. You can scale all pods across both deployments by using `replicas` as well. These new replicas can be constrained in a similar manner to previous single deployments by appending the particular deployment name in front of the constraint used. More about those new constraints can be found below in the [Assigning AWX pods to specific nodes](#assigning-awx-pods-to-specific-nodes) section. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -713,7 +713,9 @@ spec:
   control_plane_priority_class: awx-demo-high-priority
   postgres_priority_class: awx-demo-medium-priority
 ```
+#### Scaling the Web and Task Pods independently 
 
+You can scale replicas up or down for each deployment by using the `web_replicas` or `task_replicas` respectively. You can scale all pods across both deployments by using `replicas` as well. These new replicas can be constrained in a similar manner to previous single deployments by appending the particular deployment name in front of the constraint used. More about those new constraints can be found below in the [Assigning AWX pods to specific nodes](#assigning-awx-pods-to-specific-nodes) section.
```
TheRealHaoLiu: Might want to refine this a bit more. After reading this, It's not immediately apparent to how replica task_replica and web replica work

djyasin (approved):
I think this makes sense! I am able to follow your logic and I think it's easy to read now.

PR #1260, opened 2023-03-02
fix: use sha1 instead of md5 to make the installer work in FIPS

ravitejb:
##### SUMMARY

New or Enhanced Feature

##### ISSUE TYPE
Use `SHA1` instead of `md5` to make the installer work in FIPS mode.

##### ADDITIONAL INFORMATION

Related to a new annotation added here:
* https://github.com/ansible/awx-operator/pull/1222/commits/94d68bf382ec8dc4ce28d7d8d154663afd00b7fe

### DIFF
--- a/roles/installer/templates/deployments/deployment.yaml.j2
+++ b/roles/installer/templates/deployments/deployment.yaml.j2
@@ -27,7 +27,7 @@ spec:
     "secrets/app_credentials",
     "storage/persistent",
   ] %}
-        checksum-{{ template | replace('/', '-') }}: "{{ lookup('template', template + '.yaml.j2') | md5 }}"
+        checksum-{{ template | replace('/', '-') }}: "{{ lookup('template', template + '.yaml.j2') | sha1 }}"
 {% endfor %}
 {% for secret in [
     "bundle_cacert",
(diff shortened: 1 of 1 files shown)


ravitejb:
@rooftopcellist @TheRealHaoLiu This is ready for review

TheRealHaoLiu:
Ur wonderful!

TheRealHaoLiu (approved):

ravitejb:
Thanks for the review @TheRealHaoLiu 
Any idea who can help us merge this PR?

rooftopcellist:
Thanks for following up and contributing this @ravitejb !

rooftopcellist (approved):

SiliconAlley:
Thanks you guys. We where dealing since weeks to get the awx-operator working on our FIPS enabled environment. After dealing with FQCN's already this finally let us successfully install AWX.

Also thanks for contributing this @ravitejb

walterrowe:
@ravitejb @rooftopcellist 

It might be wise to use something more future proof than SHA1.

https://csrc.nist.gov/Projects/hash-functions

SP 800-131 says SHA1 is acceptable for checksums. SHA2 and SHA3 are preferred. See page 18.

https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-131Ar2.pdf

SiliconAlley:
@walterrowe 

Agree, I thought the same in the morning. We adopted it and it made us happy. Can we easily change it to a prefered one?

rooftopcellist:
I thought about adding `| checksum` here instead, but it looks like that is also SHA1.
* https://github.com/ansible/ansible/blob/bad8843124a50493141a3e3d7920353239021389/lib/ansible/plugins/filter/checksum.yml#L6

It looks like SHA2/3 are not available with the version of hashlib we currently have.  SHA256 is available, but seems like overkill, plus it is `65` characters, so I think that we will have problems with the k8s-enforced annotation character length of 63 chars (maybe that's only on names, not annotations?).

So I am inclined to leave it as is, or if anything, change it to `| checksum` so that we pick up any new hash algorithms support added to ansible-core.

However if this is a no-go for install, we could look at a newer python hashlib dep and see if that makes SHA2 availalble.  But it seems like FIPS still allows SHA1's use for checksums.

Issue #1247, opened 2023-02-17
postgres-configuration support for `sslRejectUnauthorized: false`


jonathon2nd:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that AWX Operator is open source software provided for free and that I might not receive a timely response.

### Feature Summary

When attempting to use [external postgres](https://github.com/ansible/awx-operator#external-postgresql-service), with a postgres cluster created by an [operator](https://github.com/zalando/postgres-operator), an issue arises. 

SSL is assumed and enforced. But there is no way for AWX to not reject unauthorized certs. Setting `sslmode: require` is good, but the other part needed is something like `sslRejectUnauthorized: false`.

It could default to true. But for a postgres cluster with internal access only, being able to just accept auto generated certs would be nice. If I missed something where I can enable this I am sorry, but I have checked [here](https://github.com/ansible/awx-operator#external-postgresql-service) and [here](https://github.com/ansible/awx-operator/blob/devel/docs/migration.md). Also doing a search yields nothing. 

`psycopg2.OperationalError: SSL error: certificate verify failed`

Part of operator log:
```

--------------------------- Ansible Task StdOut -------------------------------
TASK [installer : Check if there are any super users defined.] *****************
task path: /opt/ansible/roles/installer/tasks/initialize_django.yml:2
-------------------------------------------------------------------------------
{"level":"error","ts":1676661508.4393158,"logger":"logging_event_handler","msg":"","name":"awx","namespace":"awx","gvk":"awx.ansible.com/v1beta1, Kind=AWX","event_type":"runner_on_failed","job":"2126785814620255586","EventData.Task":"Check if there are any super users defined.","EventData.TaskArgs":"","EventData.FailedTaskPath":"/opt/ansible/roles/installer/tasks/initialize_django.yml:2","error":"[playbook task failed]","stacktrace":"github.com/operator-framework/operator-sdk/internal/ansible/events.loggingEventHandler.Handle\n\t/workspace/internal/ansible/events/log_events.go:111"}
--------------------------- Ansible Task StdOut -------------------------------
TASK [Check if there are any super users defined.] ********************************
[...]

PR #1277, opened 2023-03-08
PR to support postgres-configuration support for sslRejectUnauthorized: false #1247

rakesh561:
PR to support postgres-configuration support for sslRejectUnauthorized: false #1247

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -533,6 +533,8 @@ type: Opaque
 
 > It is possible to set a specific username, password, port, or database, but still have the database managed by the operator. In this case, when creating the postgres-configuration secret, the `type: managed` field should be added.
 
+**Note**. The vairable `sslrejectUnauthorized1 is valid for `external` databases only. The allowed values are: `false`, `true`.
+
 **Note**: The variable `sslmode` is valid for `external` databases only. The allowed values are: `prefer`, `disable`, `allow`, `require`, `verify-ca`, `verify-full`.
 
 Once the secret is created, you can specify it on your spec:
--- a/roles/installer/tasks/database_configuration.yml
+++ b/roles/installer/tasks/database_configuration.yml
@@ -85,7 +85,6 @@
 
 - name: Set PostgreSQL Configuration
   set_fact:
-  set_fact:
     __postgres_configuration_secret: "{{ pg_config['resources'][0]['metadata']['name'] }}"
 
 - name: Store Database Configuration
--- a/roles/installer/tasks/database_configuration.yml
+++ b/roles/installer/tasks/database_configuration.yml
@@ -85,14 +85,14 @@
 
 - name: Set PostgreSQL Configuration
   set_fact:
-    pg_config: '{{ _generated_pg_config_resources["resources"] | default([]) | length | ternary(_generated_pg_config_resources, _pg_config) }}'
-  no_log: "{{ no_log }}"
-
-- name: Set actual postgres configuration secret used
   set_fact:
     __postgres_configuration_secret: "{{ pg_config['resources'][0]['metadata']['name'] }}"
 
 - name: Store Database Configuration
+    pg_config: '{{ _generated_pg_config_resources["resources"] | default([]) | length | ternary(_generated_pg_config_resources, _pg_config) }}'
+  no_log: "{{ no_log }}"
+
+- name: Set actual postgres configuration secret used
   set_fact:
     awx_postgres_user: "{{ pg_config['resources'][0]['data']['username'] | b64decode }}"
     awx_postgres_pass: "{{ pg_config['resources'][0]['data']['password'] | b64decode }}"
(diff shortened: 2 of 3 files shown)


roles/installer/tasks/database_configuration.yml
```diff
@@ -85,21 +85,22 @@
 
 - name: Set PostgreSQL Configuration
   set_fact:
```
TheRealHaoLiu, line 87: now there's 2 set_fact

roles/installer/tasks/database_configuration.yml
```diff
@@ -85,21 +85,22 @@
...
-    pg_config: '{{ _generated_pg_config_resources["resources"] | default([]) | length | ternary(_generated_pg_config_resources, _pg_config) }}'
```
TheRealHaoLiu, line 88: is this removal intended?

roles/installer/tasks/database_configuration.yml
```diff
@@ -85,21 +85,22 @@
...
-  no_log: "{{ no_log }}"
```
rooftopcellist, line 89: I think this removal is unintended. It results in invalid syntax:
* https://github.com/ansible/awx-operator/blob/6dbe2aa134d1a4ebd03a70bc584921eff54914d6/roles/installer/tasks/database_configuration.yml#L86-L89

roles/installer/tasks/database_configuration.yml
rakesh561: That was not intended.I updated the PR set_fact got added twice in the same task.

roles/installer/tasks/database_configuration.yml
rakesh561: That was not intended.I updated the PR set_fact got added twice in the same task.

roles/installer/tasks/database_configuration.yml
rakesh561: removed the extra set_Fact that was causing issue

rakesh561:
Any suggestions/guidance in case of issues to fix them would be much appreciated would like to contribute to the open source.

roles/installer/tasks/database_configuration.yml
```diff
@@ -85,21 +85,21 @@
...
-
-- name: Store Database Configuration
   set_fact:
     awx_postgres_user: "{{ pg_config['resources'][0]['data']['username'] | b64decode }}"
     awx_postgres_pass: "{{ pg_config['resources'][0]['data']['password'] | b64decode }}"
     awx_postgres_database: "{{ pg_config['resources'][0]['data']['database'] | b64decode }}"
     awx_postgres_port: "{{ pg_config['resources'][0]['data']['port'] | b64decode }}"
     awx_postgres_host: "{{ pg_config['resources'][0]['data']['host'] | b64decode }}"
     awx_postgres_sslmode: "{{ pg_config['resources'][0]['data']['sslmode'] |  default('prefer'|b64encode) | b64decode }}"
+    awx_postgres_sslunauthorized: "{{ pg_config['resources'][0]['data']['sslrejectUnauthorized'] |  default('false'|b64encode) | b64decode }}"
```
fosterseth, line 102: do you mind adding a note about this setting to the operator docs, somewhere here

https://github.com/ansible/awx-operator/blob/6513e93e7fdf117a1518392a0d838f267a67088d/README.md?plain=1#L536

you don't need to change the example spec above that line though, just a note about this param

rakesh561: Done @fosterseth also default value i have set it to false do you think it has to be set to true and then if some one uses external database without sslrejectUnauthorized they can pass the variable false.

README.md
```diff
@@ -533,6 +533,8 @@ type: Opaque
 
 > It is possible to set a specific username, password, port, or database, but still have the database managed by the operator. In this case, when creating the postgres-configuration secret, the `type: managed` field should be added.
 
+**Note**. The vairable `sslrejectUnauthorized1 is valid for `external` databases only. The allowed values are: `false`, `true`.
```
dsavineau, line 536:
```suggestion
**Note**. The variable `sslrejectUnauthorized` is valid for `external` databases only. The allowed values are: `false`, `true`.
```

dsavineau (commented):
Could you share where's the `sslrejectUnauthorized` option is coming from ?

I've looked at the libpq and psycopg documentation/code and never seen such option.
The closest I found was in a postgresql nodejs binding project.

Also if this option is only needed for external database then we need to set this option only if the database type is `unmanaged`

rakesh561:
@dsavineau seems like sslrejectUnauthorized` is not available in the python postgres library i will close this request

Issue #1293, opened 2023-03-19
Unnecessary secrets are created by the Restore role in some cases


rooftopcellist:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that the AWX Operator is open source software provided for free and that I might not receive a timely response.

### Bug Summary

During a restore, some of the generated secrets are duplicated, but with empty data/values.  

### Impact

* This does not affect functionality, the resulting deployment works, but there are unneeded secrets created which might confuse users.  
* This bug would not be seen by users who specified custom on the original deployment.

Found while testing this PR:
* https://github.com/ansible/awx-operator/pull/1212

### AWX Operator version

1.3.0

### AWX version

21.13.0

### Kubernetes platform

Openshift

### Kubernetes/Platform version

k8s: 1.24.0
OCP: 4.11

### Modifications

no

### Steps to reproduce

* https://gist.github.com/rooftopcellist/dbb1183a22d52b24ce35147312664d8f

### Expected results

In regards to secrets, the way a restore _should_ work is:
1. Backup is created and the existing secrets used by the AWX deployment are stored in the backup PVC. 
2. Restore is started and the secrets are created from the contents of the backup, the secrets have the same names as the original deployment, even if the secrets were generated on the original deployment.
3. A new AWX object (awx-2) is created by the restore role with all the secrets specified on it (secret names from the first deployment are used).
4. The ownerReferences on those secrets are set to null so that they aren't cleaned up unintentionally if awx-2 is deleted.

### Actual results

In this case, what I am observing is that new secrets (prefixed by `awx-2-<secret_name>`) are also being created when [this task](https://github.com/ansible/awx-operator/blob/9c788ba3bde8229ccd8f4ad5a8bda1197439ae72/roles/restore/tasks/cleanup.yml#L11) is run.  It turns out we have a duplicate task for cleaning up ownerRef's on secrets in the restore role that is not working as expected.
* This the the first cleanup task that is run: [roles/installer/secrets.yml:ln97](https://github.com/ansible/awx-operator/blob/9c788ba3bde8229ccd8f4ad5a8bda1197439ae72/roles/restore/tasks/secrets.yml#L97) - we should keep this one
[...]

PR #1294, opened 2023-03-19
Stop creating empty k8s secrets during restore role cleanup

rooftopcellist:
##### SUMMARY

Resolves:
* https://github.com/ansible/awx-operator/issues/1293

During a restore, some of the generated secrets are duplicated, but with empty data/values.  This is because the task the remove ownerReferences is being run with the wrong secret names

We can also remove the defaults as they are no longer needed.  

##### ISSUE TYPE

### DIFF
--- a/roles/restore/tasks/cleanup.yml
+++ b/roles/restore/tasks/cleanup.yml
@@ -8,22 +8,6 @@
     state: absent
     force: true
 
-- name: Remove ownerReferences from secrets to avoid garbage collection
-  k8s:
-    definition:
-      apiVersion: v1
-      kind: Secret
-      metadata:
-        name: '{{ item }}'
-        namespace: '{{ ansible_operator_meta.namespace }}'
-        ownerReferences: null
-  loop:
-    - '{{ secret_key_secret }}'
-    - '{{ admin_password_secret }}'
-    - '{{ broadcast_websocket_secret }}'
-    - '{{ postgres_configuration_secret }}'
-  no_log: "{{ no_log }}"
-
 - name: Cleanup temp spec file
   file:
     path: "{{ tmp_spec.path }}"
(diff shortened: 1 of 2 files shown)


roles/restore/vars/main.yml
```diff
@@ -7,9 +7,4 @@ _postgres_image_version: 13
 backup_api_version: '{{ deployment_type }}.ansible.com/v1beta1'
 backup_kind: 'AWXBackup'
 
-# set default secret names to be used if a backup dir and claim are provided (not a backup_name)
-secret_key_secret: '{{ deployment_name }}-secret-key'
-admin_password_secret: '{{ deployment_name }}-admin-password'
-broadcast_websocket_secret: '{{ deployment_name }}-broadcast-websocket'
-postgres_configuration_secret: '{{ deployment_name }}-postgres-configuration'
```
dsavineau, line 14: If we remove the postgres_configuration_secret from the default variables then we would have an undefined variable here https://github.com/ansible/awx-operator/blob/devel/roles/restore/tasks/postgres.yml#L5
The others variables (admin, password and websocket) were only used in the cleanup task.

roles/restore/tasks/cleanup.yml
```diff
@@ -8,22 +8,6 @@
     state: absent
     force: true
 
-- name: Remove ownerReferences from secrets to avoid garbage collection
```
lucas-benedito, line 11: Removing the task fixes the empty-secret creation, but also removes ownerReferences cleanup from the restore flow entirely. This is safe only if #1353 lands first to handle it during reconciliation. Otherwise, consider fixing the variable names instead of removing the task — `installer/tasks/cleanup.yml` already has the correct resolution pattern.

PR #1295, opened 2023-03-20
[web/task split] fix scale down bug 

### LINKED ISSUE #1255: [web/task split] Fix scale_down_deployment

### Bug Summary

With the new deployments being split, we will need to address how we scale down the deployments for upgrades etc. 

In particular, the roles/installer/tasks/scale_down_deployment.yml will need to be updated to scale down both the web and tasks pods as it currently references a deployment that doesn't exist anymore. specifically `name: "{{ ansible_operator_meta.name }}"` will need to have web and task appended so it scales both of them down

### AWX Operator version

devel

### AWX version

n/a

### Kubernetes platform

kubernetes

n/a

### Modifications

no

### Steps to reproduce

Deploy operator on the feature_web-task-split branch and attempt an upgrade.

### Expected results

deployments scale down appropriately

### Actual results

Currently, the web and task deployments do not scale down as the reference is currently pointing to the old the deployment name. 

### Additional information

We will need to scale down both deployments for this.

### Operator Logs

_No response_

thedoubl3j:
##### SUMMARY

fixes #1255 

While working on cleanup, the scale down playbook didn't have the new naming conventions for the pods and would currently scale down nothing during any type of db upgrade etc. 

This adds a simple list loop to scale down both pods during any db upgrade/migration/configuration in the operator. 

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

**test case**
To test this properly, we will need to either provoke the rescue block or do a db migration or upgrade. I think there is also a way to provoke it if there is no DB specified but I need to investigate this. 

If anyone has any pointers or thoughts on how we can approach this, please share as I am running low on ideas.

### DIFF
--- a/roles/installer/tasks/scale_down_deployment.yml
+++ b/roles/installer/tasks/scale_down_deployment.yml
@@ -12,8 +12,11 @@
   kubernetes.core.k8s_scale:
     api_version: v1
     kind: Deployment
-    name: "{{ ansible_operator_meta.name }}"
+    name: "{{ item }}"
     namespace: "{{ ansible_operator_meta.namespace }}"
     replicas: 0
     wait: yes
+  loop:
+    - "{{ ansible_operator_meta.name }}-task"
+    - "{{ ansible_operator_meta.name }}-web"
   when: this_deployment['resources'] | length
--- a/roles/restore/tasks/postgres.yml
+++ b/roles/restore/tasks/postgres.yml
@@ -50,7 +50,7 @@
   k8s_info:
     api_version: v1
     kind: Deployment
-    name: "{{ deployment_name }}"
+    name: "{{ ansible_operator_meta.namespace }}-task"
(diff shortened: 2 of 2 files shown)


djyasin (approved):

TheRealHaoLiu (approved):

rooftopcellist:
@thedoubl3j It looks like there is an entry in `roles/restore` that also needs to be addressed:
* https://github.com/ansible/awx-operator/blob/9c788ba3bde8229ccd8f4ad5a8bda1197439ae72/roles/restore/tasks/postgres.yml#L49-L65

thedoubl3j:
@rooftopcellist take a look when you get a sec, I elected to check for a task deployment rather than just awx (doesn't exist anymore) or both since if one is there, the other will be as well.

gamuniz (approved):

PR #1320, opened 2023-04-03
Add websocket configuration for contour ingress

lumarel:
##### SUMMARY

As described in [this issue](https://github.com/ansible/awx/issues/9747#issuecomment-1440068816), the websocket is not available through contour, as this needs special configuration.
Here is also the [help page](https://projectcontour.io/docs/1.24/config/websockets/) from contour.

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

So this change basically implements a check if a contour ingress is installed on the cluster, and if yes, adds the websocket route, as well as the seperate path for that.

Unfortunately I wasn't able to spin up the tests, as it looks for be that these are broken somehow, if I can get these working somehow I will gladly run through them!
As well as, if any changes are wanted I will definitely try to improve the contribution.

### DIFF
--- a/roles/installer/templates/networking/ingress.yaml.j2
+++ b/roles/installer/templates/networking/ingress.yaml.j2
@@ -9,9 +9,15 @@ metadata:
   namespace: '{{ ansible_operator_meta.namespace }}'
   labels:
     {{ lookup("template", "../common/templates/labels/common.yaml.j2") | indent(width=4) | trim }}
-{% if ingress_annotations %}
+{% if ingress_annotations or ingress_controller|lower == "contour" %}
   annotations:
+{% if ingress_annotations %}
     {{ ingress_annotations | indent(width=4) }}
+{% endif %}
+{% if ingress_controller|lower == "contour" %}
+    projectcontour.io/websocket-routes: "/websocket"
+    kubernetes.io/ingress.class: contour
+{% endif %}
 {% endif %}
 spec:
 {% if ingress_class_name %}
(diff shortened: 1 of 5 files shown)


TheRealHaoLiu (changes requested):
instead of adding `is_ingress_contour` i think we can add `contour` as a `ingress_type` and use that to trigger the setting

lumarel:
@TheRealHaoLiu I thought about that in the first place, but refrained from that, because I thought it would be better if it autodetects the ingress controller, but yes sure I am going to rebuild the change to an upfront parameter.

One thing I'm just thinking about is in this case if it would be better, to create a dependent `ingress_controller` variable instead of a completely new `ingress_type`, as the latter one will complicate the documentation a lot more.

In case this is also not really the desired outcome, I will of course rewrite it to its seperate `ingress_type` and add the new type to the docs

roles/installer/templates/networking/ingress.yaml.j2
```diff
@@ -27,6 +33,15 @@ spec:
                 name: '{{ ansible_operator_meta.name }}-service'
                 port:
                   number: 80
+{% if ingress_controller|lower == "contour" %}
+          - path: '{{ ingress_path }}/websocket'
```
rooftopcellist, line 37: @lumarel hey, could you give us some context for why it is only the `/websocket` path here?

Could you also describe the use case for configuration like this?  I'm not very familiar with Contour and Envoy, thanks ahead of time!

rooftopcellist: Ok, I re-read your description, I understand now, thanks.

rooftopcellist (approved):

TheRealHaoLiu (approved):

Issue #1318, opened 2023-03-30
Allow multiple hosts to be included in the ingress


goaliejordan:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that AWX Operator is open source software provided for free and that I might not receive a timely response.

### Feature Summary

I would like to be able to use multiple hostnames within the helm chart values so that the controlled ingress spec rules can use the list to make multiple host endpoints. We use different hostname suffixes for specific certs so this would provide the functionality needed for that. Currently the Operator is removing any manually added host rules.

`apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: 
  annotations:
spec:
  rules:`

guillaumelfv:
Hi, i am working on this as we have the same need at my company. I should be able to make a PR in the next few days.

PR #1339, opened 2023-04-06
Allow multiple hosts to be defined when using ingress

### LINKED ISSUE #897: Question: Multiple hostnames
I have a hopefully simple question:

Is it possible to specify multiple hostnames?

My usecase is a machine in a local network that should be accessible via it's machine name and via a domain.
So awx should be accessible by typing `awx.server-name` and by `awx.server-name.domain.tld`.

guillaumelfv:
##### SUMMARY
Allow multiple hosts to be defined when using ingress.
Replace `hostname` and `ingress_tls_secret` with a dictionary to define ingress:
```yaml
ingress_hosts:
  - hostname: awx-demo.example.com
    tls_secret: example-com-tls
  - hostname: awx-sample.example.io
```
`ingress_hosts` default to empty and when defined `hostname` is mandatory but `tls_secret` is optional.

For the example above the ingress resources will be templatize as follow:
```yaml
  rules:
  - host: awx-demo.example.com
    http:
      paths:
      - backend:
          service:
            name: awx-service
            port:
              number: 80
        path: /
        pathType: Prefix
  - host: awx-sample.example.io
    http:
      paths:
      - backend:
          service:
            name: awx-service
            port:
              number: 80
        path: /
        pathType: Prefix
  tls:
  - hosts:
    - awx-demo.example.com
    secretName: example-com-tls
```

fixes #897 #1318 

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
`hostname` is removed from https://github.com/ansible/awx-operator#ingress-type 
```
hostname | Define the FQDN | {{ meta.name }}.example.com
```
and replaced with `ingress_hosts`
```
ingress_hosts | Define one or multiple FQDN with optional Secret that contains the TLS information | Empty string
```
This change is not backward compatible and anyone using ingress `hostname` spec will need to change to the new `ingress_hosts` spec.


guillaumelfv:
Whichever PR is merged first between this one and #1320 will need to rebase and adapt the `ingress.yaml.j2`

rooftopcellist:
We will need to make this backwards compatible before we can merge.  For example, we can't remove `ingress_tls_secret` from the CRD (same with hostname).  And task logic needs to be added to enable `hostname` or `ingress_hosts`.

Also, the old CRD parameters need to be marked as Deprecated in the CRD description.

guillaumelfv:
sorry @rooftopcellist i had to close this PR and open a new one #1377

PR #1343, opened 2023-04-08
Add add container port name to awx-web

mateuszdrab:
##### SUMMARY

In AWX Operator v2, the new awx-web pod does not have the port name which was present in previous releases.
Re-adding it for sake of backwards compatibility with (e.g. custom resources like servicemonitors refering to it, like it is in my case).

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/roles/installer/templates/deployments/web.yaml.j2
+++ b/roles/installer/templates/deployments/web.yaml.j2
@@ -140,6 +140,7 @@ spec:
 {% endif %}
           ports:
             - containerPort: 8052
+              name: http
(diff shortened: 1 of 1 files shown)


TheRealHaoLiu:
https://github.com/ansible/awx-operator/blob/checkpoint_v1/roles/installer/templates/deployments/deployment.yaml.j2#L170-L174

before web/task split we didn't have port name

rooftopcellist:
Agreed, this wasn't there before and it looks like it is unneeded.

mateuszdrab:
I agree, I wasn't thinking right that day.
It turned out my service monitor broke because the labels of the service changed. The port name itself did not.

PR #1349, opened 2023-04-11
fix task deployment tolerations

### LINKED ISSUE #1337: Operator 2.0.0 - Inconsistent templating to tolerations between task and web causes deployment failure

### Bug Summary

Deploying an AWX instance with Tolerations defined (either tolerations or task_tolerations/web_tolerations) causes the operator to render an invalid Deployment Payload.

### AWX Operator version

2.0.0

### AWX version

22.0.0

### Kubernetes platform

kubernetes

1.23.x

### Modifications

no

### Steps to reproduce

Deploy an AWX instance in defining:

task_tolerations: |
    -   effect: <effect>
        key: <key>
        operator: <Operator>

web_tolerations: |
    -   effect: <effect>
        key: <key>
        operator: <Operator>
        
Will cause the web pod to deploy successfully and the task pod to fail.

### Expected results

AWX Deployment to success, and both task and web pods to be created

### Actual results

web pod is created, task pod never creates. Errors can be found in Operator Logs.

### Additional information

This seems to be caused by a difference in the web and task toleration renderings in the templates:

Task Uses to_nice_yaml and indent
https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/deployments/task.yaml.j2#L386
https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/deployments/task.yaml.j2#L389

Web uses only indent (original behaviour from previous versions)
https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/deployments/web.yaml.j2#L270
https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/deployments/web.yaml.j2#L273

I suspect the task definition needs to be made the same as web to resolve this issue, since the input from the operator is a multiline yaml string.

### Operator Logs

failed: [localhost] (item=task) => {"ansible_loop_var": "item", "changed": false, "item": "task", "msg": "Failed to apply object: b'{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Deployment in version \\\\\\\\"v1\\\\\\\\" cannot be handled as a Deployment: json: cannot unmarshal string into Go struct field PodSpec.spec.template.spec.tolerations of type []v1.Toleration","reason":"BadRequest","code":400}\\\
'", "reason": "Bad Request"}

thedoubl3j:
##### SUMMARY

The task deployment would render broken or malformed yaml for the `task_tolerations` and `tolerations` field due to the use of `to_nice_yaml`. 

For giggles, I also tested to make sure affinity (other parameter that uses to_nice_yaml) rendered correctly and it did so it is not being removed there.

After digging deeper, passing `to_nice_yaml` to a yaml object causes this problem so that is why we are removing them. Affinity etc works as intended as they are dict objects.

fixes #1337 

##### ISSUE TYPE

##### ADDITIONAL INFORMATION
**test case**
- deploy operator as is today with tolerations applied to the task container 
```
 task_tolerations: |
   - key: "dedicated"
     operator: "Equal"
     value: "AWXtask"
     effect: "NoSchedule"
```
- observe failed deployment of the task pod, will yield malformed tolerations which cannot be applied.
- now do same as above but build from this branch and apply task tolerations
- result will be correctly rendered tolerations
```
 task_tolerations: 
   - key: "dedicated"
     operator: "Equal"
     value: "AWXtask"
     effect: "NoSchedule"
```

### DIFF
--- a/roles/installer/templates/deployments/task.yaml.j2
+++ b/roles/installer/templates/deployments/task.yaml.j2
@@ -391,10 +391,10 @@ spec:
 {% endif %}
 {% if task_tolerations %}
       tolerations:
-        {{ task_tolerations | to_nice_yaml | indent(width=8) }}
+        {{ task_tolerations | indent(width=8) }}
 {% elif tolerations %}
       tolerations:
-        {{ tolerations | to_nice_yaml | indent(width=8) }}
+        {{ tolerations | indent(width=8) }}
 {% endif %}
 {% if task_affinity %}
       affinity:


thedoubl3j:
cc @MrBones757

djyasin (approved):

TheRealHaoLiu (approved):

PR #1351, opened 2023-04-11
Add support to specify different annotation for the task and web deployments in AWX operator

### LINKED ISSUE #1331: Add support to specify different annotation for the task and web deployments in AWX operator 2.x.x

### Feature Summary

To specify the annotations, we use the `annotations` keyword in the AWX manifest:
```
---
apiVersion: "awx.ansible.com/v1beta1"
kind: "AWX"
metadata:
  name: "awx"
spec:
  annotations: |
    foo: "bar"
```
This works fine in pre-2 releases. In 2.0.0, the AWX operator creates two deployments: task and web. If annotations are the same, everything is fine, but if we want to specify different annotations, we have a problem.

Here is our use case. We use HashiCorp Vault to specify DB (and other) secrets for AWX. The HC Vault configuration is specified using annotations. We are using a self-signed cert inside the cluster, so we need a CA cert in the Vault side container. To do that, we create a ConfigMap that includes the CA cert. We mount the CA cert to the `awx-web` container. Then we use the annotation:
```
vault.hashicorp.com/agent-copy-volume-mounts: "awx-web"
```
This annotation tells Vault to copy volume mounts from the `awx-web` container to the Vault side container, which is then used for to verify the HC Vault cert.

Because in 2.0.0, there is no `awx-web` container in the "task" deployment, the task containers crash.

We suggest having the following options:
* `annotations`: Common annotations for both deployments
* `task_annotations`: Annotations specific to the task deployment
* `web_annotations`: Annotations specific to the web deployment

djyasin:
##### SUMMARY

Fixes #1331

##### ISSUE TYPE

##### Test-Case Scenarios
- 4 Test cases total
    - No annotations
        - Web and task should deploy successfully with no additional annotations
    - Just `web_annotations`
        - Only web deployment has custom annotations, matches what was set
           - Added example from defaults to spec, checked web deployment to ensure they were present, then checked other deployments to be sure they were not there. 
    - Just `task_annotations`
       - Only task deployment has  custom annotations, matches what was set
    - Common `annotations` for both deployments
       - Both deployments have common annotations 
    - Override scenario, have `web_annotations`, `task _annotations`, and `annotations`
       - Ensure these remain distinct

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -186,7 +186,13 @@ spec:
                 description: Additional labels to apply to the service
                 type: string
               annotations:
-                description: annotations for the pods
+                description: Common annotations for both Web and Task deployments.
+                type: string
+              task_annotations:
+                description: Task deployment annotations. This will override the general annotations parameter for the Task deployment.
+                type: string
+              web_annotations:
+                description: Web deployment annotations. This will override the general annotations parameter for the Web deployment.
                 type: string
               tolerations:
                 description: node tolerations for the pods
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -188,6 +188,18 @@ web_affinity: {}
 #   my.annotation/2: value2
 annotations: ''
 
+# Override annotations to awx task pods. Specify as literal block. E.g.:
+# task_annotations: |
+#   my.task-annotation/1: value
+#   my.task-annotation/2: value2
+task_annotations: ''
+
+# Override annotations to awx web pods. Specify as literal block. E.g.:
+# web_annotations: |
+#   my.web-annotation/1: value
+#   my.web-annotation/2: value2
+web_annotations: ''
+
 admin_user: admin
 admin_email: test@example.com
 
--- a/roles/installer/templates/deployments/task.yaml.j2
+++ b/roles/installer/templates/deployments/task.yaml.j2
@@ -43,7 +43,9 @@ spec:
   ] %}
         checksum-secret-{{ secret }}: "{{ lookup('ansible.builtin.vars', secret, default='')["resources"][0]["data"] | default('') | sha1 }}"
 {% endfor %}
-{% if annotations %}
+{% if task_annotations %}
+        {{ task_annotations | indent(width=8) }}
+{% elif annotations %}
         {{ annotations | indent(width=8) }}
 {% endif %}
     spec:
--- a/roles/installer/templates/deployments/web.yaml.j2
+++ b/roles/installer/templates/deployments/web.yaml.j2
@@ -44,7 +44,9 @@ spec:
   ] %}
         checksum-secret-{{ secret }}: "{{ lookup('ansible.builtin.vars', secret, default='')["resources"][0]["data"] | default('') | md5 }}"
 {% endfor %}
-{% if annotations %}
+{% if web_annotations %}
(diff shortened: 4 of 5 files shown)


roles/installer/templates/deployments/task.yaml.j2
```diff
@@ -44,7 +44,9 @@ spec:
         checksum-secret-{{ secret }}: "{{ lookup('ansible.builtin.vars', secret, default='')["resources"][0]["data"] | default('') | sha1 }}"
 {% endfor %}
 {% if annotations %}
-        {{ annotations | indent(width=8) }}
+  annotations: {{ annotations | indent(width=8) }}
+{% elif task_annotations %}
+  task_annotations: {{ task_annotations | indent(width=8) }}
```
TheRealHaoLiu:
```suggestion
  annotations: {{ task_annotations | indent(width=8) }}
```

roles/installer/templates/deployments/web.yaml.j2
```diff
@@ -45,7 +45,9 @@ spec:
         checksum-secret-{{ secret }}: "{{ lookup('ansible.builtin.vars', secret, default='')["resources"][0]["data"] | default('') | md5 }}"
 {% endfor %}
 {% if annotations %}
-        {{ annotations | indent(width=8) }}
+  annotations: {{ annotations | indent(width=8) }}
+{% elif web_annotations %}
+  web_annotations: {{ web_annotations| indent(width=8) }}
```
TheRealHaoLiu:
```suggestion
  annotations: {{ web_annotations | indent(width=8) }}
```

config/crd/bases/awx.ansible.com_awxs.yaml
```diff
@@ -183,7 +183,13 @@ spec:
                 description: Additional labels to apply to the service
                 type: string
               annotations:
-                description: annotations for the pods
+                description: Common annotations for both deployments
+                type: string
+              task_annotations:
+                description: Annotations specific to the task deployment
```
TheRealHaoLiu: is this an override of `annotations` field or is it a combination with the `annotation`

TheRealHaoLiu: from the code this looks like a override but the description here sounds like it will be additive rather than an override

TheRealHaoLiu:
please update x-descriptor for the ClusterServiceVersion resource 

example
https://github.com/ansible/awx-operator/blob/0c2f45569cadd3d41be4375a91927623e60e1b14/config/manifests/bases/awx-operator.clusterserviceversion.yaml#L688-L692

config/crd/bases/awx.ansible.com_awxs.yaml
rooftopcellist: Maybe rewording it to say:
```
                description: Task deployment annotations. This will override the general annotations parameter for the Task deployment.
```

roles/installer/defaults/main.yml
```diff
@@ -184,6 +184,18 @@ web_affinity: {}
 #   my.annotation/2: value2
 annotations: ''
 
+# Add annotations to awx task pods. Specify as literal block. E.g.:
```
TheRealHaoLiu: s/Add/Override/

jjwatt (approved):
Paired with @djyasin @TheRealHaoLiu @gamuniz coding, debugging and testing. LGTM 🤌

gamuniz (approved):

PR #1353, opened 2023-04-11
Unset k8s secret ownerRefs earlier in reconciliation loop

rooftopcellist:
- Remove ownerRefs earlier in the reconciliation loop so that they always get removed, even if an error short-circuits the reconciliation loop before the cleanup playbook is run.

##### SUMMARY

Currently, if the AWX CR is deleted before the reconciliation loop completes, or if it errors before the cleanup task, it is possible to get in a state where the generated k8s secrets are deleted.  

This may sound innocuous at a glance, but when you consider that the postgres PVC _is_ left behind and has already created the `awx` postgres user with the original pg_password valued, you can see how a re-deploy witht he same deployment_name would result in an AWX instance that cannot connect to the database.  

This should make it much harder to get in to that state.  

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

**Work-around**

Note: If you are already in this state, it is possible to get the pg_password value from the generated postgres_configuration_secret, exec into the postgresql pod, and change the `awx` user's password.

### DIFF
--- a/roles/installer/tasks/admin_password_configuration.yml
+++ b/roles/installer/tasks/admin_password_configuration.yml
@@ -40,7 +40,7 @@
 
 - name: Set admin password secret
   set_fact:
-    __admin_password_secret: '{{ _generated_admin_password["resources"] | default([]) | length | ternary(_generated_admin_password, _admin_password_secret) }}'
+    __admin_password_secret: "{{ admin_password_secret | length | ternary(_admin_password_secret, _generated_admin_password) }}"
   no_log: "{{ no_log }}"
 
 - name: Store admin password
@@ -58,4 +58,6 @@
         namespace: "{{ ansible_operator_meta.namespace }}"
         ownerReferences: null
   no_log: "{{ no_log }}"
-  when: not garbage_collect_secrets | bool
+  when:
+    - not garbage_collect_secrets | bool
+    - _generated_admin_password["resources"] | default([]) | length
--- a/roles/installer/tasks/admin_password_configuration.yml
+++ b/roles/installer/tasks/admin_password_configuration.yml
@@ -47,3 +47,15 @@
   set_fact:
     admin_password: "{{ __admin_password_secret['resources'][0]['data']['password'] | b64decode }}"
   no_log: "{{ no_log }}"
+
+- name: Remove ownerReferences
+  k8s:
+    definition:
+      apiVersion: v1
+      kind: Secret
+      metadata:
+        name: "{{ __admin_password_secret }}"
+        namespace: "{{ ansible_operator_meta.namespace }}"
+        ownerReferences: null
+  no_log: "{{ no_log }}"
+  when: not garbage_collect_secrets | bool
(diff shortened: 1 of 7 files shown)


rooftopcellist:
This currently fails because of this error, which makes no sense as this would have been needed for the existing tasks in cleanup.yml..

```
 TASK [Remove ownerReferences] ******************************** 
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to retrieve requested object: b'{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"secrets \\\\\"{\\'changed\\': False, \\'resources\\': [{\\'apiVersion\\': \\'v1\\', \\'data\\': {\\'password\\': \\'M3R1M3FiM1JxSzRXS3dzalRMSDVEWjFybnVFVDBjMmw=\\'}, \\'kind\\': \\'Secret\\', \\'metadata\\': {\\'annotations\\': {\\'kubectl.kubernetes.io\\\\\" is forbidden: User \\\\\"system:serviceaccount:awx:awx-operator-controller-manager\\\\\" cannot get resource \\\\\"secrets/last-applied-configuration\\': \\'{\\\\\\\\\\\\\"apiVersion\\\\\\\\\\\\\":\\\\\\\\\\\\\"v1\\\\\\\\\\\\\",\\\\\\\\\\\\\"kind\\\\\\\\\\\\\":\\\\\\\\\\\\\"Secret\\\\\\\\\\\\\",\\\\\\\\\\\\\"metadata\\\\\\\\\\\\\":{\\\\\\\\\\\\\"labels\\\\\\\\\\\\\":{\\\\\\\\\\\\\"app.kubernetes.io\\\\\" in API group \\\\\"\\\\\" in the namespace \\\\\"awx\\\\\"\",\"reason\":\"Forbidden\",\"details\":{\"name\":\"{\\'changed\\': False, \\'resources\\': [{\\'apiVersion\\': \\'v1\\', \\'data\\': {\\'password\\': \\'M3R1M3FiM1JxSzRXS3dzalRMSDVEWjFybnVFVDBjMmw=\\'}, \\'kind\\': \\'Secret\\', \\'metadata\\': {\\'annotations\\': {\\'kubectl.kubernetes.io\",\"kind\":\"secrets\"},\"code\":403}\\n'", "reason": "Forbidden"}
```

Adding that rule to the awx-operator-controller-manager serviceAccount did not even solve the issue...

roles/installer/tasks/admin_password_configuration.yml
```diff
@@ -40,10 +40,24 @@
...
     admin_password: "{{ __admin_password_secret['resources'][0]['data']['password'] | b64decode }}"
   no_log: "{{ no_log }}"
+
+- name: Remove ownerReferences
+  k8s:
+    definition:
+      apiVersion: v1
+      kind: Secret
+      metadata:
+        name: "{{ __admin_password_secret }}"
```
lucas-benedito, line 57: `__admin_password_secret` is a `k8s_info` result dict, not a string name. Verified on a live cluster — this renders as `{'changed': False, 'resources': [...]}` in the `name:` field. Same issue on `broadcast_websocket_configuration.yml`. Extract the name first, like `secret_key_configuration.yml` does:
```yaml
admin_password_secret_name: "{{ __admin_password_secret['resources'][0]['metadata']['name'] }}"
```

lucas-benedito (changes requested):

PR #1369, opened 2023-04-19
Update README.md

sdx23:
Adaptations to use kustomize implicitly via kubectl apply -k

##### SUMMARY

kubectl has the functionality for kustomizing built in. Depending on the kubernetes distribution used, installing kustomize as a separate binary in addition is a superfluous step that unnecessarily complicates usage.

Note, however, I've not looked into minikube whether the proposed change is relevant for their kubectl as well.

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ $ alias kubectl="minikube kubectl --"
 
 ### Basic Install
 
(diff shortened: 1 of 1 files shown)


shanemcd (approved):
This is awesome, thank you!

PR #1135, opened 2022-11-28
Added flag to disable ipv6 listener

dale-mittleman:
Copy of https://github.com/ansible/awx-operator/pull/1024.  Apologies for the redundancy, but my team is looking to see this change made and I was tired of waiting for the update to the other PR

##### SUMMARY

Introducing flag to disable nginx IPv6 listener, for environments that are not IPv6-enabled

##### ISSUE TYPE

 - Bug, Docs Fix or other nominal change

##### ADDITIONAL INFORMATION

Flag named ipv6_disabled, defaults to false

PR #1376, opened 2023-04-21
Invert the ipv6_disabled (to ipv6_enabled) flag and default to disabled

### LINKED ISSUE #1330: Set ipv6_disabled value to true in https://github.com/ansible/awx-operator/blob/devel/config/crd/bases/awx.ansible.com_awxs.yaml

### Bug Summary

Set ipv6_disabled value to true in https://github.com/ansible/awx-operator/blob/devel/config/crd/bases/awx.ansible.com_awxs.yaml.
I ran into this scenario where i had two instance with 0.19.0 operator i have upgraded them to 1.1.3 set the ipv6_disabled:true in my setting upgrade went fine no issues until after few days i went a head and created instance with 0.19.0 then the instance which got upgraded started crashing and reporting ngnix error.After full length research wat i found is that ipv6_disabled: true flag gets overwritten or removed  to false causing the existing instances which are upgraded to redeploy with new config map that contains ngnix trying to listen on ipv6 for this issue wat i did was i have modified ipv6_disabled: true in config/crd/bases/awx.ansible.com_awxs.yaml then ran make deploy that way ipv6_disabled: true is getting set during operator deployment and config map is created with ngnix listening on ipv4 instead of ipv6.Also if you look at present user base not many teams run ngnix awx on ipv6 so this will be much better way then setting default to ipv6.If this sounds good i will go head and modify crd of ipv6_disabled to true in config/crd/bases/awx.ansible.com_awxs.yaml.

### AWX Operator version

21.10.2

### AWX version

1.1.3

### Kubernetes platform

kubernetes

1.19

### Modifications

no

### Steps to reproduce

Create awx in two different namespaces with two different operator versions especially one which has ipv6 and then create operator instance which doesnt have ipv6 using make deploy restart the awx deployment in namespace which has ipv6 flag you will see awx-web container going down.

### Expected results

AWX-web container will report ngnix issue

### Actual results

Ngnix not supported family

### Additional information

N/A

### Operator Logs

N/A

justsomescripts:
##### SUMMARY

#1135 added a flag to disable IPv6 support in the Nginx server of the web container. That `ipv6_disabled` flag defaults to `false` which is a different behavior compared previous versions and leads to problems in the still common clusters without IPv6 support.

This PR inverts the `ipv6_disabled` flag to `ipv6_enabled` and leaves the default at `false`, so that IPv6 support must be explicitly enabled if needed. Ilso, it is (IMO) easier to follow the logic than to use the current double negative.

fixes #1330
##### ISSUE TYPE


TheRealHaoLiu:
I do agree that your implementation is better, current concern about merging this PR is that this will change an already established behavior and your PR will change the behavior 

It will be very complicated to preserve the previously acknowledged configuration (be it intentional or by default) 

At the moment I have no idea how we want to address the migration...

justsomescripts:
Thanks for the reply, totally makes sense. I just find the current flag kind of confusing (don't know of any other tools that do it this way) and it seems there are others that do as well, that's why I made that PR.

But of course, it's a breaking change which will prevent Nginx from starting in the worst case and I don't have any idea currently how to work around this and if it's possible at all.

rooftopcellist:
Like @TheRealHaoLiu mentioned, now that people have deployed with the default being `ipv6_disabled: false` will have that silently changed upon the next release. So this would stop working upon upgrade for anyone using an ipv6 environment.  

The default today, should work for IPv4 and IPv6, it only becomes an issue for deployments in IPv4 _only_ environments unless I am missing something.  That seems to be a special use case that doesn't cover as many users, which is why the current default was chosen (to cover the most number of users).  

Given that, I think we should leave it as is.

justsomescripts:
> That seems to be a special use case that doesn't cover as many users

Right, it only is an issue for single-stack IPv4 clusters. I thought they would be still quite common though, given that dual stack is only supported since Kubernetes 1.23 and not the default everywhere.

But if that's the standard nowadays it makes sense, would still be better IMO if the flag is `ipv6_enabled` with `true` as the default, but that probably wouldn't be worth the change.

PR #1386, opened 2023-05-01
web: Use sha1 filter instead of md5

dsavineau:
##### SUMMARY
This was fixed in 6cae8df but the task/web split rebase didn't apply this to the web deployment.
This prevents to deploy the operator when FIPS is enabled.

##### ISSUE TYPE

##### ADDITIONAL INFORMATION
```console
TASK [Apply deployment resources] ********************************
fatal: [localhost]: FAILED! =>

{"msg": "An unhandled exception occurred while running the lookup plugin 'template'. Error was a <class 'ValueError'>, original message: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS"}
```

### DIFF
--- a/roles/installer/templates/deployments/web.yaml.j2
+++ b/roles/installer/templates/deployments/web.yaml.j2
@@ -32,7 +32,7 @@ spec:
     "secrets/app_credentials",
     "storage/persistent",
   ] %}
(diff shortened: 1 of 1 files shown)


rooftopcellist (approved):
Thank you for patching this.  Good catch.

PR #1387, opened 2023-05-02
[CI] Fix failing AWX checks by pinning molecule dependency ansible-compat

relrod:
##### SUMMARY

https://github.com/ansible-community/molecule/issues/3903

##### ISSUE TYPE

### DIFF
--- a/molecule/requirements.txt
+++ b/molecule/requirements.txt
@@ -5,3 +5,4 @@ ansible-lint
 openshift!=0.13.0
 jmespath
 ansible-core
+ansible-compat<4  # https://github.com/ansible-community/molecule/issues/3903


AlanCoding:
Could you clarify exactly what this is fixing? I don't see checks in the operator failing, but I _do_ see checks in the AWX repo failing due to the error mentioned in the link you give.

AlanCoding:
Also, it would be good to have the title and the commit message have a well-formatted best-effort summary of the fix, for the purposes of release notes.

relrod:
> Could you clarify exactly what this is fixing? I don't see checks in the operator failing, but I _do_ see checks in the AWX repo failing due to the error mentioned in the link you give.

This should fix exactly that error. `ansible-compat` had a breaking change, molecule doesn't pin to the version before that change. So this works around that for now, until molecule is fixed (https://github.com/ansible-community/molecule/pull/3904)

AlanCoding (approved):

PR #1388, opened 2023-05-02
Fix variable name for loadbalancerip in template.

cilq:
##### SUMMARY
This fixes a bug in the template file of the service which checked a wrong variable thus never applying the loadbalancerip setting.

##### ISSUE TYPE

##### ADDITIONAL INFORMATION
none


fosterseth (approved):
good find, thanks for the fix

PR #1394, opened 2023-05-04
fix molecule tests by pinning docker

fosterseth:
##### SUMMARY

testing if this fixes recent CI failures

see https://github.com/ansible-collections/community.docker/issues/611

### DIFF
--- a/molecule/requirements.yml
+++ b/molecule/requirements.yml
@@ -5,5 +5,5 @@ collections:
     version: 2.3.2
   - name: operator_sdk.util
   - name: community.docker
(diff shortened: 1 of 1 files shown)


rooftopcellist (approved):

rooftopcellist:
@fosterseth thank you for looking at this.  After testing this out a bit, this PR doesn't work because when community.general was installed, it installed a more recent version of requests.  

So we needed to add a task to explicitly walk it back to a working version. I made those changes here:
* https://github.com/ansible/awx-operator/pull/1396

Issue #1383, opened 2023-04-27
Missing pre-stop hook on rsyslog in the task pod


TheRealHaoLiu:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that the AWX Operator is open source software provided for free and that I might not receive a timely response.

### AWX Operator version

devel

### AWX version

devel

### Kubernetes platform

kubernetes

### Kubernetes/Platform version

v1.24.12+ceaf338

### Modifications

no

### Steps to reproduce

rsyslog container in task pod does not have preStop hook 

### Expected results

rsyslog container don't terminate when terminationGracePeriod happens

### Actual results

rsyslog container was terminated

PR #1422, opened 2023-05-18
Updated task.yaml.j2 rsyslog container to  have preStop hook

rakesh561:
##### SUMMARY
Updated task.yaml.j2 rsyslog container to  have preStop hook

##### ISSUE TYPE

##### ADDITIONAL INFORMATION
This addresses missing pre-stop hook on rsyslog in the task pod and rsyslog container will terminate when terminationGracePeriod happens.
This will address hopefully following issue #1383

### DIFF
--- a/roles/installer/templates/deployments/task.yaml.j2
+++ b/roles/installer/templates/deployments/task.yaml.j2
@@ -392,7 +392,7 @@ spec:
               exec:
                 command:
                 - bash
-                - /var/lib/pre-stop/scripts/termination-master
+                - /var/lib/pre-stop/scripts/termination-waiter
 {% endif %}
           env:
             - name: SUPERVISOR_CONFIG_PATH
--- a/roles/installer/templates/deployments/task.yaml.j2
+++ b/roles/installer/templates/deployments/task.yaml.j2
@@ -382,6 +382,18 @@ spec:
             - name: awx-devel
               mountPath: "/awx_devel"
 {% endif %}
+{% if termination_grace_period_seconds is defined %}
+            - name: pre-stop-data
+              mountPath: /var/lib/pre-stop
+            - name: pre-stop-scripts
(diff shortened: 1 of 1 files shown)


rakesh561:
@rooftopcellist and @TheRealHaoLiu   When you have some time can you folks look into this PR.Appreciate in advance

roles/installer/templates/deployments/task.yaml.j2
```diff
@@ -381,6 +381,18 @@ spec:
...
+            - name: pre-stop-data
+              mountPath: /var/lib/pre-stop
+            - name: pre-stop-scripts
+              mountPath: /var/lib/pre-stop/scripts
+          lifecycle:
+            preStop:
+              exec:
+                command:
+                - bash
+                - /var/lib/pre-stop/scripts/termination-master
```
fosterseth: @TheRealHaoLiu @stanislav-zaprudskiy is the termination-master script intended to work with any container? would it need to be modified to work with rsyslog?

rakesh561:
@TheRealHaoLiu @stanislav-zaprudskiy  any feedback ?

roles/installer/templates/deployments/task.yaml.j2
stanislav-zaprudskiy: Generally the script requires just `bash`, mounted scripts and some system paths to be available. In this specific case, AFAICS, `rsyslog` runs the same image as `task` container - while the latter is known to run the scripts just fine.

Running `termination-master` as `preStop` hook only makes sense in a single container out of all containers of the Pod, the rest should be running `termination-waiter`. Right now the `task` container already runs it, so I would suggest switching to the waiter script for `rsyslog`.

```suggestion
                - /var/lib/pre-stop/scripts/termination-waiter
```

rakesh561:
Applied the suggested changes

fosterseth (approved):

Issue #1413, opened 2023-05-16
Web Containers Missing Security Context


rchaud:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that the AWX Operator is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Cannot start container as they start with Root privilege event with security_context_settings configured. Task yaml template contains the security_context_settings variable but web yaml is missing it.

https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/deployments/web.yaml.j2

I did not created a push since I am leaving this to the maintainer in case they want to decide to create a different settings for web and task or reuse security_context_settings on both. 

### AWX Operator version

2.0.0

### AWX version

22.0.0

### Kubernetes platform

kubernetes

### Kubernetes/Platform version

1.24

### Modifications

no

### Steps to reproduce

Deploy AWX with security_context_settings, you will see web container does not have that settings. 

### Expected results

Ability to change the user that runs the container.

### Actual results

security_context_settings is not configured

rchaud:
@TheRealHaoLiu @djyasin ^^^

rooftopcellist:
The current variable for the task deployment is called `task_privileged`, so I think we'll want to add a new variable called `web_privileged`.  What do you think @TheRealHaoLiu

rchaud:
> The current variable for the task deployment is called `task_privileged`, so I think we'll want to add a new variable called `web_privileged`. What do you think @TheRealHaoLiu

`task_privileged` or `web_privileged` seems to only run the containers privileged.

I am referring more to the security context so that we can change the FS USER, GROUP, etc

```
{% if (projects_persistence|bool and is_k8s|bool) or (security_context_settings|length) %}
      securityContext:
{% if projects_persistence|bool and is_k8s|bool %}
        fsGroup: 1000
{% endif %}
{% if security_context_settings|length %}
        {{ security_context_settings | to_nice_yaml | indent(8) }}
{% endif %}
```

All this code above seems to be missing for the web containers. And it is needed for us as Redis is trying to run as the root user and our Kubernetes environment has config in place to not allow containers running as root.

marekzebro:
[...]

PR #1425, opened 2023-05-19
modify container security context

marekzebro:
##### SUMMARY
Changes for adding/modifying securityContext of a container. It re-uses already present variable `security_context_settings`. All security context items can be added as key-value pair under a name of each container. 

##### ISSUE TYPE
 - New or Enhanced Feature
 #1413

### DIFF
--- a/README.md
+++ b/README.md
@@ -672,7 +672,7 @@ spec:
 ```
 
 #### Redis container capabilities
-> :warning: The `redis_capabilities` option is depracated. Please use `security_context_settings` to set any securityContext related settings such as container capabilites.
+> :warning: The `redis_capabilities` option is deprecated. Please use `security_context_settings` to set any securityContext related settings such as container capabilites.
 
 Depending on your kubernetes cluster and settings you might need to grant some capabilities to the redis container so it can start. Set the `redis_capabilities` option so the capabilities are added in the deployment.
 
--- a/README.md
+++ b/README.md
@@ -672,6 +672,7 @@ spec:
 ```
 
 #### Redis container capabilities
+> :warning: The `redis_capabilities` option is depracated. Please use `security_context_settings` to set any securityContext related settings such as container capabilites.
 
 Depending on your kubernetes cluster and settings you might need to grant some capabilities to the redis container so it can start. Set the `redis_capabilities` option so the capabilities are added in the deployment.
 
--- a/README.md
+++ b/README.md
@@ -655,7 +655,7 @@ spec:
 **Note**: The `image` and `image_version` are intended for local mirroring scenarios. Please note that using a version of AWX other than the one bundled with the `awx-operator` is **not** supported. For the default values, check the [main.yml](https://github.com/ansible/awx-operator/blob/devel/roles/installer/defaults/main.yml) file.
 
 #### Container security context
-To change/add the security context for a specified container, add your items as key-value pairs under the nbame of the container in question. Possible container names are `redis`, `init`, `ee`, `task`, `web`, `rsyslog`. 
+To change/add the security context for a specified container, add your items as key-value pairs under the name of the container in question. Possible container names are `redis`, `init`, `ee`, `task`, `web`, `rsyslog`. 
 
 ```yaml
 ---
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ An [Ansible AWX](https://github.com/ansible/awx) operator for Kubernetes built w
          * [Managed PostgreSQL Service](#managed-postgresql-service)
       * [Advanced Configuration](#advanced-configuration)
          * [Deploying a specific version of AWX](#deploying-a-specific-version-of-awx)
+         * [Container security context](#container-security-context)
          * [Redis container capabilities](#redis-container-capabilities)
          * [Privileged Tasks](#privileged-tasks)
          * [Containers Resource Requirements](#containers-resource-requirements)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -643,6 +644,20 @@ spec:
 
 **Note**: The `image` and `image_version` are intended for local mirroring scenarios. Please note that using a version of AWX other than the one bundled with the `awx-operator` is **not** supported. For the default values, check the [main.yml](https://github.com/ansible/awx-operator/blob/devel/roles/installer/defaults/main.yml) file.
 
+#### Container security context
+To change/add the security context for a specified container, add your items as key-value pairs under the nbame of the container in question. Possible container names are `redis`, `init`, `ee`, `task`, `web`, `rsyslog`.
```
fosterseth: nbame > name

marekzebro: Thanks for spotting it, I have fixed typo in README file.

marekzebro:
Is this change acceptable?

README.md
```diff
@@ -643,6 +644,20 @@ spec:
...
+    task:
+      runAsUser: 1000
+      runAsGroup: 1000
+    rsyslog:
+      runAsUser: 1000
+```
+
 #### Redis container capabilities
 
 Depending on your kubernetes cluster and settings you might need to grant some capabilities to the redis container so it can start. Set the `redis_capabilities` option so the capabilities are added in the deployment.
```
rooftopcellist, line 677: I think it may make more sense to change the example here to just use `security_context_settings.redis` parameter.  That way we don't have users attempting to use both methods of setting capabilities and end up with one overriting the other.  And we could mark `redis_capabilities` as a deprecated field (but leave in the conditional in the template for backwards compat in the meantime).  What do you think?

rooftopcellist:
@marekzebro Thank you for your contribution!  Yes, this is a welcome change, sorry for the wait, I was OOO, back now.  

Could you rebase this PR? I left one comment on the readme section.  This is also missing an entry in the CSV for the Openshift form descriptors.  Something like this [here](https://github.com/ansible/awx-operator/blob/devel/config/manifests/bases/awx-operator.clusterserviceversion.yaml#LL758C1-L762C52) should be sufficient:

```
      - displayName: Custom Security Context
        path: security_context_settings
        x-descriptors:
```

README.md
marekzebro: Thanks for your comment, I made changes to readme file, as well as Openshift form descriptors. How would you like to mark field`redis_capabilities` as depracated?  Would be sufficient just saying it readme, i.e. this field will become deprecated in future version, please use `security_context_settings` instead?

rchaud:
@marekzebro  @rooftopcellist @fosterseth  Is there an ETA on this?

roles/installer/templates/deployments/web.yaml.j2
```diff
@@ -276,6 +307,7 @@ spec:
               mountPath: "/var/run/redis"
             - name: rsyslog-socket
               mountPath: "/var/run/awx-rsyslog"
+          resources: {{ rsyslog_resource_requirements }}
```
rooftopcellist: @marekzebro I think you are going to need to move this line back down below the bundle_cacert_secret block or else there will be templating errors when bundle_cacert_secret is defined.

rooftopcellist (changes requested):

roles/installer/templates/deployments/web.yaml.j2
marekzebro: You are right, it's been duplicated as it was already on line 335.

roles/installer/templates/deployments/web.yaml.j2
```diff
@@ -275,7 +306,7 @@ spec:
             - name: {{ ansible_operator_meta.name }}-redis-socket
               mountPath: "/var/run/redis"
             - name: rsyslog-socket
-              mountPath: "/var/run/awx-rsyslog"
+              mountPath: "/var/run/awx-rsyslog"
```
rooftopcellist: @marekzebro Could you please rebase and remove this whitespace?

marekzebro: @rooftopcellist rebased and fixed. Please let me know in case you can see anything else.

rooftopcellist:
@marekzebro This is a good change, and it is well orchestrated. However, we intend to restructure our CRD schema to follow a format more like the following for deployment specific settings like this:

```
---
spec:
  web:
    init:
      security_context_settings:
        capabilities:
          add:
            - CHOWN
            - SETUID
            - SETGID
        runAsUser: 1000
        runAsGroup: 1000
    web:
      security_context_settings:
        capabilities:
          add:
            - CHOWN
            - SETUID
            - SETGID
        runAsUser: 1000
        runAsGroup: 1000
```

Notice that the setting are nested under the deployment name (`web` in  this context), then the container inside that deployment (for example `init`).  Then the variables would be referenced in the templates as `web.init.security_context_settings`, for example.  

I don't think we want to merge this in it's current state because users would start using it in their AWX specs, then when we ultimately refactor, those users would have functionality be broken (their security context settings would not be applied on the deployments as the key ref wouldn't be evaluated).  

Here is an example of what type of nested deployment options we would like to get to:
* https://github.com/ansible/eda-server-operator/blob/main/config/crd/bases/eda.ansible.com_edas.yaml#L88-L114

This PR might be a good place to start.  @marekzebro feel free to reach out in Matrix if you want help with this.

rooftopcellist:
For now, let's move forward with adding the security context param per deployment in this format:
* web_security_context
* task_security_context
* etc.

This way, all of v1beta1 will use the same `web_<setting>` format.  

Then, when we do a v2 awx apiVersion down the line, we can adopt the new schema I described in the earlier comment.

PR #1410, opened 2023-05-16
Added capability to set the rsyslog container resources 

rakesh561:
##### SUMMARY
This commit allows users to set requests and limits on rsyslog containers and address issue #1380 

##### ISSUE TYPE
 - New or Enhanced Feature

PR #1426, opened 2023-05-19
Fix rsyslog container resources in task deployment template

dale-mittleman:
##### SUMMARY

In https://github.com/ansible/awx-operator/pull/1410, a field named rsyslog_resource_requirements was added to the AWX CRD.  This field allows users to set requests and limits on the rsyslog container in both the web and task deployments.

In the task deployment template, the rsyslog resources were applied in the wrong location.  They were placed within the configuration block for the execution environment container, overriding the ee container resources, and failing to set the rsyslog resources as was intended.

This mistake called out in this comment: https://github.com/ansible/awx-operator/pull/1410#issuecomment-1554299712

This PR moves the rsyslog resources config to a better location in the task deployment template.

##### ISSUE TYPE

### DIFF
--- a/roles/installer/templates/deployments/task.yaml.j2
+++ b/roles/installer/templates/deployments/task.yaml.j2
@@ -343,7 +343,6 @@ spec:
 {% if ee_extra_env -%}
             {{ ee_extra_env | indent(width=12, first=True) }}
 {% endif %}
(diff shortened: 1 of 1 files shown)


fosterseth:
@dale-mittleman thanks for fixing this!

fosterseth (approved):

Issue #1171, opened 2022-12-27
[redis] Custom config is needed


timansky:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that AWX Operator is open source software provided for free and that I might not receive a timely response.

### Feature Summary

ISSUE TYPE
Feature Idea

Now there is no options to pass to configure redis. Only way is to compile own image.
It will be good to pass config parameters (custom args, setting config configmap or custom mount) like in others containers such as web ee or task)

PR #1427, opened 2023-05-22
Allow users to define custom config for Redis Container

rakesh561:
Allow users to define custom config for Redis Container

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
This will address the following issue #1171

### DIFF
--- a/README.md
+++ b/README.md
@@ -1025,6 +1025,28 @@ data:
       INSIGHTS_URL_BASE = "example.org"
       AWX_CLEANUP_PATHS = True
 ```
+
+Example configuration for Redis ConfigMap
+
+```
+`yaml
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: <resourcename>-extra-config
+  namespace: <target namespace>
+data:
+  redis.conf: |
+     daemonize yes
+     unixsocket /var/run/redis/redis.sock
+     unixsocketperm 777
+     port 0
+     bind 127.0.0.1
+     maxmemory 2mb
+     maxmemory-policy allkeys-lru
+
+```
 Example spec file for volumes and volume mounts
 
 ```yaml
@@ -1046,6 +1068,15 @@ Example spec file for volumes and volume mounts
               - key: custom.py
                 path: custom.py
             name: <resourcename>-extra-config
+
+        - name: redis-cfg
+          configMap:
+           defaultMode: 420
+           items:
+            - key: redis.conf
+              path: redis.cong
+           name: redis-config
+
         - name: shared-volume
           persistentVolumeClaim:
             claimName: my-external-volume-claim
@@ -1070,6 +1101,12 @@ Example spec file for volumes and volume mounts
           subPath: custom.py
         - name: shared-volume
           mountPath: /shared
+
+      redis_extra_volume_mounts: |
+        - name: redis.cfg
+           mountPath: /etc/redis.conf
+           subPath: redis.conf
+
 ```
 
 > :warning: **Volume and VolumeMount names cannot contain underscores(_)**
--- a/README.md
+++ b/README.md
@@ -998,7 +998,8 @@ In a scenario where custom volumes and volume mounts are required to either over
 | extra_volumes                      | Specify extra volumes to add to the application pod      | ''      |
 | web_extra_volume_mounts            | Specify volume mounts to be added to Web container       | ''      |
 | task_extra_volume_mounts           | Specify volume mounts to be added to Task container      | ''      |
-| ee_extra_volume_mounts             | Specify volume mounts to be added to Execution container | ''      |
+| ee_extra_volume_mounts            | Specify volume mounts to be added to Execution container | ''       |
+| redis_extra_volume_mounts          | Specify volume mounts to be added to redis container.    | ''      |
 | init_container_extra_volume_mounts | Specify volume mounts to be added to Init container      | ''      |
 | init_container_extra_commands      | Specify additional commands for Init container           | ''      |
 
@@ -1159,11 +1160,13 @@ type: kubernetes.io/dockerconfigjson
 
 If you need to export custom environment variables to your containers.
 
-| Name           | Description                                         | Default |
-| -------------- | --------------------------------------------------- | ------- |
-| task_extra_env | Environment variables to be added to Task container | ''      |
-| web_extra_env  | Environment variables to be added to Web container  | ''      |
-| ee_extra_env   | Environment variables to be added to EE container   | ''      |
+| Name            | Description                                         | Default |
+| --------------  | --------------------------------------------------- | ------- |
+| task_extra_env  | Environment variables to be added to Task container | ''      |
+| web_extra_env   | Environment variables to be added to Web container  | ''      |
+| ee_extra_env    | Environment variables to be added to EE container   | ''      |
+| redis_extra_env | Environment variables to be added to redis container   | ''      |
+
 
 > :warning: The `ee_extra_env` will only take effect to the globally available Execution Environments. For custom `ee`, please [customize the Pod spec](https://docs.ansible.com/ansible-tower/latest/html/administration/external_execution_envs.html#customize-the-pod-spec).
 
@@ -1180,6 +1183,9 @@ Example configuration of environment variables
     ee_extra_env: |
       - name: MYCUSTOMVAR
         value: foo
+    redis_extra_env: |
+      - name: MYCUSTOMVAR
+        value: food
 ```
 
 #### CSRF Cookie Secure Setting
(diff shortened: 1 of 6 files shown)


roles/installer/defaults/main.yml
```diff
@@ -286,6 +286,9 @@ web_command: []
 rsyslog_args:
   - /usr/bin/launch_awx_rsyslog.sh
 rsyslog_command: []
+redis_args:
+  - /usr/bin/launch_awx_redis.sh
```
fosterseth: this file doesn't exist, so I think redis_args can be empty by default

fosterseth:
thanks for this PR, this sounds like the right approach

Does this allow users to drop in a redis.conf file? then the supplied redis command/args would start the redis instance from that config?

can you provide some steps as comment on the PR on what that would look like?

rakesh561:
@fosterseth I will provide some steps

rakesh561:
@fosterseth  any one who wants to customize redis container can try out following steps.It would be nice if we can run these tests through github actions if possible

apiVersion: v1
kind: ConfigMap
metadata:
  name: redis-config
  namespace: awx
data:
  redis.conf: |
daemonize yes
unixsocket /var/run/redis/redis.sock
unixsocketperm 777
port 0
bind 127.0.0.1
 maxmemory 2mb
 maxmemory-policy allkeys-lru 



---
    spec:
    ...
      extra_volumes: |
        - name: ansible-cfg
          configMap:
            defaultMode: 420
            items:
              - key: redis.conf
                path: redis.cong
            name: redis-config


      redis_extra_volume_mounts: |
        - name: redis.cfg
          mountPath: /etc/redis.conf
          subPath: redis.conf

      redis_command:
       chmod 755 /etc/redis.conf
       redis-cli start 

     redis_extra_env:
          - name: foo
             value: bar

rakesh561:
@fosterseth  let me know what you think of above inputs or commands.

rakesh561:
@fosterseth @rooftopcellist  just checking to see if you can take a look at it

djyasin:
@rakesh561 Thank you so much for your contribution! It looks like we have some merge conflicts. Would you mind rebasing this PR for us?

rakesh561:
@djyasin  i have done the rebasing.Please let me know how it looks from your end.

roles/installer/templates/deployments/task.yaml.j2
```diff
@@ -382,6 +387,10 @@ spec:
             - name: awx-devel
               mountPath: "/awx_devel"
 {% endif %}
+<<<<<<< HEAD
```
rooftopcellist: @rakesh561 it looks like there was a mistake when resolving conflicts.

roles/installer/templates/deployments/task.yaml.j2
```diff
@@ -382,6 +387,8 @@ spec:
             - name: awx-devel
               mountPath: "/awx_devel"
 {% endif %}
+{% if redis_extra_volume_mounts -%}
+            {{ redis_extra_volume_mounts | indent(width=12, first=True) }}
 {% if termination_grace_period_seconds is defined %}
```
rooftopcellist, line 399: It looks like the redis_extra_env is missing here:

```suggestion
{% if redis_extra_env -%}
            {{ redis_extra_env | indent(width=12, first=True) }}
{% endif %}  
{% if termination_grace_period_seconds is defined %}
```

roles/installer/defaults/main.yml
```diff
@@ -286,6 +286,8 @@ web_command: []
 rsyslog_args:
   - /usr/bin/launch_awx_rsyslog.sh
 rsyslog_command: []
+redis_args: ''
+redis_command: []
```
rooftopcellist: It would be good to add docs in the README.md for these as well with an example use case.  But this is a niche case, so not strictly required for merge.

roles/installer/templates/deployments/task.yaml.j2
```diff
@@ -142,6 +142,11 @@ spec:
         - image: '{{ _redis_image }}'
           imagePullPolicy: '{{ image_pull_policy }}'
           name: redis
+{% if redis_command %}
+          command: {{ redis_command }}
+{% endif %}
+{% if redis_args %}
+          args: {{ redis_args }}
```
rooftopcellist: This line already exists:
* https://github.com/ansible/awx-operator/pull/1427/files#diff-4d0fa82fa796130ee45f174f37b43ce5faf7092d12717ad484dbdf712ff24988R155

You'll probably want to set the default to 

```
["redis-server", "/etc/redis.conf"]
```

I think that should be the value for `redis_command`, not `redis_args`, but the only way to know for sure will be to deploy it and try it out once those changes are made.

roles/installer/templates/deployments/web.yaml.j2
```diff
@@ -126,6 +126,12 @@ spec:
         - image: '{{ _redis_image }}'
           imagePullPolicy: '{{ image_pull_policy }}'
           name: redis
+ {% if redis_command %}
+          command: {{ redis_command }}
+{% endif %}
+{% if redis_args %}
+          args: {{ redis_args }}
```
rooftopcellist, line 133: Same here, this already exists, currently, your templaated redis_args wouldn't be used at all, the entry further down on line 140 would be used instead ([here](https://github.com/ansible/awx-operator/pull/1427/files#diff-a4af5f6d20ffe00a9f00fda1663b933b856c4a57435a796c4bad888fef83730dR140)).

rakesh561:
@rooftopcellist  Updated README with example and also applied suggested changes

README.md
```diff
@@ -998,7 +998,8 @@ In a scenario where custom volumes and volume mounts are required to either over
 | extra_volumes                      | Specify extra volumes to add to the application pod      | ''      |
 | web_extra_volume_mounts            | Specify volume mounts to be added to Web container       | ''      |
 | task_extra_volume_mounts           | Specify volume mounts to be added to Task container      | ''      |
-| ee_extra_volume_mounts             | Specify volume mounts to be added to Execution container | ''      |
+| ee_extra_volume_mounts            | Specify volume mounts to be added to Execution container | ''       |
```
rooftopcellist:
```suggestion
| ee_extra_volume_mounts             | Specify volume mounts to be added to Execution container | ''       |
```

config/crd/bases/awx.ansible.com_awxs.yaml
```diff
@@ -1566,6 +1574,9 @@ spec:
               web_extra_volume_mounts:
                 description: Specify volume mounts to be added to the Web container
                 type: string
+              redis_extra_volume_mounts:
+                description: Specify volume mounts to be added to the Redis container
+                type: string
               redis_image:
                 description: Registry path to the redis container to use
```
rooftopcellist, line 1581: It looks like `redis_extra_env` is missing here.

roles/installer/templates/deployments/task.yaml.j2
rooftopcellist: merge conflict needs to be resolved here

config/crd/bases/awx.ansible.com_awxs.yaml
rooftopcellist: @rakesh561 This is still needed.

roles/installer/templates/deployments/task.yaml.j2
rooftopcellist: @rakesh561 This is still needed

roles/installer/templates/deployments/task.yaml.j2
```diff
@@ -382,6 +387,15 @@ spec:
             - name: awx-devel
               mountPath: "/awx_devel"
 {% endif %}
+{% if redis_extra_volume_mounts -%}
+            {{ redis_extra_volume_mounts | indent(width=12, first=True) }}
+<<<<<<< HEAD
+=======
+{% endif %}
+>>>>>>> f32bd4e (Updated requested changes)
```
TheRealHaoLiu, line 395: can you address the merge conflict

TheRealHaoLiu:
https://github.com/ansible/awx-operator/pull/1695/files
try this out and let me know if this implementation allow you to ovverride the things you want

Denney-tech:
@TheRealHaoLiu and @kurokobo , neither of your PR's #1695 or #1697 appear to address modifying the [redis.conf](https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/configmaps/config.yaml.j2#L233) file that is mounted to the Redis container. Not everything can be defined as an environment variable, so being able to customize the container specs doesn't necessarily help here. 

What we need is a way to modify or replace the settings in redis.conf in the configmap created by the operator, perhaps by adding something like `redis_extra_settings:` to the awx CRD. Then it could do a dynaconf merge similar to how `extra_settings:` works.

kurokobo:
@Denney-tech 
Yes, you are correct. My and Hao's PRs are focused on pod spec customization, not configmap customization.
As you mentiond changing conf file requires a different way (not sure whether to continue that in this PR or create it as a separate PR).

Denney-tech:
Whatever you brilliant devs think would work the best. :)

Just wanted to reply to Hao's [comment](https://github.com/ansible/awx-operator/pull/1427#issuecomment-1908999181) that his PR, and in turn yours, didn't satisfy the objective of this PR.

TheRealHaoLiu:
@Denney-tech thanks for checking our our PR and provide valuable feedback

PR #1430, opened 2023-05-25
Update README.md

mahoutukaisali:
##### SUMMARY
Added a note of extra_settings parameters to warn users that those parameters cannot change after deploying awx. 
This is critical information because if users need to change those parameters after deploying awx, there is only one way that 
they need to rebuild instances.

##### ISSUE TYPE

### DIFF
--- a/README.md
+++ b/README.md
@@ -1220,6 +1220,8 @@ With`extra_settings`, you can pass multiple custom settings via the `awx-operato
 | -------------- | -------------- | ------- |
 | extra_settings | Extra settings | ''      |
 
+**Note:** Parameters configured in `extra_settings` cannot change after deployed awx.
+
 Example configuration of `extra_settings` parameter
 
 ```yaml
--- a/README.md
+++ b/README.md
@@ -1220,7 +1220,7 @@ With`extra_settings`, you can pass multiple custom settings via the `awx-operato
 | -------------- | -------------- | ------- |
 | extra_settings | Extra settings | ''      |
 
-**Note:** Parameters configured in `extra_settings` cannot change after deployed awx.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1220,6 +1220,8 @@ With`extra_settings`, you can pass multiple custom settings via the `awx-operato
 | -------------- | -------------- | ------- |
 | extra_settings | Extra settings | ''      |
 
+**Note:** Parameters configured in `extra_settings` cannot change after deployed awx.
```
shanemcd: I don't believe this is true... @rooftopcellist @TheRealHaoLiu is there a bug here?

TheRealHaoLiu: i dont think this is true, it will cause a rolling restart so that the setting takes effect i think

TheRealHaoLiu:
@mahoutukaisali changing the `extra_settings` is allowed after awx is deployed, changing extra_settings will cause update to the deployment of awx pods which causes a rolling restart of awx. 

can you elaborate on what you mean by  "`extra_settings` cannot change after deployed awx"

mahoutukaisali:
@TheRealHaoLiu I mean, values configured in `extra_settings` cannot change from Web UI.

README.md
rooftopcellist: Can we add a bit more detail to the doc note?

```suggestion
**Note:** Parameters configured in `extra_settings` are set as read-only settings in AWX.  As a result, they cannot be changed in the UI after deployment. If you need to change the setting after the initial deployment, you need to change it on the AWX CR spec.  
```

rooftopcellist (approved):

Issue #1432, opened 2023-05-26
Supervisord cannot start in scenario when k8s prohibits write access to pod's file system


marekzebro:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that the AWX Operator is open source software provided for free and that I might not receive a timely response.

### AWX Operator version

2.2.1

### AWX version

22.3.0

### Kubernetes platform

kubernetes

### Kubernetes/Platform version

v1.24.10

### Modifications

no

### Steps to reproduce

- Prohibit write access to container's root file system
- Run standard installation. 

### Expected results

Container should start normally in such scenario. Ideally there would be a mount point where container can freely write. 

### Actual results

Supervisord on both task & web container will fail due to inability to write into: /var/run/supervisor/

```
cat /etc/supervisord_web.conf
...
[unix_http_server]
file=/var/run/supervisor/supervisor.web.sock

[supervisorctl]
serverurl=unix:///var/run/supervisor/supervisor.web.sock ; use a unix:// URL  for a unix socket
```

### Additional information

There are number of emptyDir mounted already, this looks like has been forgotten. Such change would fix this problem right away. 

```
volumes:
  - name: {{ ansible_operator_meta.name }}-supervisord-socket
     emptyDir: {}
       
volumeMounts:
  - name: {{ ansible_operator_meta.name }}-supervisord-socket
     mountPath: "/var/run/supervisor"
```

TheRealHaoLiu:
can you elaborate on what configuration did u made to "Prohibit write access to container's root file system" so we can try to reproduce this issue

is this with `readOnlyRootFilesystem`  https://kubernetes.io/docs/tasks/configure-pod-container/security-context/ 

is there any other problem that you run into after fixing the supervisor socket file

marekzebro:
It is precisely `readOnlyRootFilesystem` pod security context. There are number of issues I'm experiencing due to read only filesystem. However, issue with supervisord is new.

I opened other issue while ago ( #571 ), where I documented some file system problems. Some of them were fixed/changed already.

Current issues in web container are:
- /tmp ( not writable )
- /var/run/supervisor ( socket file )
- fifo file cannot be created in /var/lib/awx ( https://github.com/ansible/awx/issues/14052 )

Task container issues:
- /tmp
[...]

PR #1433, opened 2023-05-26
Create mount point for supervisord

marekzebro:
##### SUMMARY
Create directory for supervisord allowing writing its socket file.

Please see:
#1432

##### ISSUE TYPE

### DIFF
--- a/roles/installer/templates/deployments/task.yaml.j2
+++ b/roles/installer/templates/deployments/task.yaml.j2
@@ -220,6 +220,8 @@ spec:
             - name: {{ ansible_operator_meta.name }}-redis-socket
               mountPath: "/var/run/redis"
             - name: rsyslog-socket
+            - name: {{ ansible_operator_meta.name }}-supervisor-socket
+              mountPath: "/var/run/supervisor"
               mountPath: "/var/run/awx-rsyslog"
             - name: "{{ ansible_operator_meta.name }}-receptor-config"
               mountPath: "/etc/receptor/"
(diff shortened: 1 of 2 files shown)


fosterseth:
@marekzebro Hi! thanks for this PR

A couple of notes:

1. If the end goal is to get AWX working with `readOnlyRootFilesystem`, we'd like to see one big PR that has all of the changes necessary to get it working. We don't think this one change is enough to get it there. We haven't looked into this ourselves, and we don't know how feasible it is for AWX. There are likely a lot of places in the code that attempts to write to root file partition.
2. mounting in emptyDir technically works, but isn't ideal. Instead we'd prefer a solution that, for example, changes socket file path to a non-root directory (for example `/home/user`).

marekzebro:
@fosterseth Thanks for reply. 

1. I can certainly prepare PR with all necessary changes for it, as I've been fixing (overriding) them for a while. However, supervisord issue has been introduced recently. Sometimes fixing small issues is easier than solving one big chunk of them. 
2. Socket files are usually placed in `/var/run` or `/run`, in any way placing socket elsewhere would create exactly same problem without having a writable mount available in this location.

djyasin:
Hello @marekzebro 
We have a few requests to help us with processing this PR.
We would like to be able to implement these changes at one time rather than incrementally. Could you place this PR behind a feature flag? 
We would also like to have all of the related issues linked to this PR. Could we have you tag those here as well, please?
We also would prefer to have the sockets in a single dir rather than in temp. 

Thank you again for all of your time!

PR #1434, opened 2023-05-27
Updated backup and restore roles so that node selector can be passed through spec

rakesh561:
##### SUMMARY
Updated backup and restore roles so that node selector can be passed through spec.
Customer can pass postgres_selector in spec to backup & restore awx deployment on ARM64.
Mainly when some one deploys there deployment on AMD64 want to backup/restore on ARM64 this will help.
I could be wrong but i did my research correct me over my understanding is if some one passes postgres_label_selector during there deployment then it will used during backup and restore so technically if some deploys on amd64 but want to backup and restore on arm64 it wont be able to do that as the fact will contain only amd64 label.Again appreciate if some one can give more insights.
 
##### ISSUE TYPE

- New or Enhanced Feature

##### ADDITIONAL INFORMATION
#1226 will be addressed.


config/crd/bases/awx.ansible.com_awxbackups.yaml
```diff
@@ -90,6 +90,9 @@ spec:
               postgres_image_version:
                 description: PostgreSQL container image version to use
                 type: string
+              postgres_selector:
```
TheRealHaoLiu: i think postgres_selector is not very descriptive

maybe something like `db_management_pod_node_select`?

roles/backup/defaults/main.yml
```diff
@@ -17,6 +17,15 @@ no_log: true
 # Variable to set when you want backups to be cleaned up when the CRD object is deleted
 clean_backup_on_delete: false
 
+
+# Add a nodeSelector for the Postgres pods to backup.
+# Specify as literal block. E.g.:
+# postgres_selector: |
+#   disktype: ssd
```
TheRealHaoLiu: can we remove `disktype` from the example

TheRealHaoLiu:
Thank you for your contribution. The implementation looks solid, we will test it. I made some comment around naming of the parameter.

rakesh561:
@TheRealHaoLiu thanks for the review and suggestions.I have updated the PR with suggested changes

roles/backup/templates/management-pod.yml.j2
```diff
@@ -24,6 +24,10 @@ spec:
     resources:
       {{ backup_resource_requirements | to_nice_yaml(indent=2) | indent(width=6, first=False) }}
 {%- endif %}
+{% if db_management_pod_node_selector %}
+      nodeSelector:
```
TheRealHaoLiu:
```suggestion
  nodeSelector:
```

TheRealHaoLiu: nodeSelector is at the same level as containers, the indention here looks incorrect

TheRealHaoLiu:
@rooftopcellist can I get a second look on this PR. LGTM

rakesh561:
Thank you @rooftopcellist  i know you are busy let me know what you think about the PR

jjwatt (approved):
LGTM

TheRealHaoLiu:
updated your PR description so our CI will pass

Thank you for your contribution! I will squash merge this PR.

PR #1443, opened 2023-06-07
Pre-create event table partitions before db backup in operator

jessicamack:
##### SUMMARY

It was discovered that pg_dump will block the events table partitioning which ends up blocking jobs from being executed during pg_dump. The solution is pre-creating partitions for the table so we wont try to create it before job execution. This PR adds a task that will run the new awx-manage command to pre-create events table partition before executing pg_dump.

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

```

```


roles/backup/defaults/main.yml
```diff
@@ -36,4 +36,7 @@ additional_labels: []
 
 # Maintain some of the recommended `app.kubernetes.io/*` labels on the resource (self)
 set_self_labels: true
+
+# Number of partitions to precreate to avoid pg_dump locks.
+partition_count: 1
```
TheRealHaoLiu:
```suggestion
precreate_partition_hours: 3
```

roles/backup/tasks/postgres.yml
```diff
@@ -82,6 +82,16 @@
     resolvable_db_host: '{{ (awx_postgres_type == "managed") | ternary(awx_postgres_host + "." + ansible_operator_meta.namespace + ".svc.cluster.local", awx_postgres_host) }}'  # yamllint disable-line rule:line-length
   no_log: "{{ no_log }}"
 
+- name: Precreate database partitions
+  k8s_exec:
+    namespace: '{{ ansible_operator_meta.namespace }}'
+    pod: "{{ tower_pod_name }}"
```
TheRealHaoLiu: just do double check, can you show me where is this var defined?

roles/backup/tasks/postgres.yml
```diff
@@ -82,6 +82,16 @@
...
+    container: "{{ ansible_operator_meta.name }}-task"
+    command: awx-manage precreate_partitions --count='{{ partition_count }}'
+  register: result
+  changed_when: "'Created partitions for' in result.stderr"
+  failed_when: result.return_code > 0
```
relrod: This should be the default fail condition already.

roles/backup/defaults/main.yml
TheRealHaoLiu: i think `precreate_partition_hours` better capture the intention here

It's number of hours worth of events table partitions to precreate

jessicamack: renamed the variable here and in the automation-platform-collection PR. updated the comment descriptions

roles/backup/tasks/postgres.yml
jessicamack: I think I carried over a name from automation-platform-collection when I was working on both PRs at the same time. Not sure where I pulled this from but it's been updated to "{{ awx_task_pod_name }}" which I believe is defined in the resources_configuration.yml

roles/backup/tasks/postgres.yml
jessicamack: removed

roles/backup/tasks/postgres.yml
TheRealHaoLiu: does resources_configuration.yml get run during backup?

AlanCoding (approved):

roles/backup/defaults/main.yml
```diff
@@ -44,4 +44,7 @@ additional_labels: []
 
 # Maintain some of the recommended `app.kubernetes.io/*` labels on the resource (self)
 set_self_labels: true
+
+# Number of hours worth of events table partitions to precreate before starting backup to avoid pg_dump locks.
```
gundalow:
```suggestion
# Number of whole hours worth of events table partitions to precreate before starting backup to avoid pg_dump locks.
```

Also does `precreate_partition_hours: 0` do anything bad?

jessicamack: @gundalow No, in the precreate_partitions command we check that the provided hours count is greater than 0. If it's 0 it will just skip the partition creation process.

roles/backup/tasks/postgres.yml
```diff
@@ -82,6 +82,15 @@
     resolvable_db_host: '{{ (awx_postgres_type == "managed") | ternary(awx_postgres_host + "." + ansible_operator_meta.namespace + ".svc.cluster.local", awx_postgres_host) }}'  # yamllint disable-line rule:line-length
   no_log: "{{ no_log }}"
 
+- name: Precreate database partitions
+  k8s_exec:
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    pod: "{{ awx_task_pod_name }}"
```
TheRealHaoLiu, line 113: undefined variable `awx_task_pod_name`

roles/backup/tasks/postgres.yml
```diff
@@ -82,6 +82,15 @@
...
+    container: "{{ ansible_operator_meta.name }}-task"
+    command: awx-manage precreate_partitions --count='{{ precreate_partition_hours }}'
+  register: result
+  changed_when: "'Created partitions for' in result.stdout"
+
```
TheRealHaoLiu: do not run this block when precreate_partition_hours==0

TheRealHaoLiu: example https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_conditionals.html#basic-conditionals-with-when

roles/backup/defaults/main.yml
```diff
@@ -44,4 +44,7 @@ additional_labels: []
 
 # Maintain some of the recommended `app.kubernetes.io/*` labels on the resource (self)
 set_self_labels: true
+
+# Number of whole hours worth of events table partitions to precreate before starting backup to avoid pg_dump locks.
+precreate_partition_hours: 3
```
TheRealHaoLiu, line 49: this variable need to be exposed to the end user via 
https://github.com/ansible/awx-operator/blob/devel/config/crd/bases/awx.ansible.com_awxbackups.yaml

also descriptor for CSV in https://github.com/ansible/awx-operator/blob/devel/config/manifests/bases/awx-operator.clusterserviceversion.yaml#L29

TheRealHaoLiu: information about what a CRD is https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/ and what https://github.com/ansible/awx-operator/blob/devel/config/crd/bases/awx.ansible.com_awxbackups.yaml does

roles/backup/tasks/postgres.yml
```diff
@@ -82,6 +82,15 @@
     resolvable_db_host: '{{ (awx_postgres_type == "managed") | ternary(awx_postgres_host + "." + ansible_operator_meta.namespace + ".svc.cluster.local", awx_postgres_host) }}'  # yamllint disable-line rule:line-length
   no_log: "{{ no_log }}"
 
+- name: Precreate database partitions
+  k8s_exec:
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    pod: "{{ awx_task_pod_name }}"
+    container: "{{ ansible_operator_meta.name }}-task"
```
TheRealHaoLiu: `ansible_operator_meta.name` will resolve to the awxbackup resource not the awx resource's name

roles/backup/tasks/postgres.yml
TheRealHaoLiu: so you probably need a task in front of this to find awx_task_pod_name 

that takes `deployment_name` to find the awx resource than use that to find the name of a task pod for that deployment

TheRealHaoLiu (changes requested):
see previous comments

AlanCoding (dismissed):
^ as per Hao review

AlanCoding:
I'd like to formalize what else needs to be done with this. If its a matter of testing, we can get help to deploy it, but it also looks like variables to expose these things in the CRD are not in place yet. If there's anything unclear about that, we should reach out to get help with that.

jessicamack:
> I'd like to formalize what else needs to be done with this. If its a matter of testing, we can get help to deploy it, but it also looks like variables to expose these things in the CRD are not in place yet. If there's anything unclear about that, we should reach out to get help with that.

@AlanCoding I have the variable exposure and CRD work already done. I'll push it now. I was working through retrieving the correct variables for the postgres.yml task when the 3.11 upgrade work came up.

roles/backup/tasks/postgres.yml
rooftopcellist: We need to query the k8s cluster for the name of the running awx pod to set that variable.  

Here is a good example of that:
* https://github.com/ansible/awx-operator/blob/devel/roles/installer/tasks/resources_configuration.yml#L2-L25

roles/backup/tasks/postgres.yml
rooftopcellist: Right, you'll need to swap this out for:

```suggestion
    container: "{{ deployment_name }}-task"
```

rooftopcellist (dismissed):

config/manifests/bases/awx-operator.clusterserviceversion.yaml
```diff
@@ -50,6 +50,11 @@ spec:
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:text
         - urn:alm:descriptor:com.tectonic.ui:advanced
+      - displayName: Precreate Partition Hours
+        path: precreate_partition_hours
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:text
+        - urn:alm:descriptor:com.tectonic.ui:advanced
```
TheRealHaoLiu, line 57:
```suggestion
        - urn:alm:descriptor:com.tectonic.ui:text
        - urn:alm:descriptor:com.tectonic.ui:advanced
        - urn:alm:descriptor:com.tectonic.ui:hidden
```

TheRealHaoLiu:
otherwise the PR looks good, have you tested the PR? if so can you describe the step that you have take to test this PR

jessicamack:
@TheRealHaoLiu 
Test path:
1. Create k8s cluster
2. Build awx-operator
3. Run backup playbook. Confirm no errors during the backup
4. Exec into the database container and confirm the presence of partitions

TheRealHaoLiu (approved):

roles/backup/tasks/postgres.yml
```diff
@@ -82,6 +82,41 @@
...
+- name: Set the resource pod name as a variable.
+  set_fact:
+    awx_task_pod_name: "{{ awx_task_pod['metadata']['name'] | default('') }}"
+
+- name: Precreate database partitions
+  k8s_exec:
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    pod: "{{ awx_task_pod_name }}"
+    container: "{{ deployment_name }}-task"
+    command: awx-manage precreate_partitions --count='{{ precreate_partition_hours }}'
```
rooftopcellist, line 115: Can this manage command be run again and again?  Are there any downsides to it being run more than once?

jessicamack: @rooftopcellist it can be run repeatedly. if a partition already exists it will be a no-op https://github.com/ansible/awx/blob/devel/awx/main/utils/common.py#L1151-L1184

rooftopcellist (approved):

roles/backup/tasks/postgres.yml
mimianddaniel: running postgres externally, I am assuming running awx-manage precreate_partition as k8 job would perform same effect on the external postgres?

PR #1445, opened 2023-06-08
Add label for awx-task deployment.

laiminhtrung1997:


fosterseth:
Hi, thanks for the PR. Can you update the PR body explaining the use case for this change?

in particular, why is a label needed, since the deployment name is unique within the cluster and filtering on deployment name should be straightforward?

btw we do have a pod label for the task pods

rooftopcellist:
These are the current labels:

```
  labels:
    app.kubernetes.io/name: awx-task
    app.kubernetes.io/operator-version: ""
    app.kubernetes.io/part-of: awx
    app.kubernetes.io/version: latest
    pod-template-hash: 79f597b9dd
```

As @fosterseth mentioned, you could either use the name, or filter for the following label:

```
    app.kubernetes.io/name: awx-task

```

> Note: I just happened to name my deployment "awx" in this case"

laiminhtrung1997:
Dear @fosterseth, @rooftopcellist 
If we get awx-task deployment with labels selector app.kubernetes.io/name=awx-task, there is 0 deployment.
eg: kubectl -n <namespace> get deployemnt -l app.kubernetes.io/name=awx-task
But if we get awx-web deployment with labels selector app.kubernetes.io/name=awx-web, there is 1 deployment.
eg: kubectl -n <namespace> get deployemnt -l app.kubernetes.io/name=awx-web
Because in web.yaml, you implemented this label app.kubernetes.io/name=awx-web, but awx-task was not.
https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/deployments/web.yaml.j2#L8
So, I just want both deployments has implemented the same and we can get awx-task deployment with that label app.kubernetes.io/name=awx-task instead of pods.

PR #1446, opened 2023-06-08
Implement preStop for awx-web deployment.

laiminhtrung1997:
##### SUMMARY
awx-web deployment is missing preStop. I think we should add this feature to awx-web as awx-task.

##### ISSUE TYPE
 - New or Enhanced Feature

### DIFF
--- a/roles/installer/templates/deployments/web.yaml.j2
+++ b/roles/installer/templates/deployments/web.yaml.j2
@@ -142,6 +142,23 @@ spec:
               mountPath: "/var/run/redis"
             - name: "{{ ansible_operator_meta.name }}-redis-data"
               mountPath: "/data"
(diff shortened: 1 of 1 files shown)


fosterseth:
Thank you for your contribution!

the preStop is really only useful for the task container, as we want to gracefully wait for jobs to finish before killing the task-related containers. 

For web containers, the preStop doesn't really offer any utility, as web requests can still be directed to the other web containers that are up.

Is there any particular reason why preStop would be needed for web containers?

rooftopcellist:
This preStop logic is only necessary for the task deployment as all of the jobs are running out of the containers in the task deployment.  So this change is unnecessary.  I'm closing this for now, please re-open or comment if this is not the case.

PR #1458, opened 2023-06-22
Do not cause web deployment restart on pre-stop scripts changes

stanislav-zaprudskiy:
##### SUMMARY

This is a small change to remove unrelated checksum annotation from `web` deployment, as this deployment is not utilizing the corresponding referenced ConfigMap and thus should not be [rolled out (restarted)](https://github.com/ansible/awx-operator/commit/94d68bf382ec8dc4ce28d7d8d154663afd00b7fe) whenever that changes.

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

The similar  annotation changes will have to be provided in future, should it be ever decided that e.g. receptor certs/configs or projects PVC are no longer needed (both for `web` and `task` deployments). Likewise, when new mounts are added the annotations should be extended.

### DIFF
--- a/roles/installer/templates/deployments/web.yaml.j2
+++ b/roles/installer/templates/deployments/web.yaml.j2
@@ -28,7 +28,6 @@ spec:
       annotations:
 {% for template in [
     "configmaps/config",
-    "configmaps/pre_stop_scripts",
     "secrets/app_credentials",
     "storage/persistent",
   ] %}


rooftopcellist (approved):
Thanks! Good catch.

TheRealHaoLiu:
Thanks @stanislav-zaprudskiy ~

Issue #1451, opened 2023-06-16
Allow ability to set securityContext in Postgres StatefulSet.


AlbinKroon:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that AWX Operator is open source software provided for free and that I might not receive a timely response.

### Feature Summary

As of now it is not possible to set/change the securityContext of the Postgres StatefulSet. This means the pod won't start when having configured the Pod Security Admission Controller to enforce restricted mode in all namespaces.

<img width="1056" alt="Screenshot 2023-06-16 at 16 27 09" src="https://github.com/ansible/awx-operator/assets/33911033/27c0c66b-9c08-4100-bd7a-5683697ba869">

Please consider adding the ability to set the securityContext in `roles/installer/templates/statefulsets/postgres.yaml.j2` under `.spec.template.spec`.

rooftopcellist:
This has been address by:
* https://github.com/ansible/awx-operator/pull/1517

PR #1461, opened 2023-06-23
Allow ability to set security context for postgres deployment.

rakesh561:
##### SUMMARY
Allow ability to set security context for postgres containers during deployment.
Example it will allow users to deploy postgres as Non Root.

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
Issue #1451  will be addressed hopefully

### DIFF
--- a/roles/installer/templates/statefulsets/postgres.yaml.j2
+++ b/roles/installer/templates/statefulsets/postgres.yaml.j2
@@ -51,6 +51,12 @@ spec:
         - image: '{{ _postgres_image }}'
           imagePullPolicy: '{{ image_pull_policy }}'
           name: postgres
+{% if security_context_settings.postgres is defined %}
+          securityContext:
+{% for key, value in security_context_settings.postgres.items() %}
+            {{ key }}: {{ value }}
+{% endfor %}
+{% endif %}
 {% if postgres_extra_args %}
           args: {{ postgres_extra_args }}
 {% endif %}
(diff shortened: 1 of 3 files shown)


rooftopcellist:
Before this can merge, we need to have entries for the security setting param for postgres on the CRD and in the CSV params.

rakesh561:
@rooftopcellist  for the CRD i believe there is entry for CSV already.

roles/installer/templates/statefulsets/postgres.yaml.j2
```diff
@@ -51,6 +51,12 @@ spec:
         - image: '{{ _postgres_image }}'
           imagePullPolicy: '{{ image_pull_policy }}'
           name: postgres
+{% if security_context_settings.postgres is defined %}
```
rooftopcellist: The security_context_settings.postgres key doesn't exist in the AWX schema.  All I see is `security_context_settings`  [here](https://github.com/ansible/awx-operator/blob/devel/config/crd/bases/awx.ansible.com_awxs.yaml#L1791). 


If we wanted to have specific security context settings for postgres (different from task/web, etc.), then we would probably want to define it under `postgres.security_context_setting`.  We don't have a precedent for this in the awx-operator yet, but there are examples in the eda-server-operator [here](https://github.com/ansible/eda-server-operator/blob/main/config/crd/bases/eda.ansible.com_edas.yaml#L1248-L1254).

For this PR, it might make more sense to just use the top level security_context_settings value everywhere (including the postgres sts with this PR).  What do you think? Would that solve your use case?

rakesh561: @rooftopcellist  i did the requested changes i went with having seperate security_context_settings for postgres instead using the same security_context_settings that is already declared.

PR #1468, opened 2023-06-26
Allow {web_,task_,}replicas to be 0 and split out molecule tests

### LINKED ISSUE #1454: AWX doesn't scale in to 0 replicas

- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.
- [X] I am **NOT** reporting a (potential) security vulnerability. (These should be emailed to `security@ansible.com` instead.)

### Bug Summary

After the task and web pods separation, AWX can no longer scale in to 0 replicas when we edit the AWX CRD.

We've installed the AWX with three replicas by setting the following parameters in the CRD object.
```
replicas: 3
task_replicas: 3
web_replicas: 3
```
As far as I understand, **web_replicas** and **task_replicas** should override **replicas** in this case.

I am trying to scale it in to 0 by settings:

```
replicas: 0
task_replicas: 0
web_replicas: 0
```
But AWX scales in only one replicas (so I have one web and one task replica) and doesn't completely scale in to 0 replicas.

I can't find anything interesting in the awx-operator logs.

### AWX version

22.3.0

### Select the relevant components

- [ ] UI
- [ ] UI (tech preview)
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

Try to change CRD to 0 replicas from 3 replicas.

### Expected results

AWX is scaled in to 0 replicas

### Actual results

AWX is scaled in to 1 replica

relrod:
##### SUMMARY

(Replaces #1464 because I think there is a :ghost: CI check there that is in a weird state.)

There is quite a bit going on here.

On the user-facing side:

- First off allow `replicas`, `web_replicas`, and `task_replicas` to be `0`. Previously this was interpreted as a falsy value and messed up the templating logic. Now we compare it to the default (empty string) instead. Fixes #1454.
- The above uncovered another issue: If we scale replicas to 0, then the operator reconciliation loop starts failing since it tries to find some pod that doesn't exist in that case. Only run those tasks if we are able.

On the testing side:

- Add tests for the above.
- However this made the (single) test file really big, so I broke it into some separately-runnable files. For example, one file to update the `AWX` instance we are testing with. You just import that with `include_tasks`.
  - I also made it so that when you do that, you can pass in an `additional_fields` var so that you can actually customize the `AWX` definition to test how changes to fields interact.
- Lowered CPU request specs, because it was failing with testing 3 replicas on GHA.
- All these new tests made CI really slow (~13 minutes up to ~25 minutes). So to fix this, I made use of ansible tags and broke a test matrix in GHA. It's also easy to expand this (just tag the new tasks and edit/add two lines in `ci.yaml`) so that if we add chunks of other tests later, they can each run in their own separate instance, in parallel.
  - To do that, I had to tag the tasks that should always run (like the tasks to set up kind and tear it down, and to do the image build). I think this is a fair trade-off for the flexibility we get with doing matrix runs now.
- Also deleted the branch filter from `ci.yaml` so that people can run the tests in the own forks on their own feature branches.

##### ISSUE TYPE

### DIFF
--- a/roles/installer/tasks/install.yml
+++ b/roles/installer/tasks/install.yml
@@ -88,6 +88,7 @@
     command: >-
       bash -c "awx-manage showmigrations | grep -v '[X]' | grep '[ ]' | wc -l"
   changed_when: false
+  when: awx_task_pod_name != ''
   register: database_check
 
 - name: Migrate the database if the K8s resources were updated.  # noqa 305
@@ -99,11 +100,13 @@
       bash -c "awx-manage migrate --noinput"
   register: migrate_result
   when:
+  - awx_task_pod_name != ''
   - database_check is defined
   - (database_check.stdout|trim) != '0'
 
 - name: Initialize Django
   include_tasks: initialize_django.yml
+  when: awx_task_pod_name != ''
 
 - name: Update status variables
   include_tasks: update_status.yml
--- a/roles/installer/tasks/resources_configuration.yml
+++ b/roles/installer/tasks/resources_configuration.yml
@@ -285,5 +285,7 @@
 
 - name: Verify the resource pod name is populated.
   assert:
-    that: awx_task_pod_name != ''
+    that:
+      - awx_task_pod_name != ''
     fail_msg: "Could not find the tower pod's name."
+  when: task_replicas | int > 0 or (task_replicas == '' and replicas > 0)
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,6 +10,11 @@ jobs:
   molecule:
     runs-on: ubuntu-latest
     name: molecule
+    strategy:
+      matrix:
+        ansible_args:
+          - --skip-tags=replicas
+          - -t replicas
     env:
       DOCKER_API_VERSION: "1.41"
     steps:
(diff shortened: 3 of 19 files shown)


roles/installer/tasks/resources_configuration.yml
```diff
@@ -285,5 +285,7 @@
 
 - name: Verify the resource pod name is populated.
   assert:
-    that: awx_task_pod_name != ''
+    that:
+      - awx_task_pod_name != ''
     fail_msg: "Could not find the tower pod's name."
```
thedoubl3j, line 290: Do we want to think about moving away from the tower naming scheme in errors since we changed the pod_names? I will leave this as a single comment since it isn't directly applied to the PR, more just for others to see and think about potentially changing it if there are others.

thedoubl3j (approved):
big plus 1 for the test changes around tags to help with the slowness and fixes for `replicas`

roles/installer/tasks/resources_configuration.yml
rooftopcellist: yes, but it doesn't have to prevent this from merging.

roles/installer/tasks/install.yml
```diff
@@ -88,6 +88,7 @@
     command: >-
       bash -c "awx-manage showmigrations | grep -v '[X]' | grep '[ ]' | wc -l"
   changed_when: false
+  when: awx_task_pod_name != ''
```
rooftopcellist, line 91: Good catch, thanks for adding this.

rooftopcellist:
I like the way you split out the replicas check to `_test_case_replicas.yml`.  Nicely done all around.

rooftopcellist (approved):

PR #1476, opened 2023-06-30
There was a typo in the CRD for the ephemeral storage which was ephemeral_storage instead of ephemeral-storage

rakesh561:
##### SUMMARY
There was a typo in the CRD for the ephemeral storage which was ephemeral_storage instead of ephemeral-storage.Will Appreciated if i can get quick review sorry for the typo in the earlier PR.

##### ISSUE TYPE


rooftopcellist (approved):

rooftopcellist:
Thanks for the fix @rakesh561 !

PR #1494, opened 2023-07-17
Molecule: Different fix for new Cython

relrod:
##### SUMMARY

The first attempt (#1492) was unfortunately wrong in two ways. We need to pin the dependency in inside the container, not as a molecule dependency. Also it seems like the Cython pin needs to happen *earlier* but an easy solution is to just pin pyyaml to an old version instead for now until https://github.com/yaml/pyyaml/issues/601 is resolved.

Here's an AWX CI run showing this fix works, this time: https://github.com/ansible/awx/pull/14251

I will revert #1493 as well, it is useless.

##### ISSUE TYPE

### DIFF
--- a/molecule/kind/converge.yml
+++ b/molecule/kind/converge.yml
@@ -15,6 +15,7 @@
           - urllib3==1.26.15
           - docker
           - docker-compose
+          - pyyaml==5.3.1
         state: present
 
     - name: Build operator image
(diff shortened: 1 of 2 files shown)


relrod:
Backing this to draft for now, I want to see if #1463 fixes the issue instead.

relrod:
Closing, I prefer #1463 to fix this.

PR #1505, opened 2023-07-28
Attempt at fixing #1418 for the task deployment

phess:
##### ISSUE TYPE

##### SUMMARY
Issue #1418 has fixed web deployments but it appears that task deployments are still suffering from the same issue.
Here I'm attempting a similar change to `task.yml.j2` that was done to `web.yml.j2` in order to check if this fixes the issue for task deployments like it did web deployments.

I'm not naming this PR "fixes #1418" because I'm clearly _trying_.

If this description is not clear enough, comments at AAP-13578 should help clarify.
If it's all unclear and incorrect, please reject or ignore this PR mercilessly.

### DIFF
--- a/roles/installer/templates/deployments/task.yaml.j2
+++ b/roles/installer/templates/deployments/task.yaml.j2
@@ -353,7 +353,6 @@ spec:
           args: {{ rsyslog_args }}
 {% endif %}
           imagePullPolicy: '{{ image_pull_policy }}'
-          resources: {{ rsyslog_resource_requirements }}
           volumeMounts:
             - name: "{{ ansible_operator_meta.name }}-application-credentials"
               mountPath: "/etc/tower/conf.d/credentials.py"
@@ -408,6 +407,7 @@ spec:
 {% if rsyslog_extra_env -%}
             {{ rsyslog_extra_env | indent(width=12, first=True) }}
 {% endif %}
+          resources: {{ rsyslog_resource_requirements }}
 {% if task_node_selector %}
       nodeSelector:
         {{ task_node_selector | indent(width=8) }}


rooftopcellist:
I have been trying to reproduce your issue with the latest in the `devel` branch of the awx-operator. I tried this in my spec and was unable to reproduce.  The task pod always gets scheduled on a node.  I also inspected the deployment yaml and found that this was getting templated correctly. 

```
spec:
  task_node_selector: |
    kubernetes.io/os: linux
 ```


I also just tried the following:

```
spec:
  node_selector: |
    kubernetes.io/os: linux  
  task_node_selector: |
    kubernetes.io/os: linux 
```

Could you share a redacted CR spec? Maybe I can spot the issue or use that to try to reproduce?

**Comment by rooftopcellist** [comment]
@phess It looks like rsyslog_resource_requirements is in the right spot in the task deployment template as is:
* https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/deployments/task.yaml.j2#L356

I think what you are seeing may be an unrelated issue.

**Comment by phess** [comment]
> Could you share a redacted CR spec? Maybe I can spot the issue or use that to try to reproduce?

Thank you. I'll see what we can get.

**Comment by davidtluong** [comment]
I reproduced it, using a toleration breaks it.

**Comment by rooftopcellist** [comment]
@phess It looks like you are using a version of the operator that does not include this patch:
* https://github.com/ansible/awx-operator/pull/1349

I was able to reproduce on an older version of the awx-operator and that patch fixed it.  This is the spec I used to reproduce:

```
spec:
  no_log: false
  task_node_selector: |
    kubernetes.io/os: linux
  task_tolerations: |
    - key: "dedicated"
  tolerations: |
    - key: "dedicated"
  service_type: clusterip
  ingress_type: route

  task_resource_requirements:
    requests:
      cpu: 100m
      memory: 128Mi  
  rsyslog_resource_requirements:
    requests:
      cpu: 100m
      memory: 128Mi 
```

Taking out the tolerations lines makes it pass, which gives me extra confidence that this is the issue in your environment.  Thank you for putting up a patch, I am going to close this now though since it appears to be fixed in `devel`.  Please re-open or comment if that was a mistake.

PR #1517, opened 2023-08-10
Allow ability to set security context for postgres deployment

rakesh561:
##### SUMMARY
Allow ability to set security context for postgres containers during deployment.
Example it will allow users to deploy postgres as Non Root.

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
Issue #1451  will be addressed hopefully

### DIFF
--- a/config/manifests/bases/awx-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/awx-operator.clusterserviceversion.yaml
@@ -62,7 +62,7 @@ spec:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:com.tectonic.ui:hidden
       - displayName: PostgreSQL Security Context Settings
-        path: postgres.security_context_settings
+        path: postgres_security_context_settings
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:com.tectonic.ui:hidden      
--- a/config/manifests/bases/awx-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/awx-operator.clusterserviceversion.yaml
@@ -61,6 +61,11 @@ spec:
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:com.tectonic.ui:hidden
+      - displayName: PostgreSQL Security Context Settings
+        path: postgres.security_context_settings
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:advanced
+        - urn:alm:descriptor:com.tectonic.ui:hidden      
       - displayName: PostgreSQL Image
         path: postgres_image
         x-descriptors:
--- a/roles/installer/templates/statefulsets/postgres.yaml.j2
+++ b/roles/installer/templates/statefulsets/postgres.yaml.j2
@@ -53,7 +53,7 @@ spec:
           name: postgres
 {% if postgres_security_context_settings|length %}
           securityContext:
-            {{ postgres_security_context_settings | to_nice_yaml | indent(8) }}
+            {{ postgres_security_context_settings | to_nice_yaml | indent(12) }}
 {% endif %}
 {% if postgres_extra_args %}
           args: {{ postgres_extra_args }}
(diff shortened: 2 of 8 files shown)


rakesh561:
@rooftopcellist  i have opened this as new pr for the postgres security context pr that i have opened earlier as i was running into some conflicts issue.Would appreciate if you can review the above PR when you have some time

AlbinKroon:
LGTM. Perhaps you could assign this PR to @rooftopcellist, if you were expecting him to review it?

rakesh561:
@rooftopcellist @TheRealHaoLiu @fosterseth  can one of you folks review the above PR.Appreciate in advance

fosterseth:
@rakesh561 we are hoping to go more this route for adding security contexts to the postgres deployment
https://github.com/ansible/awx-operator/pull/1425#issuecomment-1652315313

We'd like to see a PR that implements it this way. If you want to update this PR to reflect the feedback on that comment I linked, that would be fruitful

config/manifests/bases/awx-operator.clusterserviceversion.yaml
```diff
@@ -55,6 +55,11 @@ spec:
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:com.tectonic.ui:hidden
+      - displayName: PostgreSQL Security Context Settings
+        path: postgres.security_context_settings
```
rooftopcellist: The name you chose for this parameter is good because it fits with the current V1 AWX schema (`postgres_security_context_settings`), however the reference here will not resolve correctly.  

It should be:
```suggestion
        path: postgres_security_context_settings
```

roles/installer/templates/statefulsets/postgres.yaml.j2
```diff
@@ -51,6 +51,12 @@ spec:
         - image: '{{ _postgres_image }}'
           imagePullPolicy: '{{ image_pull_policy }}'
           name: postgres
+{% if postgres.security_context_settings is defined %}
+          securityContext:
+{% if postgres.security_context_settings|length %}
+      {{ postgres.security_context_settings | to_nice_yaml | indent(8) }}
```
rooftopcellist: Same here:

```suggestion
      {{ postgres_security_context_settings | to_nice_yaml | indent(8) }}
```

roles/installer/templates/statefulsets/postgres.yaml.j2
```diff
@@ -51,6 +51,12 @@ spec:
         - image: '{{ _postgres_image }}'
           imagePullPolicy: '{{ image_pull_policy }}'
           name: postgres
+{% if postgres.security_context_settings is defined %}
```
rooftopcellist:
```suggestion
{% if postgres_security_context_settings is defined %}
```

roles/installer/templates/statefulsets/postgres.yaml.j2
```diff
@@ -51,6 +51,12 @@ spec:
...
+          securityContext:
+{% if postgres.security_context_settings|length %}
```
rooftopcellist:
```suggestion
{% if postgres_security_context_settings|length %}
```

rooftopcellist:
@rakesh561 as it was, the PR didn't set the securityContext on the postgres StatefulSet, I pushed a couple changes to you branch and now it does:
* needed a default set in `roles/installer/defaults/main.yml`
* needed to check the length `{% if postgres_security_context_settings|length %}`, not if defined when templating. 
* removed unneeded conditional in the templating.  Without this change, `securityContext` would have been set even if the value passed by the user is empty.

I tested this out with the following in my awx CR spec:

```yaml
---
kind: AWX
metadata:
  name: awx
  namespace: awx
spec:
  postgres_security_context_settings:
    runAsNonRoot: true
```

And I see that the field has been added on the postgres StatefulSet:

```yaml
$ oc get sts awx-postgres-13  -o yaml | grep security -A 3

        securityContext:
          runAsNonRoot: true
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File

```

rooftopcellist (approved):

PR #1530, opened 2023-08-22
Add keepalive to migrate data script

TheRealHaoLiu:
##### SUMMARY

pg_restore is silent and if the data migration takes longer than 15 minute exec/rsh to the postgres pod (through which we run the data migration command) will idle timeout

this PR adds a background process that prints keepalive messages during the migration process to keep the exec/rsh alive

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/roles/installer/tasks/migrate_data.yml
+++ b/roles/installer/tasks/migrate_data.yml
@@ -59,11 +59,18 @@
     namespace: "{{ ansible_operator_meta.namespace }}"
     pod: "{{ postgres_pod_name }}"
     command: |
-      bash -c """
+      bash -c "
+      keepalive_file=\"$(mktemp)\"
+      while [[ -f \"$keepalive_file\" ]]; do echo 'Migrating data from old database...'; sleep 1; done & keepalive_pid=$!
+      echo keepalive_pid: $keepalive_pid
+      trap \"rm -f $keepalive_file; pkill -P $keepalive_pid; wait $keepalive_pid\" EXIT INT TERM
       set -e -o pipefail
       PGPASSWORD='{{ awx_old_postgres_pass }}' {{ pgdump }} | PGPASSWORD='{{ awx_postgres_pass }}' {{ pg_restore }}
-      echo 'Successful'
-      """
+      rm -f \"$keepalive_file\"
+      pkill -P $keepalive_pid || true
+      wait $keepalive_pid
+      echo done
+      "
   no_log: "{{ no_log }}"
   register: data_migration
   failed_when: "'Successful' not in data_migration.stdout"


roles/installer/tasks/migrate_data.yml
```diff
@@ -60,9 +60,12 @@
     pod: "{{ postgres_pod_name }}"
     command: |
       bash -c """
+      (while true; do echo 'Migrating data from old database...'; sleep 60; done) & keep_alive_pid=$!
       set -e -o pipefail
       PGPASSWORD='{{ awx_old_postgres_pass }}' {{ pgdump }} | PGPASSWORD='{{ awx_postgres_pass }}' {{ pg_restore }}
-      echo 'Successful'
+      kill $keep_alive_pid;
```
rooftopcellist: Please remove these semicolons on line 66 and ln 68.  I tested without the semicolons with the following scenarios:
* migrate from AWX A --> AWX B
* migrate from AWX A with a beefy db as well as a sleep 1800 --> AWX C (took 37 min.)
cc @TheRealHaoLiu

roles/installer/tasks/migrate_data.yml
```diff
@@ -59,11 +59,18 @@
...
+      keepalive_file=\"$(mktemp)\"
+      while [[ -f \"$keepalive_file\" ]]; do echo 'Migrating data from old database...'; sleep 1; done & keepalive_pid=$!
+      echo keepalive_pid: $keepalive_pid
+      trap \"rm -f $keepalive_file; pkill -P $keepalive_pid; wait $keepalive_pid\" EXIT INT TERM
       set -e -o pipefail
       PGPASSWORD='{{ awx_old_postgres_pass }}' {{ pgdump }} | PGPASSWORD='{{ awx_postgres_pass }}' {{ pg_restore }}
-      echo 'Successful'
-      """
+      rm -f \"$keepalive_file\"
+      pkill -P $keepalive_pid || true
```
rooftopcellist, line 70: `pkill `is not available in the postgresql image sadly, nor is `ps`, else we could do something like:


```
ps -o pid= --ppid $keepalive_pid | xargs kill -TERM
```

We may be forced to do something like these, or leave behind defunct processes:

```
for pid in /proc/*/status; do
  if grep -q '^PPid:\s*$keepalive_pid$' "$pid"; then
    kill -TERM "${pid#/proc/}"
    kill -TERM "${pid#/proc/*/}"
  fi
done

```

rooftopcellist: We realized that the new sclorg postgres 15 image has `ps` and `pkill`, so as long as we merge that first, we will be good.  

@relrod and I tested this out by putting this commit on top of the pg15 changes.  It failed for a number of reasons:
* the return_code was 1 for the `wait`
* the echoed message needed to be "Successful" for the `failed_when` conditional to be met.

These are the changes need:

```diff
--- a/roles/installer/tasks/migrate_data.yml
+++ b/roles/installer/tasks/migrate_data.yml
@@ -63,13 +63,12 @@
       keepalive_file=\"$(mktemp)\"
       while [[ -f \"$keepalive_file\" ]]; do echo 'Migrating data from old database...'; sleep 1; done & keepalive_pid=$!
       echo keepalive_pid: $keepalive_pid
-      trap \"rm -f $keepalive_file; pkill -P $keepalive_pid; wait $keepalive_pid\" EXIT INT TERM
       set -e -o pipefail
       PGPASSWORD='{{ awx_old_postgres_pass }}' {{ pgdump }} | PGPASSWORD='{{ awx_postgres_pass }}' {{ pg_restore }}
       rm -f \"$keepalive_file\"
-      pkill -P $keepalive_pid || true
-      wait $keepalive_pid
-      echo done
+      pkill -P $keepalive_pid > /dev/null || true
+      wait $keepalive_pid || true
+      echo 'Successful'
       "
```

Thanks to a monumental effort from @relrod , I think we finally have the change set needed to wrap up this issue.  @TheRealHaoLiu can you please make these changes when you get a chance Mon morning?

Sidenote, we also confirmed that `ps` and `pkill` are both available in the following image as well.
* registry.redhat.io/rhel8/postgresql-15:1

relrod: Possible solution without pkill/ps.

cc @rooftopcellist - can we test with this?

```bash
        command: |
          bash -c "
          function end_keepalive {
            rm -f \"$1\"
            kill $(cat /proc/$2/task/$2/children 2>/dev/null) 2>/dev/null || true
            wait $2 || true
          }
          keepalive_file=\"$(mktemp)\"
          while [[ -f \"$keepalive_file\" ]]; do
            echo 'Migrating data from old database...'
            sleep 1
          done &
          keepalive_pid=$!
          trap 'end_keepalive \"$keepalive_file\" \"$keepalive_pid\"' EXIT SIGINT SIGTERM
          echo keepalive_pid: $keepalive_pid
          set -e -o pipefail
          PGPASSWORD='{{ awx_old_postgres_pass }}' {{ pgdump }} | PGPASSWORD='{{ awx_postgres_pass }}' {{ pg_restore }}
          set +e +o pipefail
          echo 'Successful'
          "
```

relrod: Also:

- Would it be better to set the password as an env var on the container's definition or similar?
  - We could use [this syntax](https://stackoverflow.com/a/31599308) in the command definition tasks above rather than specifying the password here...
  - Either way: What happens if the postgresql password has a `'` in it?

rooftopcellist: The reason why we used the env var in favor of the syntax you linked is because that syntax would show the pg password in the logs if no_log was set to false.  

I don't have a good answer for the `'`... I expect that would cause problems.  We could add a task that checks the password and throws an error "Invalid Password" if one with quotes in the middle were passed.  

I will test out the solution you provided above.

rooftopcellist: Can confirm, the migration succeeds with those changes, and I see the data in the new database:

```
awx=# select username from auth_user;
     username     
------------------
 pre-upgrade-user
 admin
(2 rows)
```

rooftopcellist:
replaced by https://github.com/ansible/awx-operator/pull/1538

PR #1531, opened 2023-08-23
Switch to using the sclorg redis image for consistency

rooftopcellist:
##### SUMMARY

The eda-server-operator uses the sclorg redis image as well, which is closer to what we use downstream.  Consistency will improve maintainability.  

New image:
* quay.io/sclorg/redis-6-c8s:latest

##### ISSUE TYPE

### DIFF
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -237,8 +237,8 @@ extra_volumes: ''
 
 _image: quay.io/ansible/awx
 _image_version: "{{ lookup('env', 'DEFAULT_AWX_VERSION') or 'latest' }}"
-_redis_image: docker.io/redis
-_redis_image_version: 7
+_redis_image: quay.io/sclorg/redis-6-c9s
+_redis_image_version: latest
 _postgres_image: postgres
 _postgres_image_version: 13
 image_pull_policy: IfNotPresent
--- a/docs/user-guide/advanced-configuration/deploying-a-specific-version-of-awx.md
+++ b/docs/user-guide/advanced-configuration/deploying-a-specific-version-of-awx.md
@@ -9,7 +9,7 @@ There are a few variables that are customizable for awx the image management.
 | image_pull_policy   | The pull policy to adopt  | IfNotPresent                            |
 | image_pull_secrets  | The pull secrets to use   | None                                    |
 | ee_images           | A list of EEs to register | quay.io/ansible/awx-ee:latest           |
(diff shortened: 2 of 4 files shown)


roles/installer/defaults/main.yml
```diff
@@ -237,8 +237,8 @@ extra_volumes: ''
 
 _image: quay.io/ansible/awx
 _image_version: "{{ lookup('env', 'DEFAULT_AWX_VERSION') or 'latest' }}"
-_redis_image: docker.io/redis
-_redis_image_version: 7
+_redis_image: quay.io/sclorg/redis-6-c8s
```
```suggestion
_redis_image: quay.io/sclorg/redis-6-c9s
```
Since the postgres 15 PR will use a c9s based container image then we should have the same thing everywhere.
What do you think ?

fosterseth (approved):

Denney-tech:
This requires a change to the redis.conf file to specify `dir /var/lib/redis/data` due to how sclorg/redis is built vs the docker ones.

Completing pr #1427 would allow users to define redis.conf with or without the `dir` line as needed.

roles/installer/defaults/main.yml
rooftopcellist: Agreed, I just changed this to c9s.  I'll make the same changes for eda-server-operator.

rooftopcellist:
Good point @Denney-tech .  The path is changed in this PR, so it will work out of the box.  But, I suppose there may be users who want to use redis:7 from docker, so https://github.com/ansible/awx-operator/pull/1427 would allow them to set a custom path for the redis data dir (in this case /var/lib/redis/data instead of /data).  

That other PR is largely ready to go, it just needs the merge conflicts to be resolved by the contributor.  If someone has time to take over that PR and finish it up, that would be welcome.

Denney-tech:
Sorry, I must have been tired yesterday when I was looking at this.  Had spent all day figuring out why the sclorg images didn't work, figured it out, then found these two PR's. For some reason I didn't see the `dir` line added among the files changed here.  This PR satisfies my immediate need, though.

Thank you.

Denney-tech:
Actually, no I wasn't crazy, the /etc/redis.conf has not been modified by this pull request and the devel image from quay now deploys the sclorg/redis image without the proper config line.

[roles/installer/templates/configmaps/config.yaml.j2 ](https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/configmaps/config.yaml.j2#L236) needs `dir /var/lib/redis/data` appended still.

PR #1534, opened 2023-08-25
Changed replica from string to int for default

davidtluong:
##### SUMMARY

Change var for replicas from string to int

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

If spec is absent for `replicas` for the awx CR, upgrades will fail like so:

```
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'task_replicas | int > 0 or (task_replicas == '' and replicas > 0)' failed. The error was: Unexpected templating type error occurred on ({% if task_replicas | int > 0 or (task_replicas == '' and replicas > 0) %} True {% else %} False {% endif %}): '>' not supported between instances of 'AnsibleUnicode' and 'int'\n\nThe error appears to be in '/opt/ansible/roles/installer/tasks/resources_configuration.yml': line 286, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Verify the resource pod name is populated.\n  ^ here\n"}
```

### DIFF
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -273,7 +273,7 @@ _init_projects_container_image: quay.io/centos/centos:stream9
 
 create_preload_data: true
 
-replicas: "1"
+replicas: 1
 web_replicas: ''
 task_replicas: ''
 


rooftopcellist (approved):

rooftopcellist:
Thanks @rh-dluong !

PR #1535, opened 2023-08-28
Add keepalive to migrate data script with restart of postgres pod [DNM]

TheRealHaoLiu:
##### SUMMARY
pg_restore is silent and if the data migration takes longer than 15 minute exec/rsh to the postgres pod (through which we run the command) will idle timeout

this PR adds a background process that prints keepalive messages during the migration process to keep the exec/rsh alive

This PR also added a restart to the postgres pod after the restore finish (successful or failed) because the new code result in defunct processes and restarting the pod take care of that

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/roles/installer/tasks/migrate_data.yml
+++ b/roles/installer/tasks/migrate_data.yml
@@ -59,14 +59,49 @@
     namespace: "{{ ansible_operator_meta.namespace }}"
     pod: "{{ postgres_pod_name }}"
     command: |
-      bash -c """
+      bash -c "
+      (while true; do echo 'Migrating data from old database...'; sleep 60; done) & keep_alive_pid=$!
       set -e -o pipefail
       PGPASSWORD='{{ awx_old_postgres_pass }}' {{ pgdump }} | PGPASSWORD='{{ awx_postgres_pass }}' {{ pg_restore }}
+      echo 'Killing keep alive process...'
+      kill $keep_alive_pid || true
+      wait
       echo 'Successful'
-      """
+      "
   no_log: "{{ no_log }}"
   register: data_migration
-  failed_when: "'Successful' not in data_migration.stdout"
+
+# Previous step result in defunct processes in the Postgres pod that need to be cleaned up
+# Restarting the Postgres pod will clean up the defunct processes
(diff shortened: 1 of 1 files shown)


roles/installer/tasks/migrate_data.yml
```diff
@@ -59,14 +59,49 @@
...
+      - status.phase=Running
+  register: postgres_pod
+  until:
+    - "postgres_pod['resources'] | length"
+    - "postgres_pod['resources'][0]['status']['phase'] == 'Running'"
+    - "postgres_pod['resources'][0]['status']['containerStatuses'][0]['ready'] == true"
+  delay: 5
+  retries: 60
+
+- name: Fail when data migration is not successful
```
relrod, line 101: Why this instead of the `failed_when`? And why not `'Successful' not in data_migration.stdout`?

roles/installer/tasks/migrate_data.yml
```diff
@@ -59,14 +59,49 @@
...
+      wait
       echo 'Successful'
-      """
+      "
   no_log: "{{ no_log }}"
   register: data_migration
-  failed_when: "'Successful' not in data_migration.stdout"
+
+# Previous step result in defunct processes in the Postgres pod that need to be cleaned up
+# Restarting the Postgres pod will clean up the defunct processes
```
relrod, line 75: This seems like it is inviting db inconsistencies if it gets killed at the wrong moment. If you insist on doing it this way, then I think it's important to make sure we gracefully shut down postgres before we kill the pod.

TheRealHaoLiu: awx deployments are scaled down before the data migration started 

there shouldn't be anything that's doing anything against postgres at this point so i think it would be safe to just kill the pod?

roles/installer/tasks/migrate_data.yml
TheRealHaoLiu: is there anything wrong with this approach?

and u know why (or what) wrote this code in this way ;-)

rooftopcellist (approved):

PR #1538, opened 2023-08-28
Add keepalive to migrate data script

rooftopcellist:
##### SUMMARY
This PR supercedes:
* https://github.com/ansible/awx-operator/pull/1535
* https://github.com/ansible/awx-operator/pull/1530

 - pg_restore is silent and if the data migration takes longer than 15 minute exec/rsh to the postgres pod (through which we run the command) will idle timeout
 - This PR adds a background process that prints keepalive messages during the migration process to keep the exec/rsh alive

    Co-Authored-By: Hao Liu <haoli@redhat.com>
    Co-Authored-By: Rick Elrod <rick@elrod.me>

##### ISSUE TYPE

### DIFF
--- a/roles/installer/tasks/migrate_data.yml
+++ b/roles/installer/tasks/migrate_data.yml
@@ -59,14 +59,29 @@
     namespace: "{{ ansible_operator_meta.namespace }}"
     pod: "{{ postgres_pod_name }}"
     command: |
-      bash -c """
+      bash -c "
+      function end_keepalive {
+        rc=$?
+        rm -f \"$1\"
+        kill $(cat /proc/$2/task/$2/children 2>/dev/null) 2>/dev/null || true
+        wait $2 || true
+        exit $rc
+      }
+      keepalive_file=\"$(mktemp)\"
+      while [[ -f \"$keepalive_file\" ]]; do
+        echo 'Migrating data from old database...'
+        sleep 60
+      done &
+      keepalive_pid=$!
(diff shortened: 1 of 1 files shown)


roles/installer/tasks/migrate_data.yml
```diff
@@ -59,14 +59,29 @@
...
+        rc=$?
+        rm -f \"$1\"
+        kill $(cat /proc/$2/task/$2/children 2>/dev/null) 2>/dev/null || true
+        wait $2 || true
+        exit $rc
+      }
+      keepalive_file=\"$(mktemp)\"
+      while [[ -f \"$keepalive_file\" ]]; do
+        echo 'Migrating data from old database...'
+        sleep 1
```
relrod:
```suggestion
        sleep 60
```

nit: I used `sleep 1` for testing. But this is going to be really noisy if the migration takes a while. Let's bump it up a bit.

relrod (approved):
LGTM :see_no_evil: :hear_no_evil: :speak_no_evil:

roles/installer/tasks/migrate_data.yml
rooftopcellist: Thanks!

rooftopcellist:
cc @TheRealHaoLiu let us know what you think of this approach.  Personally, I think this is the best out of the open PR's we have.

TheRealHaoLiu (approved):

TheRealHaoLiu:
LGTM

CFSNM (approved):

PR #1539, opened 2023-08-29
Modify how pg password is set in postgres pod

rooftopcellist:
##### SUMMARY

    Modify how PG password env vars are set to prevent issues with special characters
     - Certain characters like quotations would not be interpreted correctly
       if passed through bash. This avoids that issue.

##### ISSUE TYPE


rooftopcellist:
Closing in favor of https://github.com/ansible/awx-operator/pull/1540

PR #1541, opened 2023-08-29
Adding default resource limits for all containers

imedaouidene:
##### SUMMARY

Adding default resource limits to all containers. 
REASON: by default only resource requests are defined when creating an AWX instance, AWX pod won't be created created when ResourceQuota is present on the namespace when no direct logs from the operator. 

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

limit values can be revisited to increase/decrease them .

### DIFF
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -291,17 +291,25 @@ task_resource_requirements:
   requests:
     cpu: 100m
     memory: 128Mi
+  limits:
+    cpu: 2000m
+    memory: 4Gi
 
 web_resource_requirements:
   requests:
     cpu: 100m
     memory: 128Mi
+  limits:
+    cpu: 1000m
+    memory: 4Gi
 
 ee_resource_requirements:
   requests:
     cpu: 100m
     memory: 64Mi
-
+  limits:
+    cpu: 1000m
+    memory: 4Gi
 # TODO: validate default resource requirements
 
 # Customize CSRF options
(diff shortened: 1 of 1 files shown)


fosterseth:
@TheRealHaoLiu @rooftopcellist we probably would need these to be settable right?

fosterseth:
setting default limits could have some adverse effects to users to exceed them without knowing, so we probably don't want to set them.

@TheRealHaoLiu opinion on setting defaults?

another option to consider is doc section that has the limits defined that users can easily copy and paste into their spec file

somewhere here maybe https://github.com/ansible/awx-operator/blob/devel/docs/user-guide/advanced-configuration/containers-resource-requirements.md

or we can add a sample config that has the limits set for ease of use https://github.com/ansible/awx-operator/blob/devel/config/samples/

imedaouidene:
> setting default limits could have some adverse effects to users to exceed them without knowing, so we probably don't want to set them.
> 
> @TheRealHaoLiu opinion on setting defaults?
> 
> another option to consider is doc section that has the limits defined that users can easily copy and paste into their spec file
> 
> somewhere here maybe https://github.com/ansible/awx-operator/blob/devel/docs/user-guide/advanced-configuration/containers-resource-requirements.md
> 
> or we can add a sample config that has the limits set for ease of use https://github.com/ansible/awx-operator/blob/devel/config/samples/

@fosterseth  the problem is when some users are using namespaces with ResourceQuotas defined, when limits.cpu/limits.memory is defined in a Quota , all new coming pods must have these limits defined. I've had the same issue , and it was not easy to find out  that limits were not defined, when creating an AWX instance with default settings ( in  presence of a ResourceQuota ) , no pod will appear, and you would have to check operators logs to know that limits are not defined, which is not very straightforward.

TheRealHaoLiu:
Thank you for your contribution and pointing out the gotcha that people might run into when ResourceQuota is present. 

Resource limit various widely between environment and situation we are not able to determine something is "reasonable" for our of contributor/users so we decided to explicitly not include any limits by default.

If we introduce default limit now I can see existing "functioning" deployments stop functioning because of the new limit.

Perhaps we can put the information somewhere in documentation? 

Btw we have a Matrix chatroom if you want to have more live discussion with us. Join us on https://app.element.io/#/room/#awx:ansible.com! I can be found on there as TheRealHaoLiu

PR #1540, opened 2023-08-29
Modify how pg password is set in postgres pod

rooftopcellist:
##### SUMMARY
Alternate approach for https://github.com/ansible/awx-operator/pull/1539

##### ISSUE TYPE

 - Bug, Docs Fix or other nominal change

PR #1542, opened 2023-08-29
Timeout stream keep alive for Upgrades and Restores

rooftopcellist:
##### SUMMARY

This PR applies the changes made to the migration logic, in the upgrade and restore flows as well.  These are the changes that were made to the migration logic:
* https://github.com/ansible/awx-operator/pull/1538
* https://github.com/ansible/awx-operator/pull/1540

This addresses two bugs:
* Updates and restores involving excessively large databases were timing out at 900s.  The will not with these keep-alive messages.
* Previously, special characters like `'` and `"` in postgres passwords would have been problematic.  Now that is handled properly by setting the env var on the StatefulSet/Pod.  

##### ISSUE TYPE

### DIFF
--- a/roles/restore/tasks/postgres.yml
+++ b/roles/restore/tasks/postgres.yml
@@ -87,11 +87,27 @@
     namespace: "{{ backup_pvc_namespace }}"
     pod: "{{ ansible_operator_meta.name }}-db-management"
     command: |
-      bash -c """
+      bash -c "
+      function end_keepalive {
+        rc=$?
+        rm -f \"$1\"
+        kill $(cat /proc/$2/task/$2/children 2>/dev/null) 2>/dev/null || true
+        wait $2 || true
+        exit $rc
+      }
+      keepalive_file=\"$(mktemp)\"
+      while [[ -f \"$keepalive_file\" ]]; do
+        echo 'Migrating data from old database...'
+        sleep 60
+      done &
+      keepalive_pid=$!
+      trap 'end_keepalive \"$keepalive_file\" \"$keepalive_pid\"' EXIT SIGINT SIGTERM
+      echo keepalive_pid: $keepalive_pid
       set -e -o pipefail
       cat {{ backup_dir }}/tower.db | PGPASSWORD='{{ awx_postgres_pass }}' {{ pg_restore }}
+      set +e +o pipefail
       echo 'Successful'
-      """
+      "
   register: data_migration
   no_log: "{{ no_log }}"
   failed_when: "'Successful' not in data_migration.stdout"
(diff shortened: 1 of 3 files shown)


dsavineau (approved):

rooftopcellist:
# Backup and Restore testing

I created a backup, which ran successfully:

```
$ cat hacking/backup-awx.yml
---
kind: AWXBackup
metadata:
  name: awxbackup-2022-08-29
spec:
  deployment_name: awx
  no_log: false
```

Then after the reconciliation loop stopped for the backup, I created a restore:

```
$ cat hacking/restore-awx.yml 
---
kind: AWXRestore
metadata:
  name: restore1
  # namespace: ca-awx
spec:
  deployment_name: awx-2
  backup_name: awxbackup-2022-08-29
  no_log: false
```

The resulting awx-2 deployment has the migrated data in it.  

I found this bug along the way, but it is unrelated to these changes:
* https://github.com/ansible/awx-operator/issues/1544

roles/restore/tasks/postgres.yml
```diff
@@ -87,11 +87,27 @@
...
+      keepalive_file=\"$(mktemp)\"
+      while [[ -f \"$keepalive_file\" ]]; do
+        echo 'Migrating data from old database...'
+        sleep 60
+      done &
+      keepalive_pid=$!
+      trap 'end_keepalive \"$keepalive_file\" \"$keepalive_pid\"' EXIT SIGINT SIGTERM
+      echo keepalive_pid: $keepalive_pid
       set -e -o pipefail
       cat {{ backup_dir }}/tower.db | PGPASSWORD='{{ awx_postgres_pass }}' {{ pg_restore }}
```
kurokobo, line 107: @rooftopcellist 
Not tested yet, but is it okay to use `{{ awx_postgres_pass }}` here? It might include `'` or `"` and is still problematic imo.

kurokobo: And `"` in password is also still problematic since the `awx_postgres_pass` is also used to embed password into `credentials.py`: https://github.com/ansible/awx-operator/blob/4c5429190c9342a25b78af26b81366d218dd4df2/roles/installer/templates/settings/credentials.py.j2

rooftopcellist: These are good points.  I started to implement a fix for the restore role, but it got complicated and I ripped it out of this PR.  Those two spots still need changes before we can truly say that quotation marks in the pg password won't cause problems.

rooftopcellist: @kurokobo 

I created a separate issue to track this:
* https://github.com/ansible/awx-operator/issues/1545

The changes in this PR and in https://github.com/ansible/awx-operator/pull/1540 were fairly straight-forward, so I made them.  But fixing the other two spots might not be so simple, so I deferred them to future PR's.  We haven't gotten any bug reports from users yet on this, it's just something that @relrod and I noticed in the code the other day, so I am hesitant to make changes that would add more tasks to the code (and slow down the loop) if no one is actually running into this issue.

PR #1550, opened 2023-08-31
Bump ansible-operator to v1.31.0

ToxicGLaDOS:
##### SUMMARY

Bump ansible-operator version to  v1.31.0

This version includes a lot of security fixes for various yum packages.

##### ISSUE TYPE

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/operator-framework/ansible-operator:v1.28.1
+FROM quay.io/operator-framework/ansible-operator:v1.31.0
 
 USER 0
 
--- a/requirements.yml
+++ b/requirements.yml
@@ -3,4 +3,4 @@ collections:
   - name: kubernetes.core
     version: '>=2.3.2'
   - name: operator_sdk.util
-    version: "0.4.0"
+    version: "0.5.0"


TheRealHaoLiu:
@rooftopcellist can you take a look at this PR

TheRealHaoLiu (approved):

TheRealHaoLiu:
Thank you @ToxicGLaDOS 

btw are you on our Matrix or the Ansible Forum?

If you like to hang out I'm on https://app.element.io/#/room/#awx:ansible.com as TheRealHaoLiu, I would love to get to know you better!

TheRealHaoLiu:
with operator sdk 1.31.0 base image we see 
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ModuleNotFoundError: No module named 'openshift'
```

ToxicGLaDOS:
> with operator sdk 1.31.0 base image we see
> 
> ```
> An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ModuleNotFoundError: No module named 'openshift'
> ```

Hmm. I tried it out locally and it seems to work. Also I'm not sure how it wouldn't have openshift because we install openshift with in [the requirements.txt](https://github.com/ansible/awx-operator/blob/devel/molecule/requirements.txt#L5C18-L5C18).

Is [this](https://github.com/ansible/awx-operator/blob/devel/.github/workflows/ci.yaml#L10C3-L43C106) the block that's failing?

ToxicGLaDOS:
@TheRealHaoLiu Just wanted to follow up. Can you confirm that the section I linked above is the part that's failing? I think that due to permissions I can't see a lot of the output of the checks so it's hard for me to understand why we would be missing openshift when it's installed with that requirements.txt.

rooftopcellist:
@ToxicGLaDOS Yes, but more specifically, it's actually failing in the installer role during the reconciliation loop.  When `molecule test -s kind` is run, it runs through the reconciliation loop, as the operator normally would when deploying awx.

In the `update_status.yml`, it hit's the error you showed above.

```
TASK [installer : Update admin password status] ********************************\\r\\n
task path: /opt/ansible/roles/installer/tasks/update_status.yml:2\\nAn exception occurred during task execution. To see the full traceback, use -vvv. The error was: ModuleNotFoundError: No module named 'openshift'\\r\\n
fatal: [localhost]: FAILED! => {\\\"changed\\\": false, \\\"error\\\": \\\"No module named 'openshift'\\\", \\\"msg\\\": 
"Failed to import the required Python library (openshift) on osdk-controller-manager-d85c658dc-snw25's Python /usr/bin/python3. Please read the module documentation and install it in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter\\\"}
```

This leads me to believe that new ansible-operator:v1.31.0 image does not include the python `openshift`, and only includes `kubernetes`.  Which indeed seems to be the case:

```bash
$ docker run -it --entrypoint='' quay.io/operator-framework/ansible-operator:v1.28.1 bash -c 'pip freeze | egrep "openshift|kubernetes"'
kubernetes==26.1.0
openshift==0.13.1
```

Here is v1.31.0

```bash
$ docker run -it --entrypoint='' quay.io/operator-framework/ansible-operator:v1.31.0 bash -c 'pip freeze | egrep "openshift|kubernetes"'
kubernetes==26.1.0
```

rooftopcellist:
@ToxicGLaDOS We could add this to the Dockerfile to solve this:

```
RUN pip3 install openshift
```

cc @TheRealHaoLiu and @dsavineau  in case they have a better solution.

ToxicGLaDOS:
@rooftopcellist Ah I think I understand. You're saying that it's the docker container running in `kind` that's failing, _not_ the the git CI container (where we do install `openshift`). Right?

dsavineau:
> RUN pip3 install openshift

I think a better solution would be to update the `operator_sdk.util` ansible collection (currently pinned to 0.4.0) [1] since the latest release (0.5.0) removed the openshift dependency and only rely on the kubernetes one [2]

[1] https://github.com/ansible/awx-operator/blob/devel/requirements.yml#L5-L6
[2] https://github.com/operator-framework/operator-sdk-ansible-util/commit/4c8148685932ffd8ef895d54eb2c135ece5ed4af

fosterseth:
@ToxicGLaDOS can you also bump operator_sdk.util in requirements.yml to 0.5.0? that should take care of the dependency issue

rooftopcellist:
@ToxicGLaDOS Yes, is is the actual operator container that is erroring because it doesn't have the `openshift` package.  However, like @dsavineau mentioned, that dependency is no longer needed if we bump the `operator_sdk.util` ansible collection to 0.5.0
* https://github.com/ansible/awx-operator/blob/aee1b464ed5a41cf002648003508507c61b03e81/requirements.yml#L6

No need to change the molecule test requirements.yml as it is unpinned and will use the latest:
* https://github.com/ansible/awx-operator/blob/aee1b464ed5a41cf002648003508507c61b03e81/molecule/requirements.yml#L6

rooftopcellist (approved):
I just tested this out, and didn't find any issues.  The image builds and the reconciliation loop completes.  Bumping operator_sdk.util to 0.5.0 did the trick. Thanks!

PR #1553, opened 2023-09-08
Enforce that Projects Persistance SC is only set to be RWX

rooftopcellist:
##### SUMMARY

Now that web and task are now separate deployments, ReadWriteOnce is no longer a valid option for the projects persistence feature since the mounted `/projects` directory will be accessed by pods from different deployments now.  ReadWriteMany is required.  As such, we should validate that this is set as ReadWriteMany, and deprecate this parameter in the future. 

##### ISSUE TYPE

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -1776,9 +1776,11 @@ spec:
                 default: 8Gi
                 type: string
               projects_storage_access_mode:
-                description: AccessMode for the /var/lib/projects PersistentVolumeClaim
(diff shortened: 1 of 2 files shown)


rooftopcellist:
I think this PR is unneeded because there is a somewhat valid use case for using RWO for the projects persistence PVC.

I don't think that the web container needs access to /var/lib/awx/projects.  If it doesn't, then we can just remove the volume mount from the web deployment and solve this issue that way.  If so, I think we should just remove the projects_persistence mounts from the web container and keep providing the ability to use a RWO PVC for projects persistence.  

If RWX is not available, it still stands that the user would need to set node_selector for the task pod if replicas is set > 1 so that all of those pods get scheduled on the same node.

PR #1554, opened 2023-09-08
Add sample AWX CR with resource limits for environments with Resource…

rooftopcellist:
##### SUMMARY
Add sample AWX CR with resource limits for environments with ResourceQuotas

This is follow-up for this PR which attempted to add the limits to the defaults.  We decided against this as it would break some existing deployments in the wild.  
* https://github.com/ansible/awx-operator/pull/1541 cc @imedaouidene 

##### ISSUE TYPE

### DIFF
--- a/config/samples/awx_v1beta1_awx_resource_limits.yaml
+++ b/config/samples/awx_v1beta1_awx_resource_limits.yaml
@@ -1,0 +1,55 @@
+---
+apiVersion: awx.ansible.com/v1beta1
+kind: AWX
+metadata:
+  name: awx-with-limits
+spec:
+  task_resource_requirements:
+    requests:
+      cpu: 100m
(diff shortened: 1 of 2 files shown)


rebeccahhh (approved):
I like this, I like upping the CPU even while bringing other values down. If I can I'll run it through some tests in the coming weeks and then if there are any update suggestions I have I'll make a PR and tag you.

rooftopcellist:
Thanks @rebeccahhh !

The big win here is that we will now provide an example of how to configure resource_requirements for all of the containers for when limits are required (AKA when customers/users have a ResourceQuota in the namespace being deployed to).  

This provides an example that users can use as a starting point to make sure they set limits on all containers needed for AWX.  They should still adjust values as needed for their deployments.

PR #1575, opened 2023-10-02
awx-operator-controller-manager reduce limit to 30 times as much as request

### LINKED ISSUE #1479: awx-operator-controller-manager reduce limits

### Feature Summary

The resources limits of the awx-operator-controller-manager pod consist of the following two containers:

* a well known kube-rbac-proxy `CPU: 500M, memory: 128Mi `
* a very huge [AWX-Manager](https://github.com/ansible/awx-operator/blob/1bddb984766c91998991db475f3ecbfb34c47b4c/config/manager/manager.yaml#L75C1-L77C25): `CPU: 2000M, memory: 4096Mi`

That's together: cpu: 2500m, memory: 4224Mi

Our cluster has set a limit range of maximum cpu: 2, memory: 4Gb for all pods. So awx-manager is the only deployment that exceeds this global limit.

Currently we have given the AWX namespace a dedicated limit-range with slightly increased values.

But maybe it would make sense to lower the limits of awx-operator-controller-manager? Most pods in deployments in the wild (e.g. AMQ streams) have a maximum limit of cpu: "1", memory: 500Mi.

Also, our productive awx-operator-controller-manager shows a minimal usage of resources (mostly cpu: 2m, memory: 50Mi over 2 weeks) 

Would you accept a PR with this?

klml:
##### SUMMARY
fixes #1479

##### ISSUE TYPE

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -72,8 +72,8 @@ spec:
             memory: "32Mi"
             cpu: "50m"
           limits:
-            memory: "4096Mi"
-            cpu: "2000m"
+            memory: "960Mi"
+            cpu: "1500m"
       serviceAccountName: controller-manager
       imagePullSecrets:
       - name: redhat-operators-pull-secret


rooftopcellist:
@klml 
cc @rebeccahhh @kdelee 

We should base these values off of the real life performance of the operator in the most. I think we probably want to install the AWX Operator, then create the following:
* AWX CR

Then once reconciled, create:
* Create 2 backups
* Modify the AWX CR

Run this command to get the resource utilization while it is reconciling and use that as a baseline to inform our decision here.

```
kubectl top pods -n awx
```

rooftopcellist:
Here is a useful script for reporting the usage of the operator pod over time:
* https://gist.github.com/rooftopcellist/682e666d808c5af3ef114b7fa7b0c655

With 3 reconciliations at the same time:
* AWX CR
* 2 AWXBackup CR


I see:

```
Monitoring Pod: awx-operator-controller-manager-78779cdf79-9jhnt
+------------------+-----------+-----------+-----------+
| Resource         |   Average |   Minimum |   Maximum |
+==================+===========+===========+===========+
| CPU (millicores) |   984.907 |         3 |      1798 |
+------------------+-----------+-----------+-----------+
| Memory (Mi)      |   325.981 |        54 |       461 |
+------------------+-----------+-----------+-----------+
```

Finally, I tested with 2 awx CR's and 2 backups:

```
$ oc get awx
NAME       AGE
awx        86s
awx-demo   75s
[chadams@awx awx-operator]$ oc get awxbackup
NAME                   AGE
1awxbackup-2022-11-3   22s
awxbackup-2022-11-3    66s

$ oc get awxrestore
NAME       AGE
restore1   2m41s
```

I created the restore after the awx instance had deployed, then edited a field on each awx CR to trigger the reconciliation loop.  So this had 5 concurrent reconciliation loops presumably.  

```
$ python monitor-operator-pod-usage.py --deployment_name "awx-operator-controller-manager" --namespace "awx" --total_time 300
Monitoring Pod: awx-operator-controller-manager-78779cdf79-nhcbt
+------------------+-----------+-----------+-----------+
| Resource         |   Average |   Minimum |   Maximum |
+==================+===========+===========+===========+
| CPU (millicores) |    304.98 |        76 |       440 |
+------------------+-----------+-----------+-----------+
| Memory (Mi)      |    469.02 |       383 |       544 |
+------------------+-----------+-----------+-----------+
```

rooftopcellist:
Given the performance report above, I think the Memory and CPU limits proposed are reasonable.  CPU is compressible, so it is OK that at times it will peak over the CPU limit when pushed hard.

rooftopcellist (approved):

PR #1584, opened 2023-10-10
Fix operator-hub publish script on mac

TheRealHaoLiu:
##### SUMMARY
Remove 
- `bundle/manifests/awx-operator.clusterserviceversion.yaml-e` 
- `bundle/metadata/annotations.yaml-e` 
after generating the bundle during operator-hub publish script

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

```

```


dsavineau (commented):
While I agree with the changes as a temporary workaround then I think we should try to make this script compatible with non GNU tools (especially for Mac) mainly with the sed options.

dsavineau:
the issue is related to the `-i `sed option which doesn't behave the same with the BSD and GNU version

One solution would be to always use `-i.bak` for all platforms (meaning backuping the source file with .bak suffix) and then remove the .bak files

rooftopcellist (approved):

TheRealHaoLiu:
applied the change based on @dsavineau 's recommendation

PR #1600, opened 2023-10-17
Install operator-sdk if needed

TheRealHaoLiu:
taken inspiration from other tools used in this repo like "YQ" and "JQ" where we try to install it in bin/ if it does not exist

this should let you run any of the make command that pre-req operator-sdk without having operator-sdk installed already

also update operator-sdk version to match current opeartor sdk version

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

```

```


rooftopcellist (approved):

thedoubl3j (approved):
tested locally and worked as expected

Issue #1590, opened 2023-10-11
remove chmod and chown from roles/backup/tasks/postgres.yml 


romankuzmik:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that the AWX Operator is open source software provided for free and that I might not receive a timely response.

### Bug Summary

roles/backup/tasks/postgres.yml  has a step to change ownership on {{ backup_dir }}/tower.db to ":root".
this should be made option or ideally taken out. This is wrong to assume that current user can assign "root" group.
all permissions on a mounted PVC are managed via pod securityContext setting. there is no need to do anything in addition.

### AWX Operator version

2.7.0

### AWX version

23.3.0

### Kubernetes platform

kubernetes

### Kubernetes/Platform version

v1.24.9

### Modifications

no

### Steps to reproduce

* run backup process as non-root user
* observe a failure on "Set permissions on file for database dump", specifically on "chown :root {{ backup_dir }}/tower.db" part

### Expected results

do not change file ownership, rely on k8s default behavior

### Actual results

 failure on "Set permissions on file for database dump", specifically on "chown :root {{ backup_dir }}/tower.db" part

PR #1602, opened 2023-10-18
1590 flag to control execution of the chmod/chown on backup data file

romankuzmik:
##### SUMMARY
Depends on a storage driver configured for Backup PVC it might not be possible to change file mode or ownership on the backup file produced. This change introduces new flag that can turn off "chmod/chown".
If we run db-management pod as root, it does not matter that PVC volume mount would allow root group ownership.
Without pod's securityContext/fsGroup specified it would normally be nobody:nobody, instead of root:root. This "chown :root" might fail with some k8s storage drivers.

##### ISSUE TYPE
 - Bug

##### ADDITIONAL INFORMATION
See [Issue 1590](https://github.com/ansible/awx-operator/issues/1590)

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxbackups.yaml
+++ b/config/crd/bases/awx.ansible.com_awxbackups.yaml
@@ -78,6 +78,10 @@ spec:
               clean_backup_on_delete:
                 description: Flag to indicate if backup should be deleted on PVC if AWXBackup object is deleted
                 type: boolean
+              set_permissions_on_backup_dump:
+                description: Flag to indicate if backup dump file mode should be set with '600' with ownership to ':root'
+                type: boolean
+                default: true
               pg_dump_suffix:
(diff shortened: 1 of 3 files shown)


chrismeyersfsu:
@romankuzmik I looked into if there is/was a need for the underlying chmod + chown and I can not find one. I say just change your PR to just delete the chmod + chown task.

https://github.com/ansible/awx-operator/pull/1602/files#diff-e953e7a3d1f04931f56d660f00c5bf9aab9e7f88ab3ece25dc6f9367537ad6f6R78

TheRealHaoLiu:
Put up a PR for removing the chmod/chown altogether https://github.com/ansible/awx-operator/pull/1667

rooftopcellist:
@romankuzmik Thank you for the PR, merging these changes in a separate PR here to get it in:
* https://github.com/ansible/awx-operator/pull/1667

Issue #1586, opened 2023-10-10
awx-operator restore from PVC is not working


romankuzmik:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that the AWX Operator is open source software provided for free and that I might not receive a timely response.

### Bug Summary

awx-operator/roles/restore/tasks/init.yml has wrong conditions on "Set variables from awxbackup object statuses if provided" block.

instead of:
```
  when:
    - backup_name != '' or backup_name is defined
```

it should be
```
  when:
    - backup_name is defined and backup_name != '' 
```

otherwise restore from PVC will fail with:
```
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'backup_name != '' or backup_name is defined' failed. The error was: error while evaluating conditional (backup_name != '' or backup_name is defined): 'backup_name' is undefined. 'backup_name' is undefined
``` 

### AWX Operator version

2.7.0

### AWX version

23.3.0

### Kubernetes platform

other (please specify in additional information)

### Kubernetes/Platform version

v1.24.9

### Modifications

no

### Steps to reproduce

1. Create restore from PVC manifest 
```
apiVersion: awx.ansible.com/v1beta1
kind: AWXRestore
metadata:
  name: somebackup
  namespace: awx
spec:
  backup_dir: mybackupishere
  backup_pvc: mypvc
  backup_source: PVC
  deployment_name: awx
  image_pull_policy: IfNotPresent
  no_log: false
  set_self_labels: true
```
3. Observe an error on operator pod

### Expected results

AWX restore to complete successfully from PVC and folder within PVC specified

### Actual results

AWX restore is in error loop

kurokobo:
This is due to the version of Ansible in the Operator SDK.

In Operator 2.6.0 or eariler, Ansible 2.9.x is used in Operator SDK 1.28.x and the conditional works as expected.

In Operator 2.7.0, Operator SDK is bumped to 1.31.x that uses Ansible 2.15.x in https://github.com/ansible/awx-operator/pull/1550 and conditional is broken.

**Operator 2.6.0**

```bash
$ docker run --rm -it --entrypoint bash quay.io/ansible/awx-operator:2.6.0

bash-4.4$ ansible --version
ansible 2.9.27
[...]

PR #1603, opened 2023-10-18
1586: awx-operator restore from PVC is not working

romankuzmik:
##### SUMMARY
An upgrade to Operator SDK 1.31.x that uses Ansible 2.15 caused 
awx-operator/roles/restore/tasks/init.yml conditions on "Set variables from awxbackup object statuses if provided" block to fail.

##### ISSUE TYPE

##### ADDITIONAL INFORMATION
See [Issue 1586](https://github.com/ansible/awx-operator/issues/1586)


rooftopcellist (approved):
Thank  you for putting together a fix for this @romankuzmik !

Issue #1587, opened 2023-10-10
awx-operator restore is stuck on pg_restore step


romankuzmik:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that the AWX Operator is open source software provided for free and that I might not receive a timely response.

### Bug Summary

AWXRestore from CR is failing on "Restore database dump to the new postgresql container" step while executing `pg_restore --clean --if-exists ...`

### AWX Operator version

2.7.0

### AWX version

23.3.0

### Kubernetes platform

kubernetes

### Kubernetes/Platform version

v1.24.9

### Modifications

no

### Steps to reproduce

1. Install AWX via operator with default/empty database
2. Perform backup create AWXBackup manifest
3. Perform restore creating AWXRestore manifest pointing to CR created in step 2
4. Observe an issue

### Expected results

pg_restore to complete

### Actual results

restore process is failing on awx-operator/roles/restore/tasks/postgres.yml "Restore database dump to the new postgresql container" step

### Operator Logs

```
pg_restore: error: could not execute query: ERROR:  cannot drop inherited constraint \\\"main_adhoccommandevent_20231005_21_pkey\\\" of relation \\\"main_adhoccommandevent_20231005_21\\\"\", \"Command was: ALTER TABLE IF EXISTS ONLY public.main_adhoccommandevent_20231005_21 DROP CONSTRAINT IF EXISTS main_adhoccommandevent_20231005_21_pkey;\",
```
and many more of the similar pg_restore errors

metabsd:
I think I have the same problem trying to migrate from postgresql12 to 13. I see an error about trying to stream the backup to the new postgresql instance. I try again tomorrow to provide more detail.

kurokobo:
@romankuzmik 
Have you performed following steps before restoration?

> *Before Restoring from a backup*, be sure to:
>   - delete the old existing AWX CR
>   - delete the persistent volume claim (PVC) for the database from the old deployment, which has a name like `postgres-13-<deployment-name>-postgres-13-0`
> 
> https://github.com/ansible/awx-operator/tree/devel/roles/restore#requirements

If you plan to reuse old PV/PVC for newly restored DB, removing containing data in PV before restoration is also recommended.

romankuzmik:
[...]

PR #1604, opened 2023-10-18
1587: drop/create database DB before restoring it

romankuzmik:
##### SUMMARY
Currently we require re-create PVC used by Postgres or remove all data manually before running restore. See https://github.com/ansible/awx-operator/tree/devel/roles/restore#requirements 

This might not be always possible neither it is convenient for end users. This PR introduces simpler solution:
* Force drop existing database
* Re-create empty database
* Resume as usual (pg_resore & Co)

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
See [Issue 1587](https://github.com/ansible/awx-operator/issues/1587)

### DIFF
--- a/roles/restore/tasks/postgres.yml
+++ b/roles/restore/tasks/postgres.yml
@@ -82,6 +82,26 @@
       -p {{ awx_postgres_port }}
   no_log: "{{ no_log }}"
 
+- name: Set drop db command
+  set_fact:
+    pg_drop_db: >-
+      echo 'DROP DATABASE {{ awx_postgres_database }} WITH (FORCE);' | PGPASSWORD='{{ awx_postgres_pass }}' psql
+      -U {{ awx_postgres_user }}
+      -h {{ resolvable_db_host }}
+      -d postgres
+      -p {{ awx_postgres_port }}
+  no_log: "{{ no_log }}"
+
+- name: Set create db command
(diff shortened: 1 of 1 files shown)


rooftopcellist:
@romankuzmik I could see how this would be useful, but we probably don't want to run it every time because of the nature of how dangerous `DROP DATABASE` is.  We probably only want users to do this when issues arise.  

What do you think about adding a `force` boolean parameter on the AWXRestore object that would optionally inject those lines into the pg_restore command?  That way users could be intentional about doing this.

TheRealHaoLiu:
@romankuzmik i created a PR based on yours https://github.com/ansible/awx-operator/pull/1639

romankuzmik:
@TheRealHaoLiu , I like it. Hope we can get an approval and merge it sooner or later.

TheRealHaoLiu:
Working on review and testing it now. If u don't mind I will push some more commits around naming changes

rooftopcellist:
Closing as this logic was merged as part of https://github.com/ansible/awx-operator/pull/1639

Thank you for your contribution!

haifang-zh:
Hi, 
I am using AWX-operator of 2.19.1 when I set force_drop_db to true db user 'awx' need privilege of create DB

PR #1626, opened 2023-11-08
Add ability to mount custom volumes in the postgres pod

rooftopcellist:
##### SUMMARY

For Posix shared memory, some users need to be able to do a custom volume mount inside the postgres container.  This makes that possible from the AWX CR spec.  

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

Docs on shared memory for posix:
* https://docs.openshift.com/online/pro/dev_guide/shared_memory.html

### DIFF
--- a/roles/installer/templates/statefulsets/postgres.yaml.j2
+++ b/roles/installer/templates/statefulsets/postgres.yaml.j2
@@ -124,11 +124,10 @@ spec:
       tolerations:
         {{ postgres_tolerations | indent(width=8) }}
 {% endif %}
-{% if postgres_extra_volumes -%}
+{% if postgres_extra_volumes %}
       volumes:
-        {{ postgres_extra_volumes | indent(width=8, first=True) }}
+        {{ postgres_extra_volumes | indent(width=8, first=False) }}
 {% endif %}
-
   volumeClaimTemplates:
     - metadata:
         name: postgres-{{ supported_pg_version }}
--- a/roles/installer/templates/statefulsets/postgres.yaml.j2
+++ b/roles/installer/templates/statefulsets/postgres.yaml.j2
@@ -112,6 +112,9 @@ spec:
             - name: postgres-{{ supported_pg_version }}
               mountPath: '{{ postgres_data_path | dirname }}'
               subPath: '{{ postgres_data_path | dirname | basename }}'
+{% if postgres_extra_volume_mounts -%}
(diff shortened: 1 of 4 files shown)


Zokormazo (approved):

TheRealHaoLiu (approved):

roles/installer/templates/statefulsets/postgres.yaml.j2
```diff
@@ -121,6 +124,11 @@ spec:
       tolerations:
         {{ postgres_tolerations | indent(width=8) }}
 {% endif %}
+{% if postgres_extra_volumes -%}
+      volumes:
+        {{ extra_volumes | indent(width=8, first=True) }}
```
rooftopcellist: One more change needed, this should be `postgres_extra_volumes`

roles/installer/templates/statefulsets/postgres.yaml.j2
```diff
@@ -121,6 +124,11 @@ spec:
       tolerations:
         {{ postgres_tolerations | indent(width=8) }}
 {% endif %}
+{% if postgres_extra_volumes -%}
+      volumes:
+        {{ postgres_extra_volumes | indent(width=8, first=True) }}
+{% endif %}
```
TheRealHaoLiu:
```suggestion
{% if postgres_extra_volumes %}
      volumes:
        {{ postgres_extra_volumes | indent(width=8, first=False) }}
{% endif %}
```

roles/installer/templates/statefulsets/postgres.yaml.j2
```diff
@@ -121,6 +124,11 @@ spec:
...
+
```
TheRealHaoLiu:
```suggestion
```

rooftopcellist:
I just tested this out and the custom volume mount does indeed get set on the StatefulSet.

```
$ oc get sts awx-postgres-13 -o yaml | grep "volumeMounts:" -A5
        volumeMounts:
        - mountPath: /var/lib/postgresql/data
          name: postgres-13
          subPath: data
        - mountPath: /dev/shm
          name: shm
```

This is the test CR I used:

```
  postgres_extra_volumes: |
    - name: shm
      emptyDir:
        medium: Memory
        sizeLimit: 1Gi
  postgres_extra_volume_mounts: |
    - name: shm
      mountPath: /dev/shm

```

PR #1637, opened 2023-11-14
Omit storage resource limit if empty

TheRealHaoLiu:
##### SUMMARY
We discovered some weird behavior observed on later Kubernetes version (OCP 4.12+)

For some reason why we apply the templates postgres resource with

```
postgres_storage_requirements:
  limit: {}
  requests:
    storage: <x>Gi
```

the `Create Database if no database is specified` task that does the k8s apply will always think the resource is "changed" and proceed to cycle the task and web pod

This resulted in AWX pods being continuously restarted

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/roles/installer/templates/statefulsets/postgres.yaml.j2
+++ b/roles/installer/templates/statefulsets/postgres.yaml.j2
@@ -33,7 +33,7 @@ spec:
         app.kubernetes.io/instance: 'postgres-{{ supported_pg_version }}-{{ ansible_operator_meta.name }}'
         app.kubernetes.io/component: 'database'
         app.kubernetes.io/part-of: '{{ ansible_operator_meta.name }}'
-        app.kubernetes.io/managed-by: '{{ deployment_type }}-operator' 
+        app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
     spec:
 {% if image_pull_secret is defined %}
       imagePullSecrets:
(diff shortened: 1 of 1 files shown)


roles/installer/templates/statefulsets/postgres.yaml.j2
```diff
@@ -137,7 +137,15 @@ spec:
 {% if postgres_storage_class is defined %}
         storageClassName: '{{ postgres_storage_class }}'
 {% endif %}
-        resources: {{ postgres_storage_requirements }}
+{% if postgres_storage_requirements.limit or postgres_storage_requirements.requests %}
```
rooftopcellist: I think you may need a conditional checking if `postgres_storage_requirements` only is defined first in addition to what you have.  I'm not sure, I will have to test it out.

roles/installer/templates/statefulsets/postgres.yaml.j2
```diff
@@ -137,7 +137,15 @@ spec:
...
+        resources:
+{% if postgres_storage_requirements.limit %}
+          limit: {{ postgres_storage_requirements.limit }}
```
davidtluong: Should we even allow the customer to set this? When I looked up if this field had any affect on pvc's, everything came up no

TheRealHaoLiu: it should be fine...

rooftopcellist:
Currently, when testing with the following spec:

```yaml
---
kind: AWX
metadata:
  name: awx
spec:
  no_log: false
  postgres_storage_requirements:
    requests:
      storage: 5Gi
```

I see this error:
```
fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'limit'. 'dict object' has no attribute 'limit'. 'dict object' has no attribute 'limit'. 'dict object' has no attribute 'limit'\n\nThe error appears to be in '/opt/ansible/roles/installer/tasks/database_configuration.yml': line 152, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n- block:\n    - name: Create Database if no database is specified\n      ^ here\n"}
```

roles/installer/templates/statefulsets/postgres.yaml.j2
TheRealHaoLiu: postgres_storage_requirements should always be defined since we have it in default

rooftopcellist (approved):

PR #1638, opened 2023-11-14
Make PostgreSQL PVC storage requirements immutable

rooftopcellist:
##### SUMMARY

k8s does not allow a user to modify a PVC's size after it has been created, so we should prevent users from trying to change this on the AWX CR after the initial deployment.  Currently, if a user _does_ change the requests or limits on the db PVC, the operator will crashloop with the following error:

```
Forbidden: updates to statefulset spec for fields other than \\'replicas\\', \\'template\\', \\'updateStrategy\\', \\'persistentVolumeClaimRetentionPolicy\\' and \\'minReadySeconds\\' are forbidden\",\"field\":\"spec\"}]},\"code\":422}\\n'", "reason": "Unprocessable Entity"}      
```

##### ISSUE TYPE

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -1680,12 +1680,19 @@ spec:
                     properties:
                       storage:
                         type: string
+                        default: 8Gi
                     type: object
+                    x-kubernetes-validations:
+                    - rule: "self == oldSelf"
+                      message: "The postgres_storage_requirements.requests field is immutable and cannot be changed once set."
                   limits:
                     properties:
                       storage:
                         type: string
                     type: object
+                    x-kubernetes-validations:
+                    - rule: "self == oldSelf"
+                      message: "The postgres_storage_requirements.limits field is immutable and cannot be changed once set."
                 type: object
               postgres_resource_requirements:
                 description: Resource requirements for the PostgreSQL container
--- a/config/manifests/bases/awx-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/awx-operator.clusterserviceversion.yaml
@@ -583,8 +583,8 @@ spec:
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:com.tectonic.ui:hidden
-      - displayName: Postgres Extra Volumes
-        description: Specify extra volumes to add to the postgres pod
+      - description: Specify extra volumes to add to the postgres pod
+        displayName: Postgres Extra Volumes
         path: postgres_extra_volumes
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced


config/manifests/bases/awx-operator.clusterserviceversion.yaml
```diff
@@ -583,8 +583,8 @@ spec:
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:com.tectonic.ui:hidden
-      - displayName: Postgres Extra Volumes
-        description: Specify extra volumes to add to the postgres pod
+      - description: Specify extra volumes to add to the postgres pod
```
rooftopcellist, line 586: I added in this so that `make bundle` doesn't create diff for it every time.

config/crd/bases/awx.ansible.com_awxs.yaml
```diff
@@ -1680,12 +1680,19 @@ spec:
                     properties:
                       storage:
                         type: string
+                    default: 8Gi
```
TheRealHaoLiu:
```suggestion
                    default: 8Gi
```
```suggestion
                        type: string
                        default: 8Gi
```

tylergmuir:
I ran into this same error that you mentioned and documented it in #1642. That said, I think we should do the reverse and still allow it in the CRD, but then in the ansible code that the operator runs, have it better handle that change to the PVC. There is already error catching in that step, the issue is that the error catching doesn't quite work as expected.

The issue I see with making this immutable is that there is no process to increase the size without doing a lot of manual work. Alternatively, if the field is mutable and the error handling in the playbook works as expected, then you still need to manually resize the PVC (due to Kubernetes not automatically resizing PVC from statefulstatesets) but then you have a resized PVC.

If we do decide to go the route of making the fields immutable, we need to also include much better documentation on how to migrate to a larger PVC.

TheRealHaoLiu (approved):

rooftopcellist:
Marking this as a draft PR for now because I think we need to workshop how best to handle this and weigh up our options more.

rooftopcellist:
@tylergmuir You bring up a good point, if a user has a storage class that _does_ support dynamic expansion, this would make things more difficult for them to utilize it.  

It would be great to take the time to figure out the steps needed for expanding the db PVC and document them in a markdown file here in the repo.  I think we would want to cover two scenarios:

A) How to migrate if your storageclass supports dynamic expansion (currently modify the AWX CR spec `postgres_storage_requirements` setting, then delete the StatefulSet and wait for the next reconciliation loop to re-create it)

B) How to use the [migration.md](https://ansible.readthedocs.io/projects/awx-operator/en/latest/migration/migration.html) logic to migrate the awx instance to a fresh instance and provision a larger PVC in the process.


I will close this PR and open an issue for adding documentation for this.  It would be nice to find a way to prevent users without SC's that support dynamic expansion from falling into this reconcilation fail loop... Maybe the option described [here](https://github.com/ansible/awx-operator/issues/1642#issuecomment-1836742730) could help with that.

rooftopcellist:
I created an issue for documenting the steps currently needed to expand a PVC.
* https://github.com/ansible/awx-operator/issues/1655

PR #1643, opened 2023-11-20
Adding in Kind specific install docs

rooftopcellist:
##### SUMMARY

Adding in docs for how to run the awx-operator and AWX on a Kind cluster.  I think it would be useful to do this for the main k8s flavors, or at least cover the Networking quirks of each.  

##### ISSUE TYPE

### DIFF
--- a/docs/installation/kind-install.md
+++ b/docs/installation/kind-install.md
@@ -1,0 +1,125 @@
+# AWX Operator on Kind
+
+## Kind Install
+
+Install Kind by running the following
+
+```
+# For Intel Macs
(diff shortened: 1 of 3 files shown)


docs/installation/kind-install.md
```diff
@@ -0,0 +1,120 @@
...
+nodes:
+- role: control-plane
+  extraPortMappings:
+  - containerPort: 32000
+    hostPort: 32000
+    listenAddress: "0.0.0.0" # Optional, defaults to "0.0.0.0"
+    protocol: tcp # Optional, defaults to tcp
+- role: worker
+then:
+kind create cluster --config=kind.config
```
dsavineau: I guess we should move this out of the yaml block and create a separate bash bock right ?

dsavineau (approved):

PR #1648, opened 2023-11-24
update deprecated collections path variable

justsomescripts:
##### SUMMARY

This PR replaces the `ANSIBLE_COLLECTIONS_PATHS` environment variable in AWX Web that has been deprecated and replaced by `ANSIBLE_COLLECTIONS_PATH` (available since Ansible 2.10).

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

**Warning message when executing jobs**
```
[DEPRECATION WARNING]: ANSIBLE_COLLECTIONS_PATHS option, does not fit var 
naming standard, use the singular form ANSIBLE_COLLECTIONS_PATH instead. This 
feature will be removed from ansible-core in version 2.19. Deprecation warnings
can be disabled by setting deprecation_warnings=False in ansible.cfg.
```

https://docs.ansible.com/ansible/latest/reference_appendices/config.html#envvar-ANSIBLE_COLLECTIONS_PATHS

### DIFF
--- a/roles/installer/templates/configmaps/config.yaml.j2
+++ b/roles/installer/templates/configmaps/config.yaml.j2
@@ -54,9 +54,6 @@ data:
 
     INTERNAL_API_URL = 'http://127.0.0.1:8052'
 
-    # Sets Ansible Collection path
-    AWX_ANSIBLE_COLLECTIONS_PATHS = '/var/lib/awx/vendor/awx_ansible_collections'
-
     # Container environments don't like chroots
     AWX_PROOT_ENABLED = False
 


shanemcd:
I think this is dead code and should get deleted. Pretty sure this is what needs to change:

https://github.com/ansible/awx/blob/3b4d95633e5302c6b67903f7b9e737d264c154fd/awx/main/tasks/jobs.py#L928

shanemcd:
Actually... it looks like we already accounted for the name change:

https://github.com/ansible/awx/blob/3b4d95633e5302c6b67903f7b9e737d264c154fd/awx/main/tasks/jobs.py#L930

I think we should probably just delete this line of code from the operator configmap.

justsomescripts:
Makes sense. thanks for clarifying @shanemcd 👍

> I think we should probably just delete this line of code from the operator configmap.

I'll update the PR to just remove this line then so it's searching for `ANSIBLE_COLLECTIONS_PATHS` in this repo is less confusing.

TheRealHaoLiu (approved):

PR #1651, opened 2023-11-30
Bump Ansible Operator SDK version to 1.32.0 for OS updates

rooftopcellist:
##### SUMMARY

Use the new v1.32.0 ansible-operator base image to pick up OS updates.

There are no repo updates required this time besides bumping the image and makefile references:
* https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.32.0/

##### ISSUE TYPE

 - New or Enhanced Feature

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
-FROM quay.io/operator-framework/ansible-operator:v1.31.0
+FROM quay.io/operator-framework/ansible-operator:v1.32.0
 
-USER 0
-
-RUN dnf install -y openssl
+USER root
+RUN dnf update --security --bugfix -y && \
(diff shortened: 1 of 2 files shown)


dsavineau (approved):
I think we should add (but could be done in a separate PR) a `dnf update` task so the awx-operator container image isn't just updated (for RPM package update) on SDK bump but for each awx-operator release too.

rooftopcellist:
Good call, I've updated the PR to include that if you want to take another look.

rooftopcellist:
We are just hitting Quay.io sporadic pull issues at the moment, CI should pass on a re-run once that is stable again.

PR #1654, opened 2023-12-01
FIx permissions /var/lib/awx/projects when projects_persistence is true.

paytroff:
##### SUMMARY

When 'projects_persistence' is false, permissions /var/lib/awx/projects are 777 with root:root.
When 'projects_persistence' is true, permissions /var/lib/awx/projects are 775 with root:1000.
But the problem is : group id 1000 does not exist in ansible/awx container, so user 'awx' cannot write into /var/lib/awx/projects directory when you sync project in AWX UI.

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

```
Before:
chgrp 1000 /var/lib/awx/projects
After:
chgrp 0 /var/lib/awx/projects
```

### DIFF
--- a/roles/installer/templates/deployments/task.yaml.j2
+++ b/roles/installer/templates/deployments/task.yaml.j2
@@ -129,7 +129,7 @@ spec:
             - -c
             - |
               chmod 775 /var/lib/awx/projects
-              chgrp 1000 /var/lib/awx/projects
+              chgrp 0 /var/lib/awx/projects
           env:
             - name: MY_POD_NAME
               valueFrom:
--- a/roles/installer/templates/deployments/web.yaml.j2
+++ b/roles/installer/templates/deployments/web.yaml.j2
@@ -119,7 +119,7 @@ spec:
             - -c
             - |
               chmod 775 /var/lib/awx/projects
-              chgrp 1000 /var/lib/awx/projects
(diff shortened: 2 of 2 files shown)


TheRealHaoLiu:
can you provide more detail on how you tested your PR

paytroff:
I have rebuilt docker image awx-operator, create an awx instance (with new image operator) with projects_persistence = true, then see which permisions on folder /var/lib/awx/projects were set in awx-task and awx-web pod.
Then in AWX UI, launch a project synchronisation, result was ok.

Do you need more details ?

dbrennand:
Hi @TheRealHaoLiu @paytroff 

I've been experiencing issues related to permissions on /var/lib/awx/projects.

This is on a fresh deployment of AWX. However, in my testing, only the awx-web pod is affected. 

For awx-task, the AWX user in the container has group 1000 assigned so it can write at the directory just fine whereas, in the awx-web pod the user doesn't have group 1000, and AWX is crashing as a result.

I'm on mobile at the moment and can post further details later today, but I think the issue is potentially with the upstream AWX image rather than the operator deployment.

dbrennand:
https://forum.ansible.com/t/awx-web-permissionerror-errno-13-permission-denied-var-lib-awx-projects/7074

PR #1661, opened 2023-12-05
Add directory option for receptors data

AaronH88:
##### SUMMARY
Adding option for setting receptors `datadir` option which will allow users to set the location of their receptor data. The default is `/tmp/receptor/` which may not be the best option for some customers. 

##### ISSUE TYPE 
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

Tested locally by editing my `awx-demo.yml` with the following var `receptor_datadir` under `spec`

```
---
kind: AWX
metadata:
  name: awx-demo
spec:
  service_type: nodeport
  receptor_datadir: /tmp/aarontest2/
```

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -1908,6 +1908,9 @@ spec:
               receptor_log_level:
                 description: Set log level of receptor service
                 type: string
+              receptor_datadir:
+                description: Set directory for receptor logs
+                type: string
               extra_settings:
                 description: Extra settings to specify for AWX
                 items:
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -1909,7 +1909,7 @@ spec:
                 description: Set log level of receptor service
                 type: string
               receptor_datadir:
-                description: Set directory for receptor logs
+                description: Set directory for receptor data
                 type: string
               extra_settings:
                 description: Extra settings to specify for AWX
--- a/config/manifests/bases/awx-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/awx-operator.clusterserviceversion.yaml
@@ -997,7 +997,7 @@ spec:
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:com.tectonic.ui:hidden
-      - displayName: Receptor Persistant Data Directory
+      - displayName: Receptor Data Directory
         path: receptor_datadir
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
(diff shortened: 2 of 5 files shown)


config/manifests/bases/awx-operator.clusterserviceversion.yaml
```diff
@@ -889,6 +889,11 @@ spec:
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:com.tectonic.ui:hidden
+      - displayName: Receptor Persistant Data Directory
```
TheRealHaoLiu: its not persistent since /var/run/receptor is an empty dir mount

roles/installer/defaults/main.yml
```diff
@@ -458,6 +458,8 @@ host_aliases: ''
 
 receptor_log_level: info
 
+receptor_datadir: /var/run/receptor/
```
TheRealHaoLiu: this will make it that by default we log into the directory?

TheRealHaoLiu:
what goes into the datadir? is it just the log?

by default on kube i think receptor logs to STDOUT which can be accessed via `kubectl log`

having a log dir with no rotation can potentially eat up storages 

the way that the PR is written there's no way for people to NOT use the datadir

also there's no persistent storage backing the directory which mean that data will be wiped out if the pod is deleted

roles/installer/defaults/main.yml
AaronH88: not logs, but it will make us by default save all the contents of the `/tmp/receptor/NODEID/WORKID/*` to `/var/run/receptor` instead

AaronH88:
> what goes into the datadir? is it just the log?
> 
> by default on kube i think receptor logs to STDOUT which can be accessed via `kubectl log`
> 
> having a log dir with no rotation can potentially eat up storages
> 
> the way that the PR is written there's no way for people to NOT use the datadir
> 
> also there's no persistent storage backing the directory which mean that data will be wiped out if the pod is deleted

Thanks for the feedback!

The receptor data dir is the location where receptor stores all data it is currently using, as you know (because you taught me) it is currently set to `/tmp/receptor/` so the idea here is to make this configurable. I am told some customers would like this :smile: 

My description on persistent storage is incorrect and I will change it now, thanks for pointing that out.

config/crd/bases/awx.ansible.com_awxs.yaml
```diff
@@ -1792,6 +1792,9 @@ spec:
               receptor_log_level:
                 description: Set log level of receptor service
                 type: string
+              receptor_datadir:
+                description: Set directory for receptor logs
```
fosterseth:
```suggestion
                description: Set directory for receptor data
```

fosterseth (approved):

TheRealHaoLiu:
O that tmp dir 🤦‍♂️

The description of the PR mention receptor log

fosterseth (approved):

TheRealHaoLiu (approved):

AaronH88:
outside the scope of what is needed, containerised is not needed right now, closing

PR #1639, opened 2023-11-14
Add option to force drop database before restore

TheRealHaoLiu:
##### SUMMARY
Based on https://github.com/ansible/awx-operator/pull/1604

Help with recovery from a bad restore or any other database "accident" where the database is left in a state that can't perform "clean" during restore process properly

Added `force_drop_db` option in AWXRestore 

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

Fixes https://github.com/ansible/awx-operator/issues/1587

```

```

PR #1665, opened 2023-12-08
Fix PG Restore Force Drop DB flag

rooftopcellist:
##### SUMMARY
Follow-up for https://github.com/ansible/awx-operator/pull/1639

- Previously, if the flag was set to true, the bash conditional failed because the boolean was not correctly interpreted.
- Use pg_restore return code to determine if the task should be marked as failed

##### ISSUE TYPE

### DIFF
--- a/roles/restore/tasks/postgres.yml
+++ b/roles/restore/tasks/postgres.yml
@@ -82,25 +82,34 @@
       -p {{ awx_postgres_port }}
   no_log: "{{ no_log }}"
 
-- name: Set drop db command
-  set_fact:
-    pg_drop_db: >-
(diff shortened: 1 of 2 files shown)


roles/restore/tasks/postgres.yml
```diff
@@ -75,32 +75,41 @@
 - name: Set pg_restore command
   set_fact:
     pg_restore: >-
-      pg_restore --clean --if-exists
+      pg_restore
```
chrismeyersfsu: put the `--clean --if-exists` back into the code.

chrismeyersfsu (approved):

rooftopcellist:
@chrismeyersfsu Thanks for the help and the review!

PR #1670, opened 2023-12-15
add support for image pull secrets for backups and restores

### LINKED ISSUE #816: Incorrect doc: How to create registry secrets for image_pull_secret
##### ISSUE TYPE
 - Documentation

##### SUMMARY

To create `image_pull_secret`, doc says do: `kubectl create secret <resoucename>-cp-pull-credentials regcred [...]`
I guess it should be `kubectl create secret docker-registry <resoucename>-cp-pull-credentials [...]` instead.

### LINKED ISSUE #1669: add support for image pull secrets with backup and restore operations

### Feature Summary

When running either a backup or a restore of an AWX deployment, there is a Postgres pod deployed to coordinate the steps. If you are running AWX in an environment where you either want to pull from a private repo or you want to use an authenticated pull from the docker registry you are unable to do that.

The functionality does exist for the initial deployment, so much of the logic can be copied from there.

tylergmuir:
##### SUMMARY

This adds support for using `image_pull_secrets` when using AWXBackup or AWXRestore to match the functionality of the AWX install.

Fixes #1669
Fixes #816 

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

```

```


tylergmuir:
A quick note on this, I opted to only add support for `image_pull_secrets` and not `image_pull_secret` since the latter is deprecated. This does mean that if someone was using `image_pull_secret` and just copied the options in the spec straight across to the spec for a backup/restore it would not work. 

Since the `image_pull_secret` option has been deprecated for over 18 months and this is brand new functionality (with documentation), I think is fair. But I am open to other opinions on the matter.

rooftopcellist (changes requested):
@tylergmuir These changes look great!  There is just one thing that needs to be added before merging.  Could you add an entry like this in the AWXBackup and AWXRestore section of this same CSV file?
* https://github.com/ansible/awx-operator/blob/devel/config/manifests/bases/awx-operator.clusterserviceversion.yaml#L371-L375

tylergmuir:
> @tylergmuir These changes look great! There is just one thing that needs to be added before merging. Could you add an entry like this in the AWXBackup and AWXRestore section of this same CSV file?
> 
> * https://github.com/ansible/awx-operator/blob/devel/config/manifests/bases/awx-operator.clusterserviceversion.yaml#L371-L375

@rooftopcellist this is completed now.

tylergmuir:
@rooftopcellist Any chance you could take a look at this and see if there is anything else that needs to be done before a merge?

PR #1674, opened 2023-12-24
Add -ness checks and refactor migrations

dhageman:
##### SUMMARY

This PR implements liveness & readiness probes supporting the split web/task container configuration.   It also adjusts database migrations to be done in the init container of the task pod. 

Addresses #414  & #926

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

It should be acknowledged that this is not the first work on implementing these features.

### DIFF
--- a/docs/user-guide/advanced-configuration/container-probes.md
+++ b/docs/user-guide/advanced-configuration/container-probes.md
@@ -1,0 +1,40 @@
+#### Container Probes
+These parameters control the usage of liveness and readiness container probes for
+the web and task containers.
+
+#### Web / Task Container Liveness Check
+
+The liveness probe queries the status of the supervisor daemon of the container.  The probe will fail if it
+detects one of the services in a state other than "RUNNING".
+
+| Name         | Description                        | Default |
+| web_liveness_period | Time period in seconds between each probe check.  The value of 0 disables the probe. | 0    |
+| web_liveness_initial_delay | Initial delay before starting probes in seconds | 5   |
+| web_liveness_failure_threshold| Number of consecutive failure events to identify failure of container | 3    |
+| web_liveness_timeout | Number of seconds to wait for a probe response from container | 1    |
+| task_liveness_period | Time period in seconds between each probe check.  The value of 0 disables the probe. | 0    |
+| task_liveness_initial_delay | Initial delay before starting probes in seconds | 5   |
+| task_liveness_failure_threshold| Number of consecutive failure events to identify failure of container | 3    |
+| task_liveness_timeout | Number of seconds to wait for a probe response from container | 1    |
+
+#### Web Container Readiness Check
+
+This is a HTTP check against the status endpoint to confirm the system is still able to respond to web requests.
+
+| Name         | Description                        | Default |
+| -------------| ---------------------------------- | ------- |
+| web_readiness_period | Time period in seconds between each probe check.  The value of 0 disables the probe. | 0    |
+| web_readiness_initial_delay | Initial delay before starting probes in seconds | 5   |
+| web_readiness_failure_threshold| Number of consecutive failure events to identify failure of container | 3    |
+| web_readiness_timeout | Number of seconds to wait for a probe response from container | 1    |
+
+#### Task Container Readiness Check
+
+This is a command probe using the builtin check command of the awx-manage utility.
+
+| Name         | Description                        | Default |
+| -------------| ---------------------------------- | ------- |
+| task_readiness_period | Time period in seconds between each probe check.  The value of 0 disables the probe. | 0    |
+| task_readiness_initial_delay | Initial delay before starting probes in seconds | 5   |
+| task_readiness_failure_threshold| Number of consecutive failure events to identify failure of container | 3    |
+| task_readiness_timeout | Number of seconds to wait for a probe response from container | 1    |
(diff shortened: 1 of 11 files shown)


docs/user-guide/advanced-configuration/container-probes.md
```diff
@@ -0,0 +1,40 @@
...
+| task_liveness_timeout | Number of seconds to wait for a probe response from container | 1    |
+
+#### Web Container Readiness Check
+
+This is a HTTP check against the status endpoint to confirm the system is still able to respond to web requests.
+
+| Name         | Description                        | Default |
+| -------------| ---------------------------------- | ------- |
+| web_readiness_period | Time period in seconds between each probe check.  The value of 0 disables the probe. | 0    |
+| web_readiness_initial_deplay | Initial delay before starting probes in seconds | 5   |
```
chrismeyersfsu: `s/web_readiness_initial_deplay/web_readiness_initial_delay/` Spelling, remove the `p`

docs/user-guide/advanced-configuration/container-probes.md
```diff
@@ -0,0 +1,40 @@
...
+the web and task containers.
+
+#### Web / Task Container Liveness Check
+
+The liveness probe queries the status of the supervisor daemon of the container.  The probe will fail if it
+detects one of the services in a state other than "RUNNING".
+
+| Name         | Description                        | Default |
+| web_liveness_period | Time period in seconds between each probe check.  The value of 0 disables the probe. | 0    |
+| web_liveness_initial_deplay | Initial delay before starting probes in seconds | 5   |
```
chrismeyersfsu: spelling, remove the `p` in `deplay`

roles/installer/tasks/install.yml
```diff
@@ -94,48 +94,6 @@
...
-      echo 'Successful'
-      "
-  register: migrate_result
-  when:
-  - awx_task_pod_name != ''
-  - database_check is defined
-  - (database_check.stdout|trim) != '0'
-
 - name: Initialize Django
   include_tasks: initialize_django.yml
```
kdelee, line 102: so is it safe to run this before migrations may have run?
it looks like this playbook creates super user etc https://github.com/ansible/awx-operator/blob/9e2eb16b675ece46d9914234462f2f432c877a02/roles/installer/tasks/initialize_django.yml#L16 

or is there some other way we know they've already run by this point? (still reading)

kdelee: ah wait resources_configuration.yml tasks run first. and we do `wait-for-migrations` in there

roles/installer/tasks/install.yml
```diff
@@ -94,48 +94,6 @@
 - name: Include resources configuration tasks
```
kdelee, line 94:
```suggestion
# Includes wait for any migrations to complete
- name: Include resources configuration tasks
```

kdelee (approved):

erz4 (approved):

roles/installer/tasks/install.yml
dhageman: I updated the task name to reflect its expanded role.  Thank you for the input!

chrismeyersfsu (approved):

rooftopcellist:
I just ran through the code one more time. The k8s job get created as expected.

image


Then runs to completion.  The logs for the migration container show the migrations have all been completed, as can be seen by the output of:
```
$ oc logs awx-migration-23.9.0-6vwzm
```

rooftopcellist (approved):

roles/installer/tasks/migrate_schema.yml
```diff
@@ -0,0 +1,57 @@
+---
+
+- name: Check for pending migrations
```
rooftopcellist, line 3: :+1: 
This task ensures that the k8s job is only run if there are new schema changes to be migrated.  

And the unique name based on the version and the random hash ensure that there are no conflicts with other k8s jobs from previous migrations.

roles/installer/templates/jobs/migration.yaml.j2
```diff
@@ -0,0 +1,57 @@
...
+kind: Job
+metadata:
+  name: '{{ ansible_operator_meta.name }}-migration-{{ version }}'
+  namespace: '{{ ansible_operator_meta.namespace }}'
+  labels:
+    {{ lookup("template", "../common/templates/labels/common.yaml.j2")  | indent(width=4) | trim }}
+    {{ lookup("template", "../common/templates/labels/version.yaml.j2") | indent(width=4) | trim }}
+spec:
+  template:
+    spec:
```
rooftopcellist, line 11: @dhageman One thought I had is that we might want to be mindful of cleaning up these migration jobs in case a deployment gets in a bad state and winds up creating them in a loop after the timeout is reached.  The job containers are automatically cleaned up after 3600 seconds, which is pretty reasonable.  

If we need to, we can adjust this via `.spec.ttlSecondsAfterFinished`.  Though that is probably a good default to start with since it strikes a nice balance between "not cluttering things" and having logs stick around long enough for debugging.

dhageman: You make an excellent point.  There is definitely a balance there.

rooftopcellist:
@dhageman thank you for your time and effort to develop this solution.  I think it is robust and will be a good pattern.

erz4:
The table in container-probes.md is missing table format, fixed that here #1748 
@dhageman

PR #1676, opened 2023-12-29
Add support for horizontal pod autoscaling

dhageman:
##### SUMMARY

Allows the operator to create horizontal pod autoscaling resources for the web and task deployments.

It does _**not**_ provide guidance on how to configure the HPA as those details are very cluster and use case specific.

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

This code does not change the default behavior of the operator.  A horizontal pod autoscaler will not be created unless it is specifically configured.

The minimum configuration required is to set the option `task_max_replicas` or `web_max_replicas` to be an integer greater than `task_replicas` or `web_replicas` respecitively.   This should create a HPA using the default values for a HPA which on OpenShift monitors CPU consumption.

A good configuration for HPA will require requests, limits, and health probes for the the web and task pods to be implemented.

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -1574,10 +1574,18 @@ spec:
                 description: Number of web instance replicas
                 type: integer
                 format: int32
+              web_manage_replicas:
+                description: Enables operator control of replicas count for the web deployment when set to 'true'
+                type: boolean
+                default: true
               task_replicas:
                 description: Number of task instance replicas
                 type: integer
                 format: int32
+              task_manage_replicas:
+                description: Enables operator control of replicas count for the task deployment when set to 'true'
+                type: boolean
+                default: true
               web_liveness_initial_delay:
                 description: Initial delay before starting liveness checks on web pod
                 type: integer
--- a/docs/user-guide/advanced-configuration/horizontal-pod-autoscaler.md
+++ b/docs/user-guide/advanced-configuration/horizontal-pod-autoscaler.md
@@ -1,0 +1,27 @@
+### Horizontal Pod Autoscaler (HPA)
+
+Horizontal Pod Autoscaler allows Kubernetes to scale the number of replicas of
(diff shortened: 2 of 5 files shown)


TheRealHaoLiu:
@stanislav-zaprudskiy you might be interested in this

config/crd/bases/awx.ansible.com_awxs.yaml
```diff
@@ -1567,10 +1567,18 @@ spec:
                 description: Number of web instance replicas
                 type: integer
                 format: int32
+              web_unmanage_replicas:
```
kdelee: :fearful: maybe `task_manage_replicas` with default `true` bit less topsy turvy?
(e.g. all "affirmative" settings, to avoid managing double negatives)

```
description: Operator controls the number of replicas when "true". When set to false, HPA or other mechanism may manage replicas.
```

dhageman: @TheRealHaoLiu , @rooftopcellist - Thoughts on this?  I think it may be a better approach?

dhageman: @kdelee - I have this switch to your suggestion.  I appreciate it!

dathan:
Patch looks good to me. Most users of awx are probably not deploying constantly so keeping a bunch of web or task pods running does not make sense. It makes sense only when they are used. In theory with this the community can collectively save big on vCPU requested bloats that site idle for the majority of time.

trippinnik:
This would really help me out. We run a lot of jobs on schedule at particular after hours times so increasing the resources available dynamically is really needed

TheRealHaoLiu (approved):

TheRealHaoLiu:
@dhageman PR looks great, can you rebase and we can get it in

rcarrillocruz (approved):

rooftopcellist (approved):

rooftopcellist:
@dhageman can you give this a rebase?

PR #1687, opened 2024-01-16
Adopt LISTENER_DATABASES setting

TheRealHaoLiu:
##### SUMMARY
Related to https://github.com/ansible/awx/pull/14755

move TCP keepalive to pg_listener specific db setting

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

```

```


AlanCoding:
I love this, this is exactly what I had in mind. My only question here is if there's any version dependency we need to consider. It might break some people if this ran with an older AWX image.

chrismeyersfsu (approved):

AlanCoding (approved):

PR #1695, opened 2024-01-24
[PoC] using kustomize to provide flexible override to awx deployments

TheRealHaoLiu:
##### SUMMARY
example
```
task_pod_spec_kustomization: |
  containers:
  - name: extra-container
    image: quay.io/ansible/awx-ee:latest
    args:
    - /bin/sh
    - -c
    - |
      echo "Hello from extra-container"
      sleep 3600
```

##### ISSUE TYPE

 - Breaking Change 
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/roles/installer/tasks/resources_configuration.yml
+++ b/roles/installer/tasks/resources_configuration.yml
@@ -245,16 +245,58 @@
   set_fact:
     _redis_image: "{{ _custom_redis_image | default(lookup('env', 'RELATED_IMAGE_AWX_REDIS')) | default(_default_redis_image, true) }}"
 
-- name: Apply deployment resources
+- name: Create directory for kustomize
+  ansible.builtin.file:
+    path: "deployments"
+    state: directory
+  register: _deployment_kustomize_dir
+
+- name: Create tmp dir for {{ item }} deployment
+  tempfile:
+    state: directory
+  register: _deployment_kustomize_dir
+
+- name: Copy over kustomization.yaml to tmp dir
+  copy:
+    src: ../templates/deployments/kustomization.yaml
+    dest: "{{ _deployment_kustomize_dir.path }}"
+
+- name: Template deployments and overrides
+  template:
+    src: deployments/{{ item }}.yaml.j2
+    dest: "{{ _deployment_kustomize_dir.path }}/{{ item }}.yaml"
+  loop:
+    - "task"
+    - "task_override"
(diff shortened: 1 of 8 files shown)


TheRealHaoLiu:
cc @kurokobo @rooftopcellist @shanemcd

kurokobo:
I am interested in this but I don't know if using kustomize is the best way to achieve overriding.
Also I don't want to use massive YAML multiline strings since it ignores syntax highlighting and make it hard to read.
Will investigate similar implementation.

kurokobo:
Thanks for implementing this, I have not tested well but by using kustomize like your PR it's hard to add custom env to specific container for example.

I'm considering to use `combine` and `community.general.lists_mergeby` filter. Will try to create an  alternative PR as a PoC.

kurokobo:
My idea: https://github.com/ansible/awx-operator/pull/1697

kurokobo:
> but by using kustomize like your PR it's hard to add custom env to specific container for example

Ah ignore my comment due to lack of my understanding 😞 
Strategic Merge by Kustomize can merge dict for each containers by using `name` as merge key, so we can add env, volumeMounts, etc. to pre-deifined containers such as awx-task, redis, etc.
Kustomize way may be safer than merging by `combine` or `lists_mergeby` that implemented in my PR #1697.

So this PR is enough for most use-cases I think. It should be better if we can define `*_pod_spec_kustomization`  as dict instead of multi-line strings (Maybe I just dislike multi-line strings excessively).

roles/installer/tasks/resources_configuration.yml
```diff
@@ -245,17 +245,59 @@
...
+    src: deployments/{{ item }}.yaml.j2
+    dest: "{{ _deployment_kustomize_dir.path }}/{{ item }}.yaml"
+  loop:
+    - "task"
+    - "task_override"
+    - "web"
+    - "web_override"
+
+- name: Run kustomize build on task deployment and store in deployments.yaml
+  command: kustomize build {{ _deployment_kustomize_dir.path }} -o {{ _deployment_kustomize_dir.path }}/deployments.yaml
```
kurokobo, line 275: We can use `kubernetes.core.kustomize` lookup instead.

https://docs.ansible.com/ansible/latest/collections/kubernetes/core/kustomize_lookup.html

PR #1696, opened 2024-01-25
Fix makefile ansible-operator URL

dmzoneill:
The binary for ansible-operator v1.32.0 does not exist.

##### SUMMARY
The version hard coded 1.32.0 does not exist.
https://github.com/ansible/awx-operator/blob/devel/Makefile#L196

Only v1.31.0 is currently available
https://github.com/operator-framework/operator-sdk/releases
This PR sets it to 1.31.0 until a newer version becomes available.

https://github.com/operator-framework/operator-sdk/releases

##### ISSUE TYPE

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,7 +177,7 @@ ifeq (,$(shell which operator-sdk 2>/dev/null))
 	@{ \
 	set -e ;\
 	mkdir -p $(dir $(OPERATOR_SDK)) ;\
-	curl -sSLo $(OPERATOR_SDK) https://github.com/operator-framework/operator-sdk/releases/download/v1.32.0/operator-sdk_$(OS)_$(ARCHA) ;\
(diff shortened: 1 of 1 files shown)


kurokobo:
@dmzoneill 
Hi, simply it has been moved to under operator-framework/ansible-operator-plugins: https://github.com/operator-framework/ansible-operator-plugins/releases

dmzoneill:
@kurokobo I've updated the makefile to reference the indicated file locations.
Please review and merge if possible.

kurokobo (commented):
Hi, thanks for updating.

Current AWX Operator is based on Operator SDK **1.32.0** instead of 1.33.0 nor 1.34.0 (https://github.com/ansible/awx-operator/pull/1651):

https://github.com/ansible/awx-operator/blob/acda72558f7122b8c39e09dd85c7017d59628106/Dockerfile#L1

So if we want to fix `Makefile` I think the same version should be specified.

TheRealHaoLiu (approved):

PR #1699, opened 2024-01-31
[draft] add empytdir mount to ee container in task pod

kdelee:
##### SUMMARY

Mount a memory backed emptyDir to the ee container /tmp so receptor writes the job events its streaming back from job pods to memory instead of the container writable layer, which is copy-on-write. 

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

so far **just works** in local kind cluster
would like to test out and compare apples:apples and measure RAM usage, job event processing performance

I've built the operator container here: quay.io/kdelee/awx-operator:receptor-emptydir

Doesn't require any special awx image

### DIFF
--- a/roles/installer/templates/deployments/task.yaml.j2
+++ b/roles/installer/templates/deployments/task.yaml.j2
@@ -319,6 +319,8 @@ spec:
               mountPath: "/var/run/receptor"
             - name: "{{ ansible_operator_meta.name }}-projects"
               mountPath: "/var/lib/awx/projects"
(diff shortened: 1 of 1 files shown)


kdelee:
Wonder if there is a desire to guard this in if statement with some var -- that said, I see the proliferation of vars as also a problem :smile:

TheRealHaoLiu (changes requested):
this should be done as a `ee_extra_volume_mounts` and `task_extra_volumes` and not used as a default...

memory are 💵

Issue #1301, opened 2023-03-27
Make image location of kube-rbac-proxy and awx-manager configurable for helm install


flyinggreenfrog:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that AWX Operator is open source software provided for free and that I might not receive a timely response.

### Feature Summary

First I'm not sure, if that's the right location/way to ask for help in this issue, so please point me to the correct, if necessary.

We have an airgap environment without internet access and our own container registry. For awx installed with awx-operator all images are configurable to our own registry. But for the awx-operator itself there are hardcoded container images used.

Our main question is, how can make the images used for awx-operator configurable in combination with a helm chart install from this repo.

Our problem is basically, what was described in #916 , but for installation we use helm chart instead of kustomize.

I'm trying to figure the helm chart creation process in awx-operator to see how it can be changed to use custom values for `gcr.io/kubebuilder/kube-rbac-proxy:v0.13.0` and `quay.io/ansible/awx-operator:1.2.0` images in an airgap environment without internet access.

Maybe someone knows how to achieve our goal?

Environment:

- AWX version: 21.12.0
- Operator version: 1.2.0
- AWX install method: helm chart

juniorsemacento:
@flyinggreenfrog  I'm face the same problem.
I do have a Rancher cluster using RKE2.
All the parts are working in Airgap environment, which is RKE2, Rancher, Longhorn and Jetstack/cert-manager.

Now I do need to install awx-operator or either using helm install with a private repo, or using kustomazation.yaml file, which still not as good as helm install.

Let me know if you had any improvements on this matter.

Thanks

danielpodwysocki:
I've ran into this as well and submitted a PR to make it possible:
https://github.com/ansible/awx-operator/pull/1705

In case anybody encounters this while it's reviewing, you can checkout my branch and run `make helm-chart`. Provided you have `kustomize` and `yq` installed, it should get a version of the chart with those values exposed built for you.

klecouvey:
Hi @danielpodwysocki! Any update on your PR?

danielpodwysocki:
^ There's been no updates/feedback from the AWX team for a while.



[...]

PR #1705, opened 2024-02-06
1301 parametrize image locations

danielpodwysocki:
##### SUMMARY

This allows users of private registries to use Values.yml to configure the operator and rbac proxy images and point at their images, while keeping the default behaviour same as on the current chart.

It removes the need to constantly rebuild the chart in-house and makes it reusable across organizations.

It also includes a separate commit with a minor yq fix - a missing `-y` flag was causing it to error when testing the `helm-chart` make target.

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

Issue link:
https://github.com/ansible/awx-operator/issues/1301

Comparison of the resulting charts - ../../awx-operator is an unpacked official chart for 2.11.0:

```
➜  awx-operator git:(1301-parametrize-image-locations) diff ../../awx-operator/ charts/awx-operator
diff --color ../../awx-operator/Chart.yaml charts/awx-operator/Chart.yaml
2c2
< appVersion: 2.11.0
---
> appVersion: 0.1.0
6c6
< version: 2.11.0
---
> version: 2.11.0-5-g165d099
Common subdirectories: ../../awx-operator/crds and charts/awx-operator/crds
Common subdirectories: ../../awx-operator/templates and charts/awx-operator/templates
diff --color ../../awx-operator/values.yaml charts/awx-operator/values.yaml
0a1,10
> # images for the operator itself, not the AWX deployment
> # For airgapped installs, modify both those and the AWX.spec - you can pass the 
> # variables as defined in https://github.com/ansible/awx-operator/blob/devel/roles/installer/tasks/resources_configuration.yml
> operator_image: quay.io/ansible/awx-operator
> operator_version: latest
> 
> 
> rbac_proxy_image: gcr.io/kubebuilder/kube-rbac-proxy
> rbac_proxy_version: v0.15.0
```

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -15,7 +15,7 @@ spec:
           capabilities:
             drop:
             - "ALL"
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.15.0
+        image: "rbac_proxy"
         args:
         - "--secure-listen-address=0.0.0.0:8443"
         - "--upstream=http://127.0.0.1:8080/"
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -15,7 +15,7 @@ spec:
           capabilities:
             drop:
             - "ALL"
-        image: "rbac_proxy"
+        image: rbac_proxy
         args:
         - "--secure-listen-address=0.0.0.0:8443"
         - "--upstream=http://127.0.0.1:8080/"
--- a/.helm/starter/values.yaml
+++ b/.helm/starter/values.yaml
@@ -5,6 +5,12 @@ operator_image: quay.io/ansible/awx-operator
 operator_version: latest
 
 
+# if you are using a private registry requiring authentication, replace the default imagePullSecrets here.
+# Those need to be defined already within the namespace to which you release this helm chart.
+# Keep in mind you also need to pass those for resources you define via the operator - defined in roles/installer/defaults/main.yml
+imagePullSecrets:
+  - redhat-operators-pull-secret
+
 rbac_proxy_image: gcr.io/kubebuilder/kube-rbac-proxy
 rbac_proxy_version: v0.15.0
 
--- a/.helm/starter/values.yaml
+++ b/.helm/starter/values.yaml
@@ -1,3 +1,13 @@
+# images for the operator itself, not the AWX deployment
+# For airgapped installs, modify both those and the AWX.spec - you can pass the 
+# variables as defined in https://github.com/ansible/awx-operator/blob/devel/roles/installer/tasks/resources_configuration.yml
+operator_image: quay.io/ansible/awx-operator
+operator_version: latest
+
+
+rbac_proxy_image: gcr.io/kubebuilder/kube-rbac-proxy
+rbac_proxy_version: v0.15.0
+
 AWX:
   # enable use of awx-deploy template
   enabled: false
--- a/Makefile
+++ b/Makefile
@@ -332,8 +332,9 @@ helm-chart: helm-chart-generate
 .PHONY: helm-chart-generate
 helm-chart-generate: kustomize helm kubectl-slice yq charts
 	@echo "== KUSTOMIZE: Set image and chart label =="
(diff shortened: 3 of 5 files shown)


danielpodwysocki:
For registries requiring authentication, this also needs a method to define imagePullSecrets.

I will update this PR when I get a moment to sort that.

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -15,7 +15,7 @@ spec:
           capabilities:
             drop:
             - "ALL"
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.15.0
+        image: "rbac_proxy"
```
kurokobo: @danielpodwysocki 
Hi, I didn't make any test for this PR locally, but I think this would break _non-helm_ based installation.
This `image` param will be used as-is in both `make deploy` and `kubectl apply -k.` ways that described in _Basic install_ section and will cause `ErrImagePull` error: https://ansible.readthedocs.io/projects/awx-operator/en/latest/installation/basic-install.html
If you want to replace this `image` with `rbac_proxy`, further modifications should be made so that the default image is correctly specified for non-helm installations.

danielpodwysocki: Thank you, wasn't aware you were also handling non-helm installs there - will circle back to this soon-ish, make it work and update the PR.

rooftopcellist: I think to make this work for normal operator installs (non-helm), you would need to add an entry to this kustomization file:
* https://github.com/ansible/awx-operator/blob/0d1fa239a56c55202167cb833a72f495928442fc/config/default/kustomization.yaml

that is similar to the existing one for the operator image, here:
* https://github.com/ansible/awx-operator/blob/devel/config/manager/kustomization.yaml#L12-L17

danielpodwysocki: I have updated the PR to not break Kustomize - apologies it took a bit, life is  sometimes very good at getting in the way of things :)

Would you mind having another look? I also made imagePullSecrets customizable (while keeping the default value).

Reading the code for the operator itself, the image_pull_secrets var defined in `roles/installer/defaults/main.yml` takes care of things already there, so that shouldn't need extra handling. 

And thank you once more for having a look!

PS. quick test I did:
```
# deploying to a local dev cluster running in kind
➜  awx-operator git:(1301-parametrize-image-locations) ✗ make deploy
namespace/awx unchanged
customresourcedefinition.apiextensions.k8s.io/awxbackups.awx.ansible.com unchanged
customresourcedefinition.apiextensions.k8s.io/awxmeshingresses.awx.ansible.com unchanged
customresourcedefinition.apiextensions.k8s.io/awxrestores.awx.ansible.com unchanged
customresourcedefinition.apiextensions.k8s.io/awxs.awx.ansible.com unchanged
serviceaccount/awx-operator-controller-manager unchanged
role.rbac.authorization.k8s.io/awx-operator-awx-manager-role configured
role.rbac.authorization.k8s.io/awx-operator-leader-election-role unchanged
clusterrole.rbac.authorization.k8s.io/awx-operator-metrics-reader unchanged
clusterrole.rbac.authorization.k8s.io/awx-operator-proxy-role unchanged
rolebinding.rbac.authorization.k8s.io/awx-operator-awx-manager-rolebinding unchanged
rolebinding.rbac.authorization.k8s.io/awx-operator-leader-election-rolebinding unchanged
clusterrolebinding.rbac.authorization.k8s.io/awx-operator-proxy-rolebinding unchanged
configmap/awx-operator-awx-manager-config unchanged
service/awx-operator-controller-manager-metrics-service unchanged
deployment.apps/awx-operator-controller-manager unchanged
➜  awx-operator git:(1301-parametrize-image-locations) ✗ kubectl describe deployments.apps -n awx awx-operator-controller-manager  | grep -i image
    Image:      gcr.io/kubebuilder/kube-rbac-proxy:v0.15.0
    Image:      quay.io/ansible/awx-operator:2.11.0-7-g3d080f0
```

danielpodwysocki:
This is a bit old and looks like it has lost traction.

Feel free to grab those changes and resubmit them if they are useful - if anyone stumbles into here from Google, I've shared how I'd handle the airgap outside the official operator in https://github.com/ansible/awx-operator/issues/1301

PR #1690, opened 2024-01-22
add 'customSecrets' and 'customVolumes' values to Helm Chart to simplifies the creation of ressources for PoC

sbilque:
##### SUMMARY
This PR allows deploying secrets and persistentvolumes resources required by AWX in the cluster when using the Helm Chart for PoC or testing. 

Resources are defined in the `customSecrets` and `customVolumes` values. 

Compared to the `extraDeploy` value, you don't need to know any special syntax to describe resources.

##### ISSUE TYPE
- New or Enhanced Feature

Example usage in a values.yml file

````yaml
AWX:
  # enable use of awx-deploy template
  ...

  spec:
     projects_persistence: true

  # configurations for external postgres instance
  postgres:
    enabled: false
    ...

customSecrets:
  enabled: true
  admin:
    enabled: true
    password: mysuperlongpassword
  secretKey:
    enabled: true
    key: supersecuresecretkey
    secretName: my-awx-secret-key
  ingressTls:
    enabled: true
    selfSignedCert: false
    key: fakekey 
    certificate: fakecertificate
  bundleCacert:
    enabled: true
    crt: <contentofmybundlecacrt>
  cpPullCredentials:
    enabled: false
    dockerconfig:
      - registry: docker.io
        username: user
        password: pwd

customVolumes:
  postgres:
    enabled: true
    hostPath: /data/postgres-13
  projects:
   enabled: true
   hostPath: /data/projects
````

Example usage with `kubernetes.core.helm` Ansible module: 

````yaml
- name: Deploy the AWX instance
  kubernetes.core.helm:
    ...  
    values_files: 
      AWX:
        spec:
           projects_persistence: true
      customSecrets:
        enabled: true
        admin:
          enabled: true
          password: !vault |
            $ANSIBLE_VAULT;1.2;AES256
            346534353436643039616632653135656534
        secretKey:
          enabled: true
          password: !vault |
            $ANSIBLE_VAULT;1.2;AES256
            346534353436643039616632653135656534
          secretName: my-awx-secret-key
        ingressTls:
          enabled: true
          selfSignedCert: false
          key: fakekey 
          certificate: fakecertificate
        bundleCacert:
          enabled: true
          crt: "{{ lookup('ansible.builtin.file', 'ca-certs.crt') }}"
        ldapCacert:
          enabled: true
          crt: "{{ lookup('ansible.builtin.file', 'ldap-certs.crt') }}"
      customVolumes:
        postgres:
        enabled: true
          hostPath: /data/postgres-13
[...]

PR #1712, opened 2024-02-14
Tolerances and Affinity addition to Helm Chart

JonTheNiceGuy:
##### SUMMARY

This change introduces the ability to define tolerances and affinity to the AWX Operator.

The cluster on which I'm working is highly defined with taints to ensure workload placement, and the current helm chart does not allow it to be deployed. While adding support for tolerances, the other common placement option - affinity - was added at the same time.

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
When the current helm chart is deployed, on a cluster where all nodes have a taint (e.g. `xl:NoSchedule=true` or `clusterservices:NoSchedule=true`) the operator will go into a permanent "pending" state and in the events it will show:
```
  Type     Reason             Age                     From                Message
  ----     ------             ----                    ----                -------
  Warning  FailedScheduling   64s (x336 over 28h)     default-scheduler   0/6 nodes are available: 3 node(s) had untolerated taint {xl: true}, 3 node(s) had untolerated taint {clusterservices: true}. preemption: 0/6 nodes are available: 6 Preemption is not helpful for scheduling..
```

Adding toleration to the helm chart and implementing it will allow it to be scheduled.

I was unable to test this change with molecule as I don't have molecule installed or configured and the documentation for the operator points to an out-of-date link for molecule install guides, and the guides on ansible.com for molecule show an error about a driver not being configured. I did test the resulting helm chart on my cluster to show the tolerations were applied. I am happy to work with developers and contributors if any additional testing is required.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -375,6 +375,8 @@ helm-chart-generate: kustomize helm kubectl-slice yq charts
 	for file in $${cluster_scoped_files}; do\
 		$(YQ) -i '.metadata.name += "-{{ .Release.Name }}"' $${file};\
 	done
+	# Add tolerations and affinity to the deployment
+	$(SED_I) -E -e 's/^(      containers:)/{{- with .Values.tolerations }}\n      tolerations:\n{{- toYaml . | nindent 8 }}\n{{- end }}\n{{- with .Values.affinity }}\n      affinity:\n{{- toYaml . | nindent 8 }}\n{{- end }}\n\1/' charts/$(CHART_NAME)/raw-files/deployment-awx-operator-controller-manager.yaml
 
 	# Correct the reference for the clusterrolebinding
 	$(YQ) -i '.roleRef.name += "-{{ .Release.Name }}"' 'charts/$(CHART_NAME)/raw-files/clusterrolebinding-awx-operator-proxy-rolebinding.yaml'
--- a/.helm/starter/values.yaml
+++ b/.helm/starter/values.yaml
@@ -17,3 +17,6 @@ AWX:
     password: Unset
     sslmode: prefer
     type: unmanaged
+
+tolerations: []
+affinity: {}


Makefile
```diff
@@ -375,6 +375,8 @@ helm-chart-generate: kustomize helm kubectl-slice yq charts
 	for file in $${cluster_scoped_files}; do\
 		$(YQ) -i '.metadata.name += "-{{ .Release.Name }}"' $${file};\
 	done
+	# Add tolerations and affinity to the deployment
+	$(SED_I) -E -e 's/^(      containers:)/{{- with .Values.tolerations }}\n      tolerations:\n{{- toYaml . | nindent 8 }}\n{{- end }}\n{{- with .Values.affinity }}\n      affinity:\n{{- toYaml . | nindent 8 }}\n{{- end }}\n\1/' charts/$(CHART_NAME)/raw-files/deployment-awx-operator-controller-manager.yaml
```
rooftopcellist, line 379: I don't see this file in my checkout of the awx-operator..

rooftopcellist:
@JonTheNiceGuy if you need to add tolerations and pod_affinity to the awx application deployments, is there a reason why you wouldn't just do so on the AWX.spec? 

Docs here: https://github.com/ansible/awx-operator/tree/devel/.helm/starter#awx

---------------
But it seems like you are trying to add them to the awx-operator manager pod itself.  The typically way to accomplish this would be with kustomize.  But if you want to do it with helm, I think you would want to follow the pattern used here:
* https://github.com/ansible/awx-operator/pull/1690/files

Rather than the sed approach in this PR.  I am not as familiar with helm myself.  So I may just be misunderstanding.  Could you explain your approach a bit more if so?

Makefile
JonTheNiceGuy: This file is created by [this command](https://github.com/ansible/awx-operator/blob/ec8c20fa8821bbb2640361b21f845e52685c9177/Makefile#L355-L358) in the Makefile. The source of that file is [config/manager/manager.yml](https://github.com/ansible/awx-operator/blob/ec8c20fa8821bbb2640361b21f845e52685c9177/config/manager/manager.yaml) but it split apart by `kustomize`.

JonTheNiceGuy:
Thanks for taking the time to look at my PR @rooftopcellist. I'll try to answer as best I can (I'm also not 100% on helm, but I'm doing my best!)

> if you need to add tolerations and pod_affinity to the awx application deployments, is there a reason why you wouldn't just do so on the AWX.spec?

Because the AWX.spec needs to be operated on by *something*, which is (as far as I can make out from my tests) is actually another pod called controller-manager. If you don't also have tolerations on that pod, and all your nodes are tainted, it can't ever start, which means the AWX spec isn't actioned into your cluster.

> The typically way to accomplish this would be with kustomize.

I would agree, except I needed to inject the helm-specific loop `{{- with .Values.tolerations }}\ntolerations:\n{{- toYaml . | nindent 8 }}\n{{- end }}` that I couldn't determine would work with Kustomize. Given that `$(SED_I)` is already defined in the Makefile, I reasoned that it's probably in use somewhere else, so why not use it. That said, if a kustomize guru can fix my assumption, then I'm more than happy to swap out my hack with a better one :smile: 

> But if you want to do it with helm, I think you would want to follow the pattern used here: https://github.com/ansible/awx-operator/pull/1690/files

See my previous comment about the AWX.spec.

You also mentioned the file in question not appearing in the check-out; this is because the Makefile actually uses [`kustomize`](https://github.com/ansible/awx-operator/blob/ec8c20fa8821bbb2640361b21f845e52685c9177/Makefile#L355-L358) to split [the manifest files](https://github.com/ansible/awx-operator/blob/ec8c20fa8821bbb2640361b21f845e52685c9177/config/manager/manager.yaml) into the parts needed for the helm chart. 

I hope this helps. Please do let me know if you need any more clarification on my approach :smile:

PR #1725, opened 2024-02-19
docs: enable search feature

kurokobo:
##### SUMMARY

This PR enables search feature on the docs page:

image

image

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

The plugin configuration for mkdocs is based on the [ansible-lint repo](https://github.com/ansible/ansible-lint/blob/main/mkdocs.yml).

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -18,6 +18,9 @@ theme:
     - navigation.indexes
     - navigation.tracking
     - toc.integrate
+    - search.highlight
+    - search.share
+    - search.suggest
   palette:
     - media: "(prefers-color-scheme: light)"
       primary: teal
@@ -92,7 +95,8 @@ exclude_docs:
 plugins:
   - autorefs
   - markdown-exec
-  - search
+  - material/search:
+      separator: '[\s\-,:!=\[\]()"`/]+|\.(?!\d)|&[lg]t;|(?!\b)(?=[A-Z][a-z])'
   - mkdocstrings:
       handlers:
         python:


mkdocs.yml
```diff
@@ -92,7 +95,8 @@ exclude_docs:
 plugins:
   - autorefs
   - markdown-exec
-  - search
+  - material/search:
+      separator: '[\s\-,:!=\[\]()"`/]+|\.(?!\d)|&[lg]t;|(?!\b)(?=[A-Z][a-z])'
```
fosterseth, line 99: can you explain what this part is? did you get this from the plugin docs?

kurokobo: Thanks for your review and sorry for lack of description.

Both two lines are came from [ansible/mkdocs-ansible](https://github.com/ansible/mkdocs-ansible/blob/538fc3de94a759620dc22ff6302ad4fe440d7837/mkdocs.yml#L52-L53), and almost the same regex for `separator` is described in the plugin docs: <https://squidfunk.github.io/mkdocs-material/plugins/search/#config.separator> (The regex for the separator is from in the ansible repo, although it has the `` ` `` added to the one in the documentation and in a different order)

> `material/search:`

This is the line to enable serch feature by using built-in `search` plugin of `material` theme. Since we use `ansible` theme, we have to append prefix `material/` (in the other words use [fully-qualified plugin name](https://github.com/squidfunk/mkdocs-material/blob/59f4147e6026ca02d08de2935d9036947e1e4b68/pyproject.toml#L77-L85)) to call plugins from different theme. The `ansible` theme is made by extending `material` theme, so any plugins from `material` works perfectly.

> ``separator: '[\s\-,:!=\[\]()"`/]+|\.(?!\d)|&[lg]t;|(?!\b)(?=[A-Z][a-z])'``

As documented, this is a known hack to improve indexing for search feature for technical docs. This complex regex is similar to `A|B|C|D` and each of them means;

- A: ``[\s\-,:!=\[\]()"`/]+``: List of characters that determined as word separator.
- B: `\.(?!\d)`: Prevent version string to be separated by `dot`, means to make `1.2.3` indexable by exact `1.2.3` instead of separated `1`, `2`, and `3`
- C: `&[lg]t;` To separate words using `<` and `>`, such as HTML/XML tags.
- D: `(?!\b)(?=[A-Z][a-z])`: Separator for `PascalCase` or `camelCase`, like making `PascalCase` indexable by two separated words; `Pascal` and `Case`.

kurokobo:
rebased

oraNod (approved):
Looks great. Thanks so much @kurokobo :tada: 

I've deployed this on my fork and it works as expected: https://oranod-awx-operator.readthedocs.io/en/latest/

fosterseth (approved):

mkdocs.yml
fosterseth: thanks for the explanation!

PR #1728, opened 2024-02-24
Add container level security context for task and web deployments

### LINKED ISSUE #383: K8S securityContext  / runAsNonRoot
We have a policy of not running any containers as root. 
By default it looks like all awx and redis containers is starting as root and will fail/not start in our cluster do to default PSP in cluster. 

If i manually update deployment manifest with SecurityContext for container is looks like its working...
```
containers:
  - name: redis
    securityContext:
      runAsUser: 999
...
  - name: awx-ee
    securityContext:
      runAsUser: 1000
...
  - name: awx-web
    securityContext:
      runAsUser: 1000
      runAsGroup: 0
...
  - name: awx-task
    securityContext:
      runAsUser: 1000
      runAsGroup: 0
```

Running as root is general not a recommended practice. 

I also had to manually create file in container awx-task.:
```
bash-4.4$ touch /var/lib/awx/rsyslog/rsyslog.conf
```
log-output before fix.
Somehow this file was not generated.
```
rsyslogd: could not open config file '/var/lib/awx/rsyslog/rsyslog.conf': No such file or directory [v8.1911.0-7.el8 try https://www.rsyslog.com/e/2104 ]
rsyslogd: run failed with error -2104 (see rsyslog.h or try https://www.rsyslog.com/e/2104 to learn what that number means)
```

What about adding a SecurityContext or a general Spec in the crds to override this?
Or sync dockerimage + deployment with correct non-root user to allow read-write where its needed.

### LINKED ISSUE #571: Installation fails when PodSecurityPolicy applied with readOnlyRootFilesystem set to true
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Operator and AWX Installation fails, when pod security policy is applied prohibiting pods running as root user, and write access to filesystem. 

```
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
metadata:
  name: restricted
spec:
  allowPrivilegeEscalation: false
  defaultAllowPrivilegeEscalation: false
  fsGroup:
    rule: MustRunAs
    ranges:
    - min: 1000
      max: 65536
  hostPorts:
  - min: 0
    max: 0
  hostPID: false
  hostIPC: false
  hostNetwork: false
  privileged: false
  readOnlyRootFilesystem: true
  requiredDropCapabilities:
    - ALL
  runAsUser:
    rule: MustRunAsNonRoot
  seLinux:
    rule: RunAsAny
  supplementalGroups:
    rule: MustRunAs
    ranges:
    - min: 1000
      max: 65536
  volumes:
  - 'emptyDir'
  - 'persistentVolumeClaim'
  - 'secret'
  - 'downwardAPI'
  - 'configMap'
  - 'projected'
  - 'cephFS
```
 

##### ENVIRONMENT
* AWX version: 19.3.0
* Operator version: 0.13.0
* Kubernetes version: v1.21.2
* AWX install method: minikube on linux

##### STEPS TO REPRODUCE

Apply PSP policy to the cluster and follow standard installation procedure

##### EXPECTED RESULTS

It is a good security practice run pods as non root user, without write access to filesystem. AWX & Operator should be able cope with such requirements. 

##### ADDITIONAL INFORMATION
Changes to default deployment.yaml.j2 file are needed to overwrite containers securityContext and run as non root user. It would be great to have an option in CRD to set default securityContext for all containers.

Changes applied to deployment.yaml.j2 allowing containers run with restrictions:
 ```
     ..
      initContainers:
         - name: init
          securityContext:
            runAsUser: 1000
       ..  
          name: redis
          securityContext:
            runAsUser: 999
       ..
          name: '{{ meta.name }}-web'
          securityContext:
            runAsUser: 1000
       ..
          name: '{{ meta.name }}-task'
          securityContext:
            runAsUser: 1000
       ..
          name: '{{ meta.name }}-ee'
          securityContext:
            runAsUser: 1000
```
Containers need numerous directories being writeable, possible fix is to attach them by default as emptyDir volumes:
```
web_extra_volume_mounts: |
  - name: web-tmp
     mountPath: /tmp
  - name: web-tower-logs
     mountPath: /var/log/tower
  - name: web-awx-public
     mountPath: /var/lib/awx/public
task_extra_volume_mounts: |
  - name: task-tmp
     mountPath: /tmp
  - name: task-tower-logs
     mountPath: /var/log/tower
  - name: ee-tmp
     mountPath: /tmp
ee_extra_volume_mounts: |
  - name: ee-awx-ansible
     mountPath: /var/lib/awx/.ansible
  - name: ee-home-runner-ansible
     mountPath: /home/runner/.ansible
```
Containers are modifying content of /etc/passwd files when UID is higher than 500. Without correct user mapping application would fail. Wouldn't be a better option to template passwd files and attach them as a configMap to each container?

This is part of a script failing when write restrictions are applied
```
/usr/bin/launch_awx.sh
#!/usr/bin/env bash
if [ `id -u` -ge 500 ]; then
    echo "awx:x:`id -u`:`id -g`:,,,:/var/lib/awx:/bin/bash" >> /tmp/passwd
    cat /tmp/passwd > /etc/passwd
    rm /tmp/passwd
fi
```
config-watcher (web & task) is trying write to  /var/lib/awx/.configsha and fails to start as directory is non writable.  It would be better to write into temporary directory instead.
```
cat /usr/bin/config-watcher
27 def main():
28     settings_file = "/etc/tower/settings.py"
29     hash_file = "/var/lib/awx/.configsha"
```
Supervisord on web & task container have a setting for uwsgi for writing a pipe to /var/lib/awx/awxfifo, again due to read only filesystem this would fail breaking functionality. Can this be moved to temporary directory i.e /tmp

```
/etc/supervisord.conf & /etc/supervisord_task.conf 
[program:uwsgi]
command = /var/lib/awx/venv/awx/bin/uwsgi --socket 127.0.0.1:8050 --module=awx.wsgi:application --vacuum --processes=5 --harakiri=120 --no-orphans --master --max-requests=1000 --master-fifo=/var/lib/awx/awxfifo --lazy-apps -b 32768
directory = /var/lib/awx
```
A worker container ( automation-job-* ) fails due to inability changing /etc/passwd file when restrictions are in place. This is being done by a script /usr/bin/entrypoint. Again having passwd file created accordingly before container creation would fix the problem.

Related issues:
#560 
#383

### LINKED ISSUE #890: security_context_settings won't work
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

In our network we have Open Policy Agent running which forces us to use non root containers. So we need to use non root containers or setting a custom security_context_settings parameter which overwrites the default user. 
In the following commit https://github.com/ansible/awx-operator/commit/138964f7ab6089628ecc3ab38699403b5a55cb7f and PR https://github.com/ansible/awx-operator/pull/645
i can see that there is a way to specify a custom runAsUser and runAsGroup variable 

example code from above:

```
security_context_settings:
  runAsUser: 1000
  runAsGroup: 1000
```

If i set these settings in my yaml op the awx-operator, it won't work
```
9m36s       Normal    Pulled                   pod/awx-postgres-0          Successfully pulled image "<private_registry>/centos/centos:stream8" in 15.126499374s
7m29s       Warning   Failed                   pod/awx-postgres-0          Error: container has runAsNonRoot and image will run as root (pod: "awx-postgres-0_awx(c2f22b67-a6eb-4a32-b35a-61ca7cc9905b)", container: database-check)
```

##### ENVIRONMENT
* AWX version: devel
* Operator version: devel
* Kubernetes version: v1.20.11-eks-f17b81
* AWX install method: kubernetes on EKS

##### STEPS TO REPRODUCE

Deploying the awx operator (based on the minikube steps)

```
---
kind: AWX
metadata:
  name: awx
  namespace: awx
spec:
  service_type: nodeport
  postgres_image: <private_registry>/bitnami/postgresql
  postgres_image_version: "14.2.0"
  init_container_image: <private_registry>/centos/centos
  init_container_image_version: "stream8"
  redis_image: <private_registry>/bitnami/redis
  redis_image_version: "latest"
  task_privileged: false
  image: <private_registry>/ansible/awx-operator
  image_version: "devel"
  ee_images:
    - name: awx-ee
      image: <private_registry>/ansible/awx-ee:0.6.0
  security_context_settings:
    runAsUser: 1000
    runAsGroup: 1000

```

##### EXPECTED RESULTS

That the containers that the operator uses (such as the database-check) will use the custom runAsUser and runAsGroup variables 

##### ACTUAL RESULTS

It won't work beceause the containers still starts as root while i specify a custom user in the security_context_settings settings.
related issue: https://github.com/ansible/awx-operator/issues/383

### LINKED ISSUE #1413: Web Containers Missing Security Context

gdasson:
##### SUMMARY
The security context settings offered today only provide the option to set pod level security context for web and task deployments. This PR adds the option to allow container level security context for all of the containers under web and task deployments.

fixes: #1413
fixes: #890
fixes: #571
fixes: #383

**This change doesn't dictate the values and let the users decide and configure the values on need basis. This makes it a safer approach to implement without breaking any functionality**

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
Two of the existing variable settings will become irrelevant after this change: 
* `redis_capabilities` can be covered under `redis_security_context_settings` after this change
* `task_privileged` can be covered under `task_security_context_settings` after this change

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -47,9 +47,13 @@ spec:
                 description: apiVersion of the deployment type
                 type: string
               task_privileged:
-                description: If a privileged security context should be enabled
+                description: (Deprecated) If a privileged security context should be enabled
                 type: boolean
                 default: false
+              task_security_context:
+                description: Key/values that will be set under the container-level securityContext field
+                type: object
+                x-kubernetes-preserve-unknown-fields: true
               admin_user:
                 description: Username to use for the admin account
                 type: string
@@ -1715,12 +1719,20 @@ spec:
               ee_extra_volume_mounts:
                 description: Specify volume mounts to be added to Execution container
                 type: string
+              ee_security_context:
+                description: Key/values that will be set under the container-level securityContext field
+                type: object
+                x-kubernetes-preserve-unknown-fields: true
               task_extra_volume_mounts:
                 description: Specify volume mounts to be added to Task container
                 type: string
               web_extra_volume_mounts:
                 description: Specify volume mounts to be added to the Web container
                 type: string
+              web_security_context:
+                description: Key/values that will be set under the container-level securityContext field
+                type: object
+                x-kubernetes-preserve-unknown-fields: true
               postgres_extra_volume_mounts:
                 description: Specify volume mounts to be added to Postgres container
                 type: string
@@ -1745,6 +1757,10 @@ spec:
               rsyslog_extra_volume_mounts:
                 description: Specify volume mounts to be added to the Rsyslog container
                 type: string
+              rsyslog_security_context:
+                description: Key/values that will be set under the container-level securityContext field
+                type: object
+                x-kubernetes-preserve-unknown-fields: true
               redis_image:
                 description: Registry path to the redis container to use
                 type: string
@@ -1752,10 +1768,14 @@ spec:
                 description: Redis container image version to use
                 type: string
               redis_capabilities:
-                description: Redis container capabilities
+                description: (Deprecated) Redis container capabilities
                 type: array
                 items:
                   type: string
+              redis_security_context:
+                description: Key/values that will be set under the container-level securityContext field
+                type: object
+                x-kubernetes-preserve-unknown-fields: true
               init_container_image:
                 description: Registry path to the init container to use
                 type: string
@@ -1768,6 +1788,10 @@ spec:
               init_container_extra_volume_mounts:
                 description: Specify volume mounts to be added to the init container
                 type: string
+              init_security_context:
+                description: Key/values that will be set under the container-level securityContext field
+                type: object
+                x-kubernetes-preserve-unknown-fields: true
               init_projects_container_image:
                 description: Registry path to the init projects container to use
                 type: string
@@ -1902,7 +1926,11 @@ spec:
                 description: Set session cookie secure mode for web
                 type: string
               postgres_security_context_settings:
-                description: Key/values that will be set under the pod-level securityContext field
+                description: (Deprecated) Key/values that will be set under the pod-level securityContext field
+                type: object
+                x-kubernetes-preserve-unknown-fields: true
+              postgres_security_context:
+                description: Key/values that will be set under the container-level securityContext field
                 type: object
                 x-kubernetes-preserve-unknown-fields: true
               receptor_log_level:
--- a/config/manifests/bases/awx-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/awx-operator.clusterserviceversion.yaml
@@ -134,8 +134,13 @@ spec:
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:com.tectonic.ui:hidden
-      - displayName: PostgreSQL Security Context Settings
+      - displayName: PostgreSQL Security Context Settings (Deprecated)
         path: postgres_security_context_settings
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:advanced
+        - urn:alm:descriptor:com.tectonic.ui:hidden
+      - displayName: PostgreSQL Security Context
+        path: postgres_security_context
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:com.tectonic.ui:hidden
@@ -486,6 +491,11 @@ spec:
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:com.tectonic.ui:resourceRequirements
+      - displayName: Web Security Context Settings
+        path: web_security_context
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:advanced
+        - urn:alm:descriptor:com.tectonic.ui:hidden 
       - displayName: Task Container Resource Requirements
         path: task_resource_requirements
         x-descriptors:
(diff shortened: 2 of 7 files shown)


rooftopcellist:
@gdasson Looks like this PR follows the plan laid out in this comment:
* https://github.com/ansible/awx-operator/pull/1425#issuecomment-1854578532
Continuing the work there.  

Changes suggested before merge:
* I think `_settings` at the end of each of these is redundant and can be removed, what do you think?  Otherwise, I think this approach will work for v1beta1. 
* Unfortunately, we have to provide backwards compatibility for the existing securityContext settings,
  * Example of how to fix for task_privileged: [here](https://github.com/ansible/awx-operator/pull/1425/files#diff-4d0fa82fa796130ee45f174f37b43ce5faf7092d12717ad484dbdf712ff24988R200-R208)
  * Example of how to fix for redis_capabilities: [here](https://github.com/ansible/awx-operator/pull/1425/files#diff-4d0fa82fa796130ee45f174f37b43ce5faf7092d12717ad484dbdf712ff24988R157-R165)
  * Same for `postgres_security_context_settings`
* Can you also mark the task_privileged, postgres_security_context_settings and redis_capabilities parameters as Deprecated? ([for example](https://github.com/ansible/awx-operator/blob/devel/config/crd/bases/awx.ansible.com_awxs.yaml#L127))

> Note: We may re-write this to be settings nested under each component when we make a v2 CRD schema, but that is down the line and it would be good to give users a way to configure this now.

gdasson:
@rooftopcellist : Thanks for the review and feedback. I have now incorporated your review comments. Thanks.

gdasson:
@rooftopcellist or other reviewers : Any updates on this please?


cc: @bharathdevopz

djyasin:
Hello @gdasson,
Would you mind rebasing this and resolving any conflicts?

maarous:
Hello, Any updates on this please?

Doenerpapst:
Hello, any updates on this please? Regarding the current security vulnerability https://nvd.nist.gov/vuln/detail/CVE-2024-53899 this would be a useful addition

dominykasn:
Hi, any updates?

gdasson:
@djyasin : Rebased and fixed merge conflicts. Thanks.

PR #1740, opened 2024-03-01
Bind EE images version with DEFAULT_AWX_VERSION

### LINKED ISSUE #1484: control_plane_ee_image still uses "latest" tag in release

### Bug Summary

While `image_version` is based on `DEFAULT_AWX_VERSION` environment variable that gets injected during staging process, I have found out, that `control_plane_ee_image` via `_control_plane_ee_image` is still hardcoded to `awx-ee:latest`.
This can cause unexpected behaviour, when used with default EE (which is also `awx-ee:latest`), as the default EE seems to be pulled with each job run when there is an update to that tag.
After such unaware change of `awx-ee:latest` image, when the `awx-task` pod terminates for whatever reason, it will start with never code as intended.

As now finally awx-ee images are tagged with tags matching awx, I would recommend to use the same tag on ee as well.

For me, as a workaround, i use following:
```
spec:
...
  init_container_image: "{{ _image }}-ee"
  init_container_image_version: "{{ _image_version }}"
  control_plane_ee_image: "{{ _image }}-ee:{{ _image_version }}"
...
```
but this should not be needed if implemented properly.

### AWX Operator version

2.3.0

### AWX version

22.4.0

### Kubernetes platform

other (please specify in additional information)

k3d 5.4.3, kubernetes version v1.23.6+k3s1

### Modifications

no

### Steps to reproduce

* install default k3d
* wait for `awx-ee:latest` tag to move
* run any job, that will fetch updated `awx-ee:latest`
* delete `awx-task` pod
* new awx-task will have changed controlplane image

### Expected results

After installation of AWX via awx-operator, no image should change automatically until deployment of new awx-operator version

### Actual results

awx-ee and control_plane images might get changed when latest tag moves

### Additional information

_No response_

### Operator Logs

_No response_

bartowl:
##### SUMMARY

This PR fixes #1484 by binding following images to `DEFAULT_AWX_VERSION` instead of current `latest` tag:
- `ee_images` - default execution environment (will add new EE with each upgrade, leaving previous as-is)
- `control_plane_ee_image` - will switch it to new version with each update
- `init_container_image` - will be also switched to new version with each update

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

Currently those were linked with `latest` version of images which might resulted in unexpected effects for ee_image once the `latest` tag was updated.
For `control_plane_ee_image` and `init_container_image` - they were not changed until container re-create, but should this happen they would also be updated (possibly unaware/unexpected)

In order to prevent switching to updated image versions on upgrade it is possible to define `control_plane_ee_image` and (AFAIK undocumented) `init_container_image`/`init_container_image_version` in AWX resource spec.

New EE image will always be added upon upgrade, but it will be clearly marked with AWX version.

Additionally it also makes small fix for 'or' in check of `init_container_image_version` presence which could leave to errors when init_container_image_version would not be specified and `init_container_image` be defined (separate commit)

```

```

### DIFF
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -278,10 +278,10 @@ init_container_extra_commands: ''
 init_container_extra_volume_mounts: ''
 
 ee_images:
-  - name: AWX EE (latest)
-    image: quay.io/ansible/awx-ee:latest
+  - name: "AWX EE ({{ _image_version }})"
+    image: "quay.io/ansible/awx-ee:{{ _image_version }}"
 
-_control_plane_ee_image: quay.io/ansible/awx-ee:latest
+_control_plane_ee_image: "quay.io/ansible/awx-ee:{{ _image_version }}"
 
 _init_container_image: "{{ _control_plane_ee_image.split(':')[0] }}"
 _init_container_image_version: "{{ _control_plane_ee_image.split(':')[1] }}"
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -278,6 +278,8 @@ init_container_extra_commands: ''
 init_container_extra_volume_mounts: ''
 
 ee_images:
+  - name: "AWX EE (latest)"
+    image: "quay.io/ansible/awx-ee:latest"
   - name: "AWX EE ({{ _image_version }})"
     image: "quay.io/ansible/awx-ee:{{ _image_version }}"
 
--- a/docs/user-guide/advanced-configuration/deploying-a-specific-version-of-awx.md
+++ b/docs/user-guide/advanced-configuration/deploying-a-specific-version-of-awx.md
@@ -2,15 +2,15 @@
 
 There are a few variables that are customizable for awx the image management.
 
-| Name                | Description               | Default                                 |
-| ------------------- | ------------------------- | --------------------------------------  |
-| image               | Path of the image to pull | quay.io/ansible/awx                     |
-| image_version       | Image version to pull     | value of DEFAULT_AWX_VERSION or latest  |
-| image_pull_policy   | The pull policy to adopt  | IfNotPresent                            |
-| image_pull_secrets  | The pull secrets to use   | None                                    |
-| ee_images           | A list of EEs to register | quay.io/ansible/awx-ee:latest           |
-| redis_image         | Path of the image to pull | docker.io/redis                         |
-| redis_image_version | Image version to pull     | latest                                  |
+| Name                | Description               | Default                                    |
+| ------------------- | ------------------------- | -----------------------------------------  |
+| image               | Path of the image to pull | quay.io/ansible/awx                        |
+| image_version       | Image version to pull     | value of DEFAULT_AWX_VERSION or latest     |
+| image_pull_policy   | The pull policy to adopt  | IfNotPresent                               |
+| image_pull_secrets  | The pull secrets to use   | None                                       |
+| ee_images           | A list of EEs to register | quay.io/ansible/awx-ee:DEFAULT_AWX_VERSION |
+| redis_image         | Path of the image to pull | docker.io/redis                            |
+| redis_image_version | Image version to pull     | latest                                     |
 
 Example of customization could be:
 
--- a/roles/installer/tasks/set_images.yml
+++ b/roles/installer/tasks/set_images.yml
@@ -8,8 +8,8 @@
   set_fact:
     _custom_init_container_image: "{{ init_container_image }}:{{ init_container_image_version }}"
   when:
-    - init_container_image | default([]) | length
-    - init_container_image_version is defined or init_container_image_version != ''
+    - init_container_image | default('_undefined',true) != '_undefined'
+    - init_container_image_version | default('_undefined',true) != '_undefined'
 
 - name: Set Init image URL
   set_fact:
--- a/roles/mesh_ingress/defaults/main.yml
+++ b/roles/mesh_ingress/defaults/main.yml
@@ -9,7 +9,7 @@ ingress_controller: ''
 
 set_self_owneref: true
 
-_control_plane_ee_image: quay.io/ansible/awx-ee:latest
+_control_plane_ee_image: "quay.io/ansible/awx-ee:{{ lookup('env', 'DEFAULT_AWX_VERSION') or 'latest' }}"
 _image_pull_policy: Always
 image_pull_secrets: []
 


fosterseth:
@bartowl thanks for this PR! will pull down and test it out soon

TheRealHaoLiu:
https://github.com/ansible/awx/blob/devel/.github/workflows/stage.yml#L109-L115
set the `DEFAULT_AWX_VERSION` as build arg

https://github.com/ansible/awx-operator/blob/devel/Dockerfile#L9C1-L12C41
takes the version and set `DEFAULT_AWX_VERSION` as a env var in the container 

https://github.com/ansible/awx-operator/blob/devel/roles/installer/defaults/main.yml#L256
set default of `_image_version` to `DEFAULT_AWX_VERSION`

and this PR takes that and uses it, that looks correct

TheRealHaoLiu (dismissed):
https://github.com/ansible/awx-operator/blob/devel/roles/mesh_ingress/defaults/main.yml#L12
one missing place to fix

TheRealHaoLiu:
now that i committed to this PR i shouldn't be the approver anymore 🤣  @fosterseth double check my work plz

TheRealHaoLiu:
**test process** 
- build and deploy to openshift with 
```
IMAGE_TAG_BASE=quay.io/haoliu/awx-operator PLATFORMS=linux/amd64 BUILD_ARGS="--build-arg DEFAULT_AWX_VERSION=23.9.0" make docker-buildx deploy NAMESPACE=awx-test
```
- deploy awx with 
```
kind: AWX
metadata:
  name: awx
spec:
  ingress_type: Route
```
- verify awx-ee in awx-task deployment
```
oc get deployment awx-task -o yaml | oc neat | grep quay.io/ansible/awx-ee
```
result
```
        image: quay.io/ansible/awx-ee:23.9.0 <- init container
        image: quay.io/ansible/awx-ee:23.9.0 <- controlplane-ee
```
- deploy awxmeshingress with 
```
---
apiVersion: awx.ansible.com/v1alpha1
kind: AWXMeshIngress
metadata:
  name: awx-mesh-ingress
spec:
  deployment_name: awx
```
... and hit a bug that i created somewhere else...regret my decision to try to do something at 5PM...
quickly fixed the stupid bug and submitted https://github.com/ansible/awx-operator/pull/1753 ...
walked away and fired up my grill to get ready to cook dinner...
- confirm mesh ingress deployment with 
```
oc get deployment awx-mesh-ingress -o yaml | oc neat | grep quay.io/ansible/awx-ee
        image: quay.io/ansible/awx-ee:23.9.0
```

TheRealHaoLiu:
and realize that i should probably do a upgrade test...
basically the same thing except im gonna do 
```
IMAGE_TAG_BASE=quay.io/haoliu/awx-operator PLATFORMS=linux/amd64 BUILD_ARGS="--build-arg DEFAULT_AWX_VERSION=23.8.1" make docker-buildx deploy NAMESPACE=awx-test
```
than deploy awx while waiting for my grill to come up to temp
```
IMAGE_TAG_BASE=quay.io/haoliu/awx-operator PLATFORMS=linux/amd64 BUILD_ARGS="--build-arg DEFAULT_AWX_VERSION=23.9.0" make docker-buildx deploy NAMESPACE=awx-test
```

verify that we don't create duplicate default ee in awx...(we probably shouldn't)

TheRealHaoLiu:
ok already notice the difference... this will change Execution Environment in AWX EE 
image

I'm not sure if how i feel about this behavioral change...

TheRealHaoLiu:
after the upgrade 
image

job execution will use the "latest" if no execution environment is defined so... i guess its ok?

brb dropping the steak on the grill...

TheRealHaoLiu:
@shanemcd what do you think about this behavioral change?

roles/installer/defaults/main.yml
```diff
@@ -278,10 +278,10 @@ init_container_extra_commands: ''
 init_container_extra_volume_mounts: ''
 
 ee_images:
-  - name: AWX EE (latest)
-    image: quay.io/ansible/awx-ee:latest
+  - name: "AWX EE ({{ _image_version }})"
+    image: "quay.io/ansible/awx-ee:{{ _image_version }}"
```
TheRealHaoLiu, line 282: this will cause an accumulation of AWX EE images that will be present in AWX

TheRealHaoLiu, line 284: maybe change to
```suggestion
  - name: "AWX EE (latest)"
    image: "quay.io/ansible/awx-ee:{{ _image_version }}"
```
to preserve old behavior?

TheRealHaoLiu, line 284: or
```suggestion
  - name: "AWX EE (latest)"
    image: "quay.io/ansible/awx-ee:{{ _image_version }}"
  - name: "AWX EE ({{ _image_version }})"
    image: "quay.io/ansible/awx-ee:{{ _image_version }}"
```

TheRealHaoLiu:
this issue to the matrix for the community to vote on the behavior please thumbs up the suggested change for the desire behavior that you want

roles/installer/defaults/main.yml
rooftopcellist: It's a tough decision.

On one hand, users who understand the implications and how they should be used would like both EE's to be  added (versioned and latest).  

However, sometimes it is better to be opinionated, in which case, we'd want to go with the versioned EE because that is what was tested with that version of AWX...
I lean towards loading both EE's. But I like the idea of polling the community on this PR and in Matrix to get feedback on which is desired.

kurokobo: The first item on this list will be adopted as the global default EE for the job templates that has no default EE. Upgrading will not remove existing old EEs, but the global default EE should always be replaced by the first one on this list upon upgrade.
Therefore, if the top of this list is the versioned image, existing jobs may start running on the new EE image after the upgrade.

In the future, I would like to remove the `latest` tag from this list.
The outdated `latest` images or the difference between the `latest` images cached on each nodes in the cluster is asking for trouble. If a job starts using a new EE and starts to fail, it is easier to troubleshoot and do a workaround if the versioned image is used rather than `latest`. I want to remove the question such as "How to renew `latest` iamge?" from the FAQ on the community.

However, since this is the first release where the EE image version is bound, at this stage, to reduce users confusion, I vote for keeping both `latest` and versioned.
 Also I would like to remove `latest` in 24.x or 25.x.

bartowl: with this suggestion it would not create new EE entry, but update the "latest" one. I guess this is exactly the behaviour that I wanted to prevent - that suddenly EE image changes without notice. By using verisoned ones one could set the default on ogranisation level and thus avoid unexpected changes when doing updrade. And after the upgrade is done, due to new EE slot with current version it would make it possible to slowly test all job templates against the new EE and eventually update Organisation to switch everything to the highest version.

Applying the suggestion the moment after upgrade everything would run with different EE without even simple possibility to downgrade. One difference from using literaly "latest" tag is - this switch would happen only when upgrading, reviously it happened after each new release showed up.

bartowl: If, then this seems better that previous suggestion because it leaves the auto-upgradeable "latest" ee yet still creating new ones with frozen version tags. When set on organisation or project level this would achieve my goal by preserving EE version through upgrade and release of new version. If someone wants to be on cutting edge, can use "latest" tag, otherwise keep to frozen tags. This would also create new EE entry with each new AWX version. I would choose that over the previous suggestion.

Please also keep in mind, that all previous versions already out there (before this PR eventually gets merged) will still have "latest" tag EE version that would not be removed nor changed. Therefore this solution even better fits the case

roles/installer/defaults/main.yml
```diff
@@ -278,10 +278,12 @@ init_container_extra_commands: ''
 init_container_extra_volume_mounts: ''
 
 ee_images:
-  - name: AWX EE (latest)
-    image: quay.io/ansible/awx-ee:latest
+  - name: "AWX EE (default)"
+    image: "quay.io/ansible/awx-ee:{{ _image_version }}"
```
kurokobo: @TheRealHaoLiu 
Thanks for updating, perhaps this line is incorrect.

```suggestion
    image: quay.io/ansible/awx-ee:latest
```

TheRealHaoLiu: i still want to have versioned image there instead of floating tag "latest"

changing `AWX EE (latest)` to `AWX EE (default)` to denote that its no longer `latest`

do you think we should still have an `AWX EE (latest)`?

TheRealHaoLiu: ok ur right, lets put back the latest for backward compatibility 

the main thing I'm trying to do at least is for the controlplane-ee to STOP using latest

rooftopcellist (approved):

PR #1747, opened 2024-03-06
Update PostgreSQL docs about finding default version

rooftopcellist:
##### SUMMARY

Some folks in Matrix noted that there are a couple docs that didn't get updated when we moved to using PostgreSQL 15 by default.  This PR cleans that up. 

##### ISSUE TYPE

### DIFF
--- a/docs/user-guide/database-configuration.md
+++ b/docs/user-guide/database-configuration.md
@@ -1,10 +1,10 @@
 ### Database Configuration
 
-#### Postgres Version
+#### PostgreSQL Version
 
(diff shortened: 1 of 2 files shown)


docs/user-guide/database-configuration.md
```diff
@@ -2,9 +2,9 @@
 
 #### Postgres Version
```
tvo318: We need to spell this out: PostgreSQL

tvo318 (commented):
Everything looks good, my only comment is to replace postgres with its actual spelled out name.

tvo318 (approved):
Nice work, thank you!

PR #1754, opened 2024-03-07
adding support for ansible metrics-utility

aknochow:
New or Enhanced Feature

adding support ansible metrics-utility


roles/installer/templates/cronjobs/metrics-report.yaml.j2
```diff
@@ -0,0 +1,56 @@
...
+              items:
+                - key: credentials.py
+                  path: 'credentials.py'
+          - name: {{ ansible_operator_meta.name }}-settings
+            configMap:
+              name: '{{ ansible_operator_meta.name }}-{{ deployment_type }}-configmap'
+              items:
+                - key: settings
+                  path: settings.py
+          restartPolicy: OnFailure
```
rooftopcellist: @aknochow if we are restarting on failure, we probably want to set a number of times to retry.

```
spec:
  jobTemplate:
    spec:
        backoffLimit: 2
```

roles/installer/templates/storage/metrics.yaml.j2
```diff
@@ -0,0 +1,15 @@
...
+  namespace: {{ ansible_operator_meta.namespace }}
+  ownerReferences: null
+  labels:
+    {{ lookup("template", "../common/templates/labels/common.yaml.j2") | indent(width=4) | trim }}
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: {{ metrics_storage_size | default('5Gi') }}
```
rooftopcellist: PVC's are often hard to resize unless you have a storage class that supports dynamic resizing.  Maybe we should set this a bit higher?  Depends on how much data we expect here.

dsavineau: `metrics_storage_size` isn't in the spec so how one could change this value ? Should we align this with the `xxx_storage_requirements` variable ?
Also should we add the ability to set the storage class here ?

roles/installer/templates/cronjobs/metrics-gather.yaml.j2
```diff
@@ -0,0 +1,56 @@
...
+  labels:
+    app.kubernetes.io/name: '{{ ansible_operator_meta.name }}-metrics-gather'
+    {{ lookup("template", "../common/templates/labels/common.yaml.j2") | indent(width=4) | trim }}
+spec:
+  schedule: "{{ _metrics_cronjob_gather_schedule }}"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
```
dsavineau: I think we should have some resource requirements for the metrics jobs too (`xxx_resource_requirements`)

roles/installer/templates/cronjobs/metrics-gather.yaml.j2
```diff
@@ -0,0 +1,56 @@
+---
+apiVersion: v1
```
dsavineau: I never know if we should be explicit on the api version value since cronjobs are using batch

```suggestion
apiVersion: batch/v1
```

roles/installer/templates/cronjobs/metrics-gather.yaml.j2
```diff
@@ -0,0 +1,56 @@
...
+  namespace: '{{ ansible_operator_meta.namespace }}'
+  labels:
+    app.kubernetes.io/name: '{{ ansible_operator_meta.name }}-metrics-gather'
+    {{ lookup("template", "../common/templates/labels/common.yaml.j2") | indent(width=4) | trim }}
+spec:
+  schedule: "{{ _metrics_cronjob_gather_schedule }}"
+  jobTemplate:
+    spec:
+      template:
+        spec:
```
dsavineau: We probably want to support `imagePullSecrets` here too

roles/installer/tasks/enable_metrics.yml
```diff
@@ -0,0 +1,32 @@
...
+- name: Check provided PVC claim exists
+  kubernetes.core.k8s_info:
+    name: "{{ _metrics_pvc_claim }}"
+    kind: PersistentVolumeClaim
+    namespace: "{{ ansible_operator_meta.namespace }}"
+  register: provided_pvc
+  when:
+    - _metrics_pvc_claim | length
+
+- block:
```
dsavineau: I don't think we need block statements here.
Also we could probably group the three k8s resources creation in a single k8s apply task + loop

https://github.com/ansible/awx-operator/blob/devel/roles/installer/tasks/resources_configuration.yml#L203-L216

roles/installer/tasks/install.yml
```diff
@@ -102,6 +102,10 @@
   include_tasks: initialize_django.yml
   when: awx_web_pod_name != ''
 
+- name: Enable optional metrics
+  include_tasks: enable_metrics.yml
+  when: metrics_enabled
```
dsavineau:
```suggestion
  when: metrics_enabled | bool
```

roles/installer/defaults/main.yml
```diff
@@ -481,3 +481,16 @@ nginx_worker_processes: 1
...
 nginx_worker_cpu_affinity: 'auto'
 nginx_listen_queue_size: "{{ uwsgi_listen_queue_size }}"
+
+_postgres_image: quay.io/sclorg/postgresql-15-c9s
+_postgres_image_version: latest
+
+# metrics-utility (github.com/ansible/metrics-utility)
+_metrics_image: "{{ metrics_image | default(_image) }}"
+_metrics_image_version: "{{ metrics_image_version | default(_image_version) }}"
+_metrics_enabled: "{{ metrics_enabled | default('false') }}"
```
```suggestion
_metrics_enabled: "{{ metrics_enabled | default(false) }}"
```

roles/installer/templates/cronjobs/metrics-gather.yaml.j2
```diff
@@ -0,0 +1,56 @@
...
+metadata:
+  name: {{ ansible_operator_meta.name }}-metrics-gather
+  namespace: '{{ ansible_operator_meta.namespace }}'
+  labels:
+    app.kubernetes.io/name: '{{ ansible_operator_meta.name }}-metrics-gather'
+    {{ lookup("template", "../common/templates/labels/common.yaml.j2") | indent(width=4) | trim }}
+spec:
+  schedule: "{{ _metrics_cronjob_gather_schedule }}"
+  jobTemplate:
+    spec:
```
dsavineau: Should we also add some metadata here (labels/annotations) like task/web deployments ?

https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/deployments/task.yaml.j2#L22-L35

aknochow:
I made a bunch of updates based on your comments, still need to take care of a few more.

roles/installer/templates/storage/metrics-utility.yaml.j2
```diff
@@ -0,0 +1,18 @@
...
+    {{ lookup("template", "../common/templates/labels/common.yaml.j2") | indent(width=4) | trim }}
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: {{ _metrics_utility_pvc_claim_size }}
+  {% if metrics_utility_pvc_storage_class is defined %}
+  storageClassName: {{ metrics_utility_pvc_storage_class }}
+  {% endif %}
```
dsavineau: Minor indent thing but for `metrics_utility_pvc_storage_class` then we don't have a parameter in the CRD to expose this right ? so user aren't able to override it
```suggestion
{% if metrics_utility_pvc_storage_class is defined %}
  storageClassName: {{ metrics_utility_pvc_storage_class }}
{% endif %}
```

roles/installer/defaults/main.yml
```diff
@@ -484,3 +484,14 @@ nginx_worker_processes: 1
 nginx_worker_connections: "{{ uwsgi_listen_queue_size }}"
 nginx_worker_cpu_affinity: 'auto'
 nginx_listen_queue_size: "{{ uwsgi_listen_queue_size }}"
+
+# metrics-utility (github.com/ansible/metrics-utility)
+_metrics_utility_enabled: "{{ metrics_utility_enabled | default('false') }}"
```
dsavineau: boolean value doesn't need to be quoted
```suggestion
_metrics_utility_enabled: "{{ metrics_utility_enabled | default(false) }}"
```

roles/installer/tasks/enable_metrics_utility.yml
```diff
@@ -0,0 +1,24 @@
+---
+# Check to make sure provided pvc exists, error loudly if not.  Otherwise, the management pod will just stay in pending state forever.
+- name: Check provided PVC claim exists
+  kubernetes.core.k8s_info:
+    name: "{{ _metrics_utility_pvc_claim }}"
+    kind: PersistentVolumeClaim
+    namespace: "{{ ansible_operator_meta.namespace }}"
+  register: provided_pvc
```
dsavineau: this `provided_pvc` registered value isn't used elsewhere. Could we remove it ?

roles/installer/templates/storage/metrics-utility.yaml.j2
aknochow: It was there but I named it `metrics_utility_pvc_claim_storage_class`.  Fixed.

dsavineau (approved):

PR #1779, opened 2024-03-18
docs: add description of init container image definition

zaki-lknr:
##### SUMMARY

Add `init_container_image` parameter.
This parameter is not dependent of `ee_images` settings. Therefore, use it when we want to pull a custom image or from private cotainer registry (e.g. disconnected internet).

##### ISSUE TYPE

### DIFF
--- a/docs/user-guide/advanced-configuration/deploying-a-specific-version-of-awx.md
+++ b/docs/user-guide/advanced-configuration/deploying-a-specific-version-of-awx.md
@@ -28,7 +28,10 @@ spec:
   ee_images:
     - name: my-custom-awx-ee
       image: myorg/my-custom-awx-ee
+  control_plane_ee_image: myorg/my-custom-awx-ee:latest
   init_container_image: myorg/my-custom-awx-ee
+  init_container_image_version: latest
+  init_projects_container_image: myorg/my-mirrored-centos:stream9
 ```
 
 **Note**: The `image` and `image_version` are intended for local mirroring scenarios. Please note that using a version of AWX other than the one bundled with the `awx-operator` is **not** supported. For the default values, check the [main.yml](https://github.com/ansible/awx-operator/blob/devel/roles/installer/defaults/main.yml) file.
--- a/docs/user-guide/advanced-configuration/deploying-a-specific-version-of-awx.md
+++ b/docs/user-guide/advanced-configuration/deploying-a-specific-version-of-awx.md
@@ -2,17 +2,19 @@
 
 There are a few variables that are customizable for awx the image management.
 
-| Name                         | Description               | Default                                    |
-| ---------------------------- | ------------------------- | ------------------------------------------ |
-| image                        | Path of the image to pull | quay.io/ansible/awx                        |
-| image_version                | Image version to pull     | value of DEFAULT_AWX_VERSION or latest     |
-| image_pull_policy            | The pull policy to adopt  | IfNotPresent                               |
-| image_pull_secrets           | The pull secrets to use   | None                                       |
-| ee_images                    | A list of EEs to register | quay.io/ansible/awx-ee:DEFAULT_AWX_VERSION |
-| redis_image                  | Path of the image to pull | docker.io/redis                            |
-| redis_image_version          | Image version to pull     | latest                                     |
-| init_container_image         | Path of the image to pull | quay.io/ansible/awx-ee                     |
-| init_container_image_version | Image version to pull     | value of DEFAULT_AWX_VERSION or latest     |
+| Name                          | Description               | Default                                    |
+| ----------------------------- | ------------------------- | ------------------------------------------ |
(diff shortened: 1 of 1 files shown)


docs/user-guide/advanced-configuration/deploying-a-specific-version-of-awx.md
```diff
@@ -26,6 +28,7 @@ spec:
   ee_images:
     - name: my-custom-awx-ee
       image: myorg/my-custom-awx-ee
+  init_container_image: myorg/my-custom-awx-ee
```
kurokobo, line 34: I think both `init_container_image` and `init_container_image_version` should be in example, since specified value is ignored if only one or the other is specified. In this case, following `when` never be `true`: https://github.com/ansible/awx-operator/blob/49d7a566b22524d7d56b965689c81ffba8fbfb1a/roles/installer/tasks/set_images.yml#L7-L12

Additionally, in the case that `init_container_image` is required, `control_plane_ee_image` may also be required. So if you have time, could you please add `control_plane_ee_image` in the table above and in this example? (Of course I can by creating my PR, but I'd like to respect your accomplishment)

Actually there is another parameter `init_projects_container_image` which defaulted to `quay.io/centos/centos:stream9`, so should this also be documented in the table?

```suggestion
  control_plane_ee_image: myorg/my-custom-awx-ee:latest
  init_container_image: myorg/my-custom-awx-ee
  init_container_image_version: latest
  init_projects_container_image: myorg/my-mirrored-centos:stream9
```

docs/user-guide/advanced-configuration/deploying-a-specific-version-of-awx.md
```diff
@@ -2,15 +2,17 @@
...
+| ---------------------------- | ------------------------- | ------------------------------------------ |
+| image                        | Path of the image to pull | quay.io/ansible/awx                        |
+| image_version                | Image version to pull     | value of DEFAULT_AWX_VERSION or latest     |
+| image_pull_policy            | The pull policy to adopt  | IfNotPresent                               |
+| image_pull_secrets           | The pull secrets to use   | None                                       |
+| ee_images                    | A list of EEs to register | quay.io/ansible/awx-ee:DEFAULT_AWX_VERSION |
+| redis_image                  | Path of the image to pull | docker.io/redis                            |
+| redis_image_version          | Image version to pull     | latest                                     |
+| init_container_image         | Path of the image to pull | quay.io/ansible/awx-ee                     |
+| init_container_image_version | Image version to pull     | value of DEFAULT_AWX_VERSION or latest     |
```
kurokobo: We should consider adding `control_plane_ee_image` and `init_projects_container_image` to this table. Could you please add these parameters, or may I create another PR?

```suggestion
| Name                          | Description               | Default                                    |
| ----------------------------- | ------------------------- | ------------------------------------------ |
| image                         | Path of the image to pull | quay.io/ansible/awx                        |
| image_version                 | Image version to pull     | value of DEFAULT_AWX_VERSION or latest     |
| image_pull_policy             | The pull policy to adopt  | IfNotPresent                               |
| image_pull_secrets            | The pull secrets to use   | None                                       |
| ee_images                     | A list of EEs to register | quay.io/ansible/awx-ee:DEFAULT_AWX_VERSION |
| redis_image                   | Path of the image to pull | docker.io/redis                            |
| redis_image_version           | Image version to pull     | latest                                     |
| control_plane_ee_image        | Image version to pull     | quay.io/ansible/awx-ee:DEFAULT_AWX_VERSION |
| init_container_image          | Path of the image to pull | quay.io/ansible/awx-ee                     |
| init_container_image_version  | Image version to pull     | value of DEFAULT_AWX_VERSION or latest     |
| init_projects_container_image | Image version to pull     | quay.io/centos/centos:stream9              |
```

kurokobo (commented):
@zaki-lknr 
Hi, thanks for sending helpful PR to improve docs! 😃 
I've leave some minor comments, could you please check if you would like?

docs/user-guide/advanced-configuration/deploying-a-specific-version-of-awx.md
zaki-lknr: @kurokobo 
I have confirmed that it will not work unless the init_container_image_version is also specified. Thanks your comments.

kurokobo (approved):
Thanks for updating!
LGTM, but I don't have any permission to do anything, so I can only comment on this😞

rooftopcellist (approved):
@zaki-lknr Thanks for adding this improvement to the docs!

I think it might be useful to (in a follow-up PR) add a sample AWX custom resource yaml file for disconnected environment, like what we did [here](https://github.com/ansible/awx-operator/blob/devel/config/samples/awx_v1beta1_awx_resource_limits.yaml) for resource requests and limits configuration, etc.  Then link to it from these docs so that folks can copy, paste and fill out their specific internal registry urls.

PR #1783, opened 2024-03-20
csv: Fix metrics utility fields

dsavineau:
##### SUMMARY

The metrics utility fields were configured under the statusDescriptors section rather than specDescriptors so displaying those fields in the UI wasn't done correctly (not under the Advanced section nor using the correct field type).

This also changes the `metrics_utility_configmap` descriptor from `urn:alm:descriptor:com.tectonic.ui:selector:core:v1:ConfigMap` to `urn:alm:descriptor:io.kubernetes:ConfigMap` because the first value doesn't work.

Finally, all metrics utility fields are only displayed (in the Advanced section) when `metrics_utility_enabled` is enabled (not default).

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

metrics_utility_disabled
metrics_utility_enabled


aknochow (approved):

rooftopcellist (approved):

rooftopcellist:
Thanks for the PR!

PR #1795, opened 2024-03-27
Update doc exaple for postgres_data_path

rooftopcellist:
##### SUMMARY

The new postgres_data_path is `/var/lib/postgresql/data/pgdata` for the sclorg image.  The docs hadn't been updated.

##### ISSUE TYPE

### DIFF
--- a/docs/user-guide/database-configuration.md
+++ b/docs/user-guide/database-configuration.md
@@ -63,7 +63,7 @@ The following variables are customizable for the managed PostgreSQL service
 | postgres_resource_requirements                | PostgreSQL container resource requirements    | requests: {cpu: 10m, memory: 64Mi}      |
 | postgres_storage_requirements                 | PostgreSQL container storage requirements     | requests: {storage: 8Gi}                |
 | postgres_storage_class                        | PostgreSQL PV storage class                   | Empty string                            |
-| postgres_data_path                            | PostgreSQL data path                          | `/var/lib/postgresql/data/pgdata`       |
+| postgres_data_path                            | PostgreSQL data path                          | `/var/lib/pgsql/data/pgdata`            |
 | postgres_priority_class                       | Priority class used for PostgreSQL pod        | Empty string                            |
 
 Example of customization could be:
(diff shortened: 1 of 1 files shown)


dhageman (approved):

docs/user-guide/database-configuration.md
```diff
@@ -63,7 +63,7 @@ The following variables are customizable for the managed PostgreSQL service
 | postgres_resource_requirements                | PostgreSQL container resource requirements    | requests: {cpu: 10m, memory: 64Mi}      |
 | postgres_storage_requirements                 | PostgreSQL container storage requirements     | requests: {storage: 8Gi}                |
 | postgres_storage_class                        | PostgreSQL PV storage class                   | Empty string                            |
-| postgres_data_path                            | PostgreSQL data path                          | `/var/lib/postgresql/data/pgdata`       |
+| postgres_data_path                            | PostgreSQL data path                          | `/var/lib/pgsql/data/pgdata`            |
```
kurokobo: Correct path is `/var/lib/pgsql/data/userdata` instead of `/var/lib/pgsql/data/pgdata`.

See the `common.sh` in sclorg repo: https://github.com/sclorg/postgresql-container/blob/3836b381f8f91c1897b9cb12bccfa74e4d283fdd/15/root/usr/share/container-scripts/postgresql/common.sh#L268

Also EDA Server Operator had already configured to have `/var/lib/pgsql/data/userdata`: https://github.com/ansible/eda-server-operator/blob/d1f9b7fb41e207f6732c9a70ecd120e04cf80fdc/roles/postgres/defaults/main.yml#L42

```suggestion
| postgres_data_path                            | PostgreSQL data path                          | `/var/lib/pgsql/data/userdata`          |
```

kurokobo (commented):
@rooftopcellist 
Thanks for working on this, in addition to the docs, we have to correct path for:

- Default value for this variable: https://github.com/ansible/awx-operator/blob/3c7059870453b0603266546c0f96423c1280570c/roles/installer/defaults/main.yml#L404
- Search path for `PG_VERSION` file (add `elif` for `/var/lib/pgsql/data/userdata/PG_VERSION`?): https://github.com/ansible/awx-operator/blob/3c7059870453b0603266546c0f96423c1280570c/roles/installer/tasks/database_configuration.yml#L163-L170

docs/user-guide/database-configuration.md
kurokobo: The users should not change this value when using sclorg's PSQL, so I think it is a good idea not to put it in the documentation in the first place, or at the very least, add a note that not to change it when using a sclorg image.
Changing this value to any other path would put the DB data files outside of PV.

rooftopcellist:
Fixed by https://github.com/ansible/awx-operator/pull/1798

PR #1798, opened 2024-03-28
Remove the ability to customize the postgres_data_dir

rooftopcellist:
##### SUMMARY

Remove the ability to customize the postgres_data_dir
* in the sclorg Postgresql 15 image, the PGDATA directory is hardcoded
* if users were to modify this directory, they would only change the directory the pvc is mounted to, not the directory PostgreSQL uses. This would result in loss of data.

##### ISSUE TYPE

cc @kurokobo

### DIFF
--- a/.helm/starter/templates/storage/postgres-pv.yaml
+++ b/.helm/starter/templates/storage/postgres-pv.yaml
@@ -13,7 +13,7 @@ spec:
     storage: {{ default "8Gi" .size | quote }}
   storageClassName: {{ include "postgres.storageClassName" $ }}
   hostPath:
-    path: {{ required "customVolumes.postgres.hostPath or spec.postgres_data_path are required!" (default ($.Values.AWX.spec).postgres_data_path .hostPath) | quote }}
+    path: /var/lib/pgsql/data/userdata
 {{- end }}
 {{- end }}
 {{- end }}
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -1807,9 +1807,6 @@ spec:
               postgres_priority_class:
                 description: Assign a preexisting priority class to the postgres pod
                 type: string
-              postgres_data_path:
-                description: Path where the PostgreSQL data are located
-                type: string
               postgres_extra_args:
                 type: array
                 items:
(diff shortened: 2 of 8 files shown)


fosterseth (approved):

dsavineau (approved):

kurokobo:
I haven't verified this yet, but when we upgrade the CRDs to this PR's while there is already an AWX CR with `postgres_data_path`, what happens to the AWX CR?

rooftopcellist:
@kurokobo The k8s validator still applies the AWX CR, but omits the `postgres_data_path` field.  As an extra measure, I renamed the internal variables to `_postgres_data_path` so that there is no way to override the path.  

The caveat to this approach is that users won't be able to use the old `postgres` image from dockerhub for managed deployments because the PVC will no longer be mounted.  But I think that is OK, being a _managed_ deployment, I think it is OK be opinionated here to simplify things.  Users can always create a postgres statefulset and pvc themselves and configure it as an external db if they want to use the `postgres` image from dockerhub.  

I left in the `postgres_image` and `postgres_image_version` parameters though as they still serve the purpose of making it possible to specify a mirrored image.  Users who mirror for air-gapped/disconnected environments will need to update their mirror to the new postgres image.  

I added a warning about this in the release notes for 2.13.1 and 2.14.0.

kurokobo:
@rooftopcellist 
Thanks for clarifying details for k8s validator. I also tested this PR by;

- Deploy minimal AWX with `postgres_data_path` by Operator **2.12.2**
  ```yaml
  ---
  kind: AWX
  metadata:
    namespace: awx
    name: awx-emo
  spec:
    service_type: nodeport
    nodeport_port: 30080
    postgres_data_path: /completely/customized/psql/data/path
  ```
- Create new users and launch jobs multiple times to make changes on DB
- Upgrade Operator to this PR:
  ```bash
  gh pr checkout https://github.com/ansible/awx-operator/pull/1798
  IMG=registry.example.com/ansible/awx-operator:pr1798 BUILD_ARGS="--build-arg DEFAULT_AWX_VERSION=24.1.0" make docker-build docker-push deploy
  ```

and I can confirm this PR works as expected 👍 

- AWX CR still there, just `postgres_data_path` in AWX CR is removed by CRD changes
- New reconcile is completed without any errors
- PSQL is upgraded from 13 to 15
- PSQL mounts new PVC on `/var/lib/pgsql/data`
- AWX is live
- All data in PSQL are migrated; users and jobs that created before upgrading are persisted after upgrading
- Restarting PSQL 15 does not loss any data; I can still see users and jobs after restarting

rooftopcellist:
@kurokobo Thank you for testing this out so thoroughly!

Issue #1775, opened 2024-03-15
Upgrade from 2.12.2 to 2.13.1 failed - mkdir: cannot create directory '/var/lib/pgsql/data/userdata': Permission denied


craph:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that the AWX Operator is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Unable to upgrade from 2.12.2 to 2.13.1
```
mkdir: cannot create directory '/var/lib/pgsql/data/userdata': Permission denied
```

CrashLoopBackOff in the Postgres container.

### AWX Operator version

2.13.1

### AWX version

24.0.0

### Kubernetes platform

kubernetes

### Kubernetes/Platform version

v1.25.16+rke2r1

### Modifications

no

### Steps to reproduce

upgrade from 2.12.2 to 2.13.1

### Expected results

migration from postgres 13 to 15 should work without any permissions issues

### Actual results

Unable to upgrade from 2.12.2 to 2.13.1
```
mkdir: cannot create directory '/var/lib/pgsql/data/userdata': Permission denied
```

CrashLoopBackOff in the Postgres container.

craph:
Looks similar to https://github.com/ansible/awx-operator/issues/1770

craph:
In the documentation from [here](https://quay.io/repository/sclorg/postgresql-15-c9s) there is this paragraph : 

> Notice: When mouting a directory from the host into the container, ensure that the mounted directory has the appropriate permissions and that the owner and group of the directory matches the user UID or name which is running inside the container.
> 
> Typically processes in container run under UID 26, so -- on GNU/Linux 

is there any reason to change container image from official postgres to sclorg one ? 

Thank you very much for your help.

craph:
In the Dockerfile [here](https://github.com/sclorg/postgresql-container/blob/master/15/Dockerfile.c9s) we can see :
```
# This image must forever use UID 26 for postgres user so our volumes are
# safe in the future. This should *never* change, the last test is there
# to make sure of that.
RUN  { yum -y module enable postgresql:15 || :; } && \
    INSTALL_PKGS="rsync tar gettext bind-utils nss_wrapper postgresql-server postgresql-contrib" && \
[...]

Issue #1770, opened 2024-03-13
Postgres 15 pod: cannot create directory '/var/lib/pgsql/data/userdata': Permission denied"


Tfinn92:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [X] I understand that the AWX Operator is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Updating to 2.13.1 through helm results in the postgres15 pod having the following error:
`cannot create directory '/var/lib/pgsql/data/userdata': Permission denied"`

### AWX Operator version

2.13.1

### AWX version

24.0.0

### Kubernetes platform

kubernetes

### Kubernetes/Platform version

Rancher RKE2 v1.26.8+rke2r1 and another on v1.27.10+rke2r1

### Modifications

no

### Steps to reproduce

Have cluster with 2.12.2 installed and run `helm upgrade awx-operator awx-operator/awx-operator`

### Expected results

pods come up no problem

### Actual results

postgres15 pod CrashLoopBackOff 
Logs show `"mkdir: cannot create directory '/var/lib/pgsql/data/userdata': Permission denied"`

kurokobo:
@Tfinn92 
New PSQL is running as UID:26, so PV has to be writable by this UID.

kurokobo:
@TheRealHaoLiu @fosterseth 
The same consideration can be applied for backup. UID:26 has to have write perm on PV to create backup directory.

ivantomica:
Same issue here, but when I scaled down deployment of the `awx-postgres-15-0 `statefulset to `0`, and mounted PVC to the other container (temp busybox one), I found PV contained only `data/` directory and it was empty:

```
/mnt/pv-data # ls -alh
total 28K    
drwxr-xr-x    4 root     root        4.0K Mar 14 07:14 .
drwxr-xr-x    3 root     root        4.0K Mar 14 07:14 ..
drwxr-xr-x    2 root     root        4.0K Mar 12 20:20 data
drwx------    2 root     root       16.0K Mar 12 20:20 lost+found
/mnt/pv-data # ls -alh data/
total 8K     
drwxr-xr-x    2 root     root        4.0K Mar 12 20:20 .
drwxr-xr-x    4 root     root        4.0K Mar 14 07:14 ..
```

[...]

PR #1799, opened 2024-03-28
Add postgres init container

fosterseth:
##### SUMMARY

Potentially resolves:
* https://github.com/ansible/awx-operator/issues/1770
* https://github.com/ansible/awx-operator/issues/1775

Create a postgres init container that will run user-provided `init_postgres_extra_commands`

for example, in the awx spec file:

```yaml
  init_postgres_extra_commands: |
    sudo touch /var/lib/pgsql/data/foo
    sudo touch /var/lib/pgsql/data/bar
    chown 26:26 /var/lib/pgsql/data/foo
    chown root:root /var/lib/pgsql/data/bar
```

##### ISSUE TYPE

 - New or Enhanced Feature

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -1811,6 +1811,9 @@ spec:
                 type: array
                 items:
                   type: string
+              postgres_init_container_extra_commands:
+                description: Extra commands for the init postgres container
+                type: string
               postgres_extra_volumes:
                 description: Specify extra volumes to add to the application pod
                 type: string
--- a/roles/installer/templates/statefulsets/postgres.yaml.j2
+++ b/roles/installer/templates/statefulsets/postgres.yaml.j2
@@ -47,6 +47,26 @@ spec:
 {% if postgres_priority_class is defined %}
       priorityClassName: '{{ postgres_priority_class }}'
 {% endif %}
+{% if postgres_init_container_extra_commands is defined %}
+      initContainers:
+        - name: init
+          image: '{{ _postgres_image }}'
+          imagePullPolicy: '{{ image_pull_policy }}'
+          securityContext:
+            runAsUser: 0
+          command:
+            - /bin/sh
+            - -c
+            - |
+              {{ postgres_init_container_extra_commands | indent(width=14) }}
+          volumeMounts:
+            - name: postgres-{{ supported_pg_version }}
+              mountPath: '{{ _postgres_data_path | dirname }}'
+              subPath: '{{ _postgres_data_path | dirname | basename }}'
+{% if postgres_extra_volume_mounts %}
+            {{ postgres_extra_volume_mounts | indent(width=12, first=True) }}
+{% endif %}
+{% endif %}
       containers:
         - image: '{{ _postgres_image }}'
           imagePullPolicy: '{{ image_pull_policy }}'
@@ -113,7 +133,7 @@ spec:
             - name: postgres-{{ supported_pg_version }}
               mountPath: '{{ _postgres_data_path | dirname }}'
               subPath: '{{ _postgres_data_path | dirname | basename }}'
-{% if postgres_extra_volume_mounts -%}
+{% if postgres_extra_volume_mounts %}
             {{ postgres_extra_volume_mounts | indent(width=12, first=True) }}
 {% endif %}
           resources: {{ postgres_resource_requirements }}
--- a/docs/user-guide/database-configuration.md
+++ b/docs/user-guide/database-configuration.md
@@ -99,3 +99,15 @@ We recommend you use the default image sclorg image. If you are coming from a de
 You can no longer configure a custom `postgres_data_path` because it is hardcoded in the quay.io/sclorg/postgresql-15-c9s image.
 
 If you override the postgres image to use a custom postgres image like postgres:15 for example, the default data directory path may be different. These images cannot be used interchangeably.
+
+#### Postgres init extra commands
+
+Users can define arbitrary commands to run from a postgres init container.
+
+For example this may be useful for setting permissions and ownership of the postgres data directory.
+
+```yaml
+postgres_init_container_extra_commands: |
+  chown 26:0 /var/lib/pgsql/data
+  chmod 700 /var/lib/pgsql/data
+```


roles/installer/defaults/main.yml
```diff
@@ -401,7 +401,7 @@ postgres_init_container_resource_requirements:
     memory: 64Mi
 # Assign a preexisting priority class to the postgres pod
 postgres_priority_class: ''
-postgres_data_path: '/var/lib/postgresql/data/pgdata'
+postgres_data_path: '/var/lib/pgsql/data/pgdata'
```
kurokobo: This will conflicts with https://github.com/ansible/awx-operator/pull/1798, and the path is incorrect.

```suggestion
postgres_data_path: '/var/lib/pgsql/data/userdata'
```

kurokobo (commented):
Hi, thanks for working on this.

If we don't implement any default value for `init_postgres_extra_commands`, I think there should be a complete example that can use as it is instead of `foo` nor `bar`, in the docs, something like;

```yaml
  init_postgres_extra_commands: |
    chown 26:0 /var/lib/pgsql/data
    chmod 700 /var/lib/pgsql/data 
```

fosterseth:
@kurokobo thanks, added a snippet to the docs

config/crd/bases/awx.ansible.com_awxs.yaml
```diff
@@ -1748,6 +1748,9 @@ spec:
               init_container_extra_commands:
                 description: Extra commands for the init container
                 type: string
+              init_postgres_extra_commands:
```
hhk7734: How about changing it to `postgres_init_container_extra_commands`? In [docs/user-guide/database-configuration.md](https://github.com/ansible/awx-operator/blob/devel/docs/user-guide/database-configuration.md), all option start with `postgres_`.

roles/installer/templates/statefulsets/postgres.yaml.j2
```diff
@@ -46,6 +46,29 @@ spec:
...
+          volumeMounts:
+            - name: postgres-{{ supported_pg_version }}
+              mountPath: '{{ postgres_data_path | dirname }}'
+              subPath: '{{ postgres_data_path | dirname | basename }}'
+{% if postgres_extra_volume_mounts -%}
+            {{ postgres_extra_volume_mounts | indent(width=12, first=True) }}
+{% endif %}
+{% endif %}
+{% if init_container_extra_volume_mounts -%}
+            {{ init_container_extra_volume_mounts | indent(width=12, first=True) }}
```
hhk7734: This is not in volumes. I think this needs to be removed.

roles/installer/templates/statefulsets/postgres.yaml.j2
```diff
@@ -46,6 +46,29 @@ spec:
 {% endif %}
 {% if postgres_priority_class is defined %}
       priorityClassName: '{{ postgres_priority_class }}'
+{% endif %}
+{% if init_postgres_extra_commands %}
```
kurokobo: This is causing error if `init_postgres_extra_commands` is not used in AWX CR:

```bash
fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'init_postgres_extra_commands' is undefined. 'init_postgres_extra_commands' is undefined. 'init_postgres_extra_commands' is undefined. 'init_postgres_extra_commands' is undefined\n\nThe error appears to be in '/opt/ansible/roles/installer/tasks/database_configuration.yml': line 186, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n- block:\n    - name: Create Database if no database is specified\n      ^ here\n"}
```

We should use `is defined` for this conditional.

```suggestion
{% if init_postgres_extra_commands is defined %}
```

roles/installer/templates/statefulsets/postgres.yaml.j2
```diff
@@ -46,6 +46,29 @@ spec:
...
+            runAsUser: 0
+          command:
+            - /bin/sh
+            - -c
+            - |
+              {{ init_postgres_extra_commands | indent(width=14) }}
+          volumeMounts:
+            - name: postgres-{{ supported_pg_version }}
+              mountPath: '{{ postgres_data_path | dirname }}'
+              subPath: '{{ postgres_data_path | dirname | basename }}'
```
kurokobo: `postgres_data_path` is removed by https://github.com/ansible/awx-operator/pull/1798. Using this causing error:

```bash
fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: Unable to look up a name or access an attribute in template string (# Postgres StatefulSet.\n...
```

We should use `_postgres_data_path` instead.

```suggestion
              mountPath: '{{ _postgres_data_path | dirname }}'
              subPath: '{{ _postgres_data_path | dirname | basename }}'
```

kurokobo (commented):
Hi, I'm testing this PR and found some issues. Could you please check my comments?
Also there are some updates on `devel` branch so if you can please rebase your branch.

roles/installer/templates/statefulsets/postgres.yaml.j2
kurokobo: @hhk7734 
I think this is intended, since any other `*_extra_volume_mounts` are designed to use with single `extra_volumes`: https://ansible.readthedocs.io/projects/awx-operator/en/latest/user-guide/advanced-configuration/custom-volume-and-volume-mount-options.html

hhk7734: If `init_container_extra_volume_mounts` is designed to use with `extra_volumes`, it need to add `extra_volumes` to volumes in this template.

kurokobo: Oh, sorry, it looks like `postgres_extra_volumes` is used for the additional PSQL volumes instead of `extra_volumes`.
I don't see it mentioned in the documentation. This is a different issue: https://github.com/ansible/awx-operator/pull/1799/files#diff-57afb141659730e9bae27cd3ea91bfe5b1709a2569d98759221305423d544c03L128-L131

rooftopcellist: I don't think that `init_container_extra_volume_mounts` block belongs here.  The `postgres_extra_volume_mounts` parameter is sufficient.

config/crd/bases/awx.ansible.com_awxs.yaml
rooftopcellist: I totally agree.  I will push a change for this shortly, thanks!

roles/installer/templates/statefulsets/postgres.yaml.j2
rooftopcellist: Thanks!  I added this and your other code suggestions.

rooftopcellist:
I tested this out with @kurokobo 's awx-on-k3s README.md and yaml snippets using this changes and it works now _without_ having to run `sudo chown 26:0 /data/postgres-15/data` on the host directory that gets mounted.  

First I stood up a k3s cluster:
```
curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=v1.28.7+k3s1 sh -s - --write-kubeconfig-mode 644

# Set KUBECONFIG
export KUBECONFIG=/etc/rancher/k3s/k3s.yaml
```

Instead of running `kubectl apply -k operator`, I ran the following so I would have the changes from this PR:

```
# Build & Push Operator Image
make docker-build docker-push IMG=quay.io/username/awx-operator:dev

# Deploy Operator
make deploy NAMESPACE=$NAMESPACE IMG=quay.io/username/awx-operator:dev
```


```
kubectl -n awx get pods

# Create SSL cert and key
AWX_HOST="awx.example.com"
openssl req -x509 -nodes -days 3650 -newkey rsa:2048 -out ./base/tls.crt -keyout ./base/tls.key -subj "/CN=${AWX_HOST}/O=${AWX_HOST}" -addext "subjectAltName = DNS:${AWX_HOST}"
cat base/awx.yaml 

# Pre-create data directory for database at /data
sudo mkdir -p /data/postgres-15/data
sudo mkdir -p /data/projects
# sudo chown 26:0 /data/postgres-15/data
sudo chown 1000:0 /data/projects
sudo chmod 700 /data/postgres-15/data

# Apply AWX custom resource
kubectl apply -k base
```

I observed the the migrations k8s job completed without any permissions issues.  Similarly, there are no longer any permissions issues in the postgres pod.  

Everything deployed successfully and I can reach the UI at `awx.example.com` locally.

rooftopcellist:
@kurokobo @craph 
@TheRealHaoLiu and I wanted to find a way to run these init commands by default, but could not figure out a way to do so that did not involve adding a lot of complexity.  This is because in Openshift, we cannot easily run a container as root (uid=0) and should never do so by default.  Furthermore, there may be k8s cluster/storageclass combinations that this might cause issues for.

We thought it might be easier for users to configure this as a flag instead of a set of commands.  Let me know if you have feedback on the approach on https://github.com/ansible/awx-operator/pull/1805 versus the approach on this PR.

kurokobo:
@rooftopcellist 
Hi, thanks for working on this!
I vote for the flag PR #1805. It would provide a simple and better user experience.

craph:
Hi @rooftopcellist,
Thank you very much for your work on this subject.
I haven't tested it on my side because I revert my production instance to version 2.12.2

May be the PR #1805 is a best solution to provide a better user experience with a good updated documentation to explain the spec.

Best regards,

rooftopcellist:
@craph thanks for weighing in.  I added in docs here to help make things clean for future users:
* https://github.com/ansible/awx-operator/pull/1805/files#diff-e93c5e0f5a35e585b520af34a34e8a7ef808bf70cfe28d7e0fc9582142dadb41R102-R119

I will close this PR in favor of https://github.com/ansible/awx-operator/pull/1805

PR #1801, opened 2024-03-30
Add the ability to delete the old postgres pvc automatically if desired

rooftopcellist:
- this parameter will be hidden on the UI and should only be used for test runs etc. Users should be intention about deleting the old PVC, which is why by default, this will not delete the old PVC from the previous PostgreSQL version

##### SUMMARY
This adds back the functionality we had before, but with the correct boolean value...

##### ISSUE TYPE

### DIFF
--- a/roles/installer/tasks/upgrade_postgres.yml
+++ b/roles/installer/tasks/upgrade_postgres.yml
@@ -153,3 +153,15 @@
   loop:
     - "{{ ansible_operator_meta.name }}-postgres"
     - "{{ ansible_operator_meta.name }}-postgres-13"
+
+- name: Remove old persistent volume claim
+  k8s:
+    kind: PersistentVolumeClaim
+    api_version: v1
+    namespace: "{{ ansible_operator_meta.namespace }}"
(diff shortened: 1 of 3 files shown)


roles/installer/tasks/upgrade_postgres.yml
```diff
@@ -153,3 +153,15 @@
...
+
+- name: Remove old persistent volume claim
+  k8s:
+    kind: PersistentVolumeClaim
+    api_version: v1
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    name: "{{ item }}"
+    state: absent
+  loop:
+    - "postgres-{{ ansible_operator_meta.name }}-postgres-13-0"
```
lucas-benedito, line 165: Missing `| bool` filter — CR spec values can arrive as strings, and `not "false"` is truthy in Jinja2, so PVCs would never be deleted. Same pattern as `cleanup.yml`:
```yaml
when: not postgres_keep_pvc_after_upgrade | bool
```

lucas-benedito (changes requested):

PR #1804, opened 2024-04-02
add tolerations/nodeselector to migration job template (fixes #1774)

### LINKED ISSUE #1774: Missing Toleration for Migration pod.

- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.
- [X] I am **NOT** reporting a (potential) security vulnerability. (These should be emailed to `security@ansible.com` instead.)

### Bug Summary

All other pods has tolerations except migration job missing tolerations and nodeselector. 

### AWX version

24.0.0

### Select the relevant components

- [ ] UI
- [ ] UI (tech preview)
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [X] Other

### Installation method

kubernetes

### Modifications

no

### Ansible version

core 2.14.6

### Operating system

Ubuntu Server 22.04

### Web browser

_No response_

### Steps to reproduce

awx created with the following configuration. All other pods except migration can define the tolerations.

kind: AWX
metadata:
  name: awx
  namespace: awx
spec:
  ingress_class_name: nginx
  ingress_path: /awx
  ingress_type: ingress
  node_selector: |
    node-role.kubernetes.io/control-plane: ""
  postgres_selector: |
    node-role.kubernetes.io/control-plane: ""
  postgres_tolerations: |
    - effect: NoSchedule
      key: node-role.kubernetes.io/control-plane
      operator: Exists
  task_node_selector: |
    node-role.kubernetes.io/control-plane: ""
  task_tolerations: |
    - effect: NoSchedule
      key: node-role.kubernetes.io/control-plane
      operator: Exists
  tolerations: |
    - effect: NoSchedule
      key: node-role.kubernetes.io/control-plane
      operator: Exists
  web_node_selector: |
    node-role.kubernetes.io/control-plane: ""
  web_tolerations: |
    - effect: NoSchedule
      key: node-role.kubernetes.io/control-plane
      operator: Exists

### Expected results

Migration pod can run to completion status and task pod can be in running state.

### Actual results

MIgration pod stuck in pending status and task pod stuck in init status.

### Additional information

_No response_

ranvit:
##### SUMMARY
As described in https://github.com/ansible/awx-operator/issues/1774, the migration pod has no provisions to add tolerations to it. This PR's solution is to use the postgres pod's toleration/selector on the migration job. 

The docs would need to be updated to alert the user about this relationship between postgres and the migration pod

##### ISSUE TYPE

### DIFF
--- a/docs/user-guide/advanced-configuration/assigning-awx-pods-to-specific-nodes.md
+++ b/docs/user-guide/advanced-configuration/assigning-awx-pods-to-specific-nodes.md
@@ -88,3 +88,8 @@ spec:
               - S2
           topologyKey: topology.kubernetes.io/zone
 ```
+
+#### Special Note on DB-Migration Job Scheduling
+
+For the **db-migration job**, which applies database migrations at cluster startup, you can specify scheduling settings using the `task_*` configurations such as `task_node_selector`, `task_tolerations`, etc.  
+If these task-specific settings are not defined, the job will automatically use the global AWX configurations like `node_selector` and `tolerations`.
--- a/roles/installer/templates/jobs/migration.yaml.j2
+++ b/roles/installer/templates/jobs/migration.yaml.j2
@@ -43,6 +43,34 @@ spec:
         - name: {{ secret }}
 {% endfor %}
 {% endif %}
+{% if task_node_selector %}
+      nodeSelector:
+        {{ task_node_selector | indent(width=8) }}
+{% elif node_selector %}
+      nodeSelector:
+        {{ node_selector | indent(width=8) }}
+{% endif %}
+{% if task_topology_spread_constraints %}
+      topologySpreadConstraints:
+        {{ task_topology_spread_constraints | indent(width=8) }}
+{% elif topology_spread_constraints %}
+      topologySpreadConstraints:
+        {{ topology_spread_constraints | indent(width=8) }}
+{% endif %}
+{% if task_tolerations %}
+      tolerations:
+        {{ task_tolerations | indent(width=8) }}
+{% elif tolerations %}
+      tolerations:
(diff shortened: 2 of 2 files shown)


rooftopcellist:
@ranvit Is there a reason why `postgres_tolerations` and `postgres_node_selector` is used instead of `task_tolerations` and `task_node_selector`?

My thinking here is that specifying `postgres_` params might not make sense in some scenarios, for example external db scenarios.  

The k8s db migration job uses the task container, and previously this was running in the task container, so using `task_*` params for this might be a bit more intuitive.  

Also, could you add some docs here to make this a bit more discoverable for users?
* https://github.com/ansible/awx-operator/blob/devel/docs/user-guide/advanced-configuration/assigning-awx-pods-to-specific-nodes.md

ranvit:
@rooftopcellist thank you for the feedback. It makes sense for this migration job to utilize the task_* settings. 

I decided to implement this the same way the Task pods have been configured [here](https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/deployments/task.yaml.j2#L472-L499) so that we get the benefit of falling back to the global node assignment settings

Also updated the docs!

rooftopcellist (approved):
I tested this out on my cluster and it works as expected as far as I can tell.  No errors on fresh install either.  

@ranvit Thanks for the PR!

ranvit:
@rooftopcellist how do you feel about adding the same tolerations/nodeSelector fallback on the Backup role's [db-management pod](https://github.com/ansible/awx-operator/blob/devel/roles/backup/templates/management-pod.yml.j2)?

I could go through every k8s pod template in the repo and add the same fix

PR #1805, opened 2024-04-02
Add postgres init container to resolve permissions for some k3s deployments

rooftopcellist:
Alternate PR for https://github.com/ansible/awx-operator/pull/1799

Resolves:
* https://github.com/ansible/awx-operator/issues/1770
* https://github.com/ansible/awx-operator/issues/1775

This approach allows users to simply set:

```
spec:
  postgres_data_volume_init: true
```

This removes the unused `postgres_init_container_resource_requirements` parameter which was never removed when the last postgres init container was removed.  I cannot think of a good reason to not just use `postgres_resource_requirements`.

cc @kurokobo 

##### ISSUE TYPE

 - New or Enhanced Feature

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -1459,28 +1459,6 @@ spec:
                         type: string
                     type: object
                 type: object
-              postgres_init_container_resource_requirements:
-                description: Resource requirements for the postgres init container
-                properties:
-                  requests:
-                    properties:
-                      cpu:
-                        type: string
-                      memory:
-                        type: string
-                      storage:
-                        type: string
-                    type: object
-                  limits:
-                    properties:
-                      cpu:
-                        type: string
-                      memory:
-                        type: string
-                      storage:
-                        type: string
-                    type: object
-                type: object
               redis_resource_requirements:
                 description: Resource requirements for the redis container
                 properties:
@@ -1811,6 +1789,12 @@ spec:
                 type: array
                 items:
                   type: string
+              postgres_data_volume_init:
+                description: Sets permissions on the /var/lib/postgresql/data for postgres container using an init container (not Openshift)
+                type: boolean
+              postgres_init_container_commands:
+                description: Customize the postgres init container commands (Non Openshift)
+                type: string
               postgres_extra_volumes:
                 description: Specify extra volumes to add to the application pod
                 type: string
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -1459,6 +1459,28 @@ spec:
                         type: string
                     type: object
                 type: object
+              postgres_init_container_resource_requirements:
+                description: (Deprecated, use postgres_resource_requirements parameter) Resource requirements for the postgres init container
+                properties:
+                  requests:
+                    properties:
+                      cpu:
+                        type: string
+                      memory:
+                        type: string
+                      storage:
+                        type: string
+                    type: object
+                  limits:
+                    properties:
+                      cpu:
+                        type: string
+                      memory:
+                        type: string
+                      storage:
+                        type: string
+                    type: object
+                type: object
               redis_resource_requirements:
                 description: Resource requirements for the redis container
                 properties:
@@ -1790,7 +1812,7 @@ spec:
                 items:
                   type: string
               postgres_data_volume_init:
-                description: Sets permissions on the /var/lib/postgresql/data for postgres container using an init container (not Openshift)
+                description: Sets permissions on the /var/lib/pgdata/data for postgres container using an init container (not Openshift)
                 type: boolean
               postgres_init_container_commands:
                 description: Customize the postgres init container commands (Non Openshift)
--- a/config/manifests/bases/awx-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/awx-operator.clusterserviceversion.yaml
@@ -456,13 +456,16 @@ spec:
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:com.tectonic.ui:resourceRequirements
-      - description: The PostgreSQL init container is not used when an external DB
-          is configured
-        displayName: PostgreSQL Init Container Resource Requirements
-        path: postgres_init_container_resource_requirements
+      - description: Sets permissions on the /var/lib/postgresql/data for postgres container using an init container (not Openshift)
+        displayName: PostgreSQL initialize data volume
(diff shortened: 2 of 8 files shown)


roles/installer/templates/statefulsets/postgres.yaml.j2
```diff
@@ -46,6 +46,27 @@ spec:
 {% endif %}
 {% if postgres_priority_class is defined %}
       priorityClassName: '{{ postgres_priority_class }}'
+{% endif %}
+{% if postgres_data_volume_init is defined and not is_openshift %}
```
kurokobo: Not tested at all on my side, but since `postgres_data_volume_init` has a default value, I think `is define` doesn't work as expected.

```suggestion
{% if postgres_data_volume_init and not is_openshift %}
```

config/manifests/bases/awx-operator.clusterserviceversion.yaml
```diff
@@ -456,13 +456,16 @@ spec:
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:com.tectonic.ui:resourceRequirements
-      - description: The PostgreSQL init container is not used when an external DB
-          is configured
-        displayName: PostgreSQL Init Container Resource Requirements
-        path: postgres_init_container_resource_requirements
+      - description: Sets permissions on the /var/lib/postgresql/data for postgres container using an init container (not Openshift)
```
craph: Is it `/var/lib/postgresql/data` or `/var/lib/pgsql/data` ?
```suggestion
      - description: Sets permissions on the /var/lib/pgsql/data for postgres container using an init container (not Openshift)
```

config/crd/bases/awx.ansible.com_awxs.yaml
```diff
@@ -1811,6 +1789,12 @@ spec:
                 type: array
                 items:
                   type: string
+              postgres_data_volume_init:
+                description: Sets permissions on the /var/lib/postgresql/data for postgres container using an init container (not Openshift)
```
craph: Is it `/var/lib/postgresql/data` or `/var/lib/pgsql/data` ?
```suggestion
                description: Sets permissions on the /var/lib/pgdata/data for postgres container using an init container (not Openshift)
```

roles/installer/templates/statefulsets/postgres.yaml.j2
rooftopcellist: Good catch, that is true.  I'll commit these changes and test it out with the awx-on-k3s scenario.  Thanks @kurokobo and @craph !

rooftopcellist:
@kurokobo I just realized that in your awx-on-k3s repo, there are a couple places where you reference postgres_init_container_resource_requirements

image

If a user has this specified on their AWX spec on upgrade or fresh install, they will get this error:

```
Error from server (BadRequest): error when creating "base": AWX in version "v1beta1" cannot be handled as a AWX: strict decoding error: unknown field "spec.postgres_init_container_resource_requirements"
```

It's probably safer for me to just add back in that parameter and mark it as deprecated, and add a backwards compat shim.  I'll add that now.

fosterseth (approved):

rooftopcellist:
I just added the changes mentioned so that the validator does not error if the user has `postgres_init_container_resource_requirements` defined.  It also accepts that parameter and uses it preferentially over `postgres_resource_requirements` for the init container only if provided.  Though I marked it as deprecated so that we can eventually remove it as there is no real need to have two separate variables for this imo.

**Testing**

I tested this out with the awx-on-k3s repo with the follow AWX spec:

```
---
kind: AWX
metadata:
  name: awx
spec:
  # all of the defaults from https://github.com/kurokobo/awx-on-k3s/blob/main/base/awx.yaml

  postgres_data_volume_init: true
```

I confirmed that the deployment came up successfully, and that the pg init param works:

```
$ kubectl -n awx get pod awx-postgres-15-0 -o yaml | grep resources: -B1
...
    name: init
    resources: {}
```

Then also tested with:

```
---
kind: AWX
metadata:
  name: awx
spec:
  # all of the defaults from https://github.com/kurokobo/awx-on-k3s/blob/main/base/awx.yaml

  # postgres_init_container_resource_requirements: {} #commented out
  postgres_data_volume_init: true
  postgres_init_container_commands: |
    chown 26:0 /var/lib/pgsql/data
    chmod 700 /var/lib/pgsql/data
```

I confirmed that the deployment came up successfully.

Finally, I did one more with `postgres_resource_requirements` set to something custom just to make sure there were no regressions introduced there.

```
spec:
  postgres_resource_requirements:
    requests:
      cpu: 100m
      memory: 256Mi
    limits:
      cpu: 500m
      memory: 512Mi
```

After the deployment finished, I checked that the requests and limits were set correctly.

```
$ kubectl -n awx get pod awx-postgres-15-0 -o yaml | grep resources: -B1 -A6
      protocol: TCP
    resources:
      limits:
        cpu: 500m
        memory: 512Mi
      requests:
        cpu: 100m
        memory: 256Mi
--
    name: init
    resources:
      limits:
        cpu: 500m
        memory: 512Mi
      requests:
        cpu: 100m
        memory: 256Mi
```

rooftopcellist:
As follow-up I opened a PR to update the `main` branch of the awx-on-k3s repo cc @kurokobo 
* https://github.com/kurokobo/awx-on-k3s/pull/330

kurokobo:
@rooftopcellist 
> It's probably safer for me to just add back in that parameter and mark it as deprecated, and add a backwards compat shim. I'll add that now.

Thanks for the confirmation and sending PR, my guide is only unofficial one, so I think my guide should respect the Operator's implementation, and I don't think the Operator needs to adapt my guide. But yes, just marking the param as deprecated is safer. Thanks!

xamindar:
Thank you, adding this finally got that dang pod to work. This needs to be default:
postgres_data_volume_init: true

kcjones91:
```---
kind: AWX
metadata:
  name: awx-demo
spec:
  admin_user: admin
  service_type: nodeport
  postgres_security_context_settings:
  fsGroup: 26
  postgres_data_volume_init: true
  postgres_init_container_commands: |
    chown 26:0 /var/lib/pgsql/data
    chmod 700 /var/lib/pgsql/data
```
I am running into this on my end. Probably an easy fix? 

```
kubectl logs awx-demo-postgres-15-0 -n awx-dev -c init
chown: changing ownership of '/var/lib/pgsql/data': Permission denied
chmod: changing permissions of '/var/lib/pgsql/data': Permission denied
```
```
swipe@swipe-worker-1:/mnt$ ls -la /mnt/data/postgres/
total 0
drwxrwxrwx. 2 swipe swipe  6 Dec  9 15:48 .
drwxr-xr-x. 3 swipe swipe 22 Dec  9 13:17 ..
```

PR #1807, opened 2024-04-03
Fix "external_hostname" should not be required while using Route ingress

TheRealHaoLiu:
##### SUMMARY
Fix "external_hostname" should not be required while using Route ingress

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/roles/mesh_ingress/templates/ingress.yml.j2
+++ b/roles/mesh_ingress/templates/ingress.yml.j2
@@ -70,7 +70,7 @@ metadata:
   name: {{ ansible_operator_meta.name }}
   namespace: {{ ansible_operator_meta.namespace }}
 spec:
-{% if external_hostname %}
+{% if external_hostname is defined %}
   host: {{ external_hostname }}
 {% endif %}
   port:


fosterseth:
we have a task that sets the `external_hostname` if ingress_type is "route"

```yaml
- name: Set external_hostname
  set_fact:
    external_hostname: "{{ ingress.result.status.ingress[0].host }}"
  when: ingress_type | lower == 'route'
```

and in the jinja2 template it is only referenced under a `{% if ingress_type|lower == "route" %}` conditional

so not sure in which situation this wouldn't be defined when referenced.

fosterseth (approved):

TheRealHaoLiu:
> we have a task that sets the `external_hostname` if ingress_type is "route"
> 
> ```yaml
> - name: Set external_hostname
>   set_fact:
>     external_hostname: "{{ ingress.result.status.ingress[0].host }}"
>   when: ingress_type | lower == 'route'
> ```
> 
> and in the jinja2 template it is only referenced under a `{% if ingress_type|lower == "route" %}` conditional
> 
> so not sure in which situation this wouldn't be defined when referenced.

the task execute after the route is created

so bug affect all case when ingress_type=route

fosterseth:
ah didn't realize the ordering there, got it

PR #1809, opened 2024-04-04
Add postgres_pod_security_context

hhk7734:
##### SUMMARY

`security_context_settings` set web/task PodSecurityContext, but `postgres_security_context_settings` set postgresql container SecurityContext. This confuses me. So I add `postgres_pod_security_context`

##### ISSUE TYPE

 - New or Enhanced Feature
 - Docs Fix

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/roles/installer/templates/statefulsets/postgres.yaml.j2
+++ b/roles/installer/templates/statefulsets/postgres.yaml.j2
@@ -47,6 +47,10 @@ spec:
 {% if postgres_priority_class is defined %}
       priorityClassName: '{{ postgres_priority_class }}'
 {% endif %}
+{% if postgres_pod_security_context %}
+      securityContext:
+        {{ postgres_pod_security_context | to_nice_yaml | indent(width=8) }}
+{% endif %}
 {% if postgres_data_volume_init and not is_openshift %}
       initContainers:
         - name: init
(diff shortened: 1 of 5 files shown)


roles/installer/templates/statefulsets/postgres.yaml.j2
```diff
@@ -47,6 +47,10 @@ spec:
 {% if postgres_priority_class is defined %}
       priorityClassName: '{{ postgres_priority_class }}'
 {% endif %}
+{% if postgres_pod_security_context %}
+      securityContext:
+        {{ postgres_pod_security_context | to_nice_yaml | indent(width=8) }}
+{% endif %}
```
rooftopcellist, line 53: @hhk7734 It looks like you added the `postgres_pod_security_context` entry at the statefulSet level, so it would affect the init container and the postgres container.  Is there is use case for having a separate parameter for setting the security context setting for only the postgres container?  

I am thinking we could effectively just move the existing `postgres_security_context` entry here (instead of on the postgres container).

hhk7734: I used `PodSecurityContext.fsGroup=26` when trying to solve an issue with https://github.com/ansible/awx-operator/pull/1805.

1805 PR was already merged. There are no longer any situations where this PR should be used. Should I close this PR?

rooftopcellist: @hhk7734 yep, I agree.  I'll go ahead and close.  Thanks!

PR #1821, opened 2024-04-11
Unified Annotation Handling without Duplicates

alexiseichst:
The objective here is to manage multiple entries and unify **web_annotations** and **annotations** while avoiding duplicates.

##### ISSUE TYPE

### DIFF
--- a/roles/installer/templates/deployments/web.yaml.j2
+++ b/roles/installer/templates/deployments/web.yaml.j2
@@ -44,11 +44,14 @@ spec:
   ] %}
         checksum-secret-{{ secret }}: "{{ lookup('ansible.builtin.vars', secret, default='')["resources"][0]["data"] | default('') | sha1 }}"
 {% endfor %}
-{% if web_annotations %}
-        {{ web_annotations | indent(width=8) }}
-{% elif annotations %}
-        {{ annotations | indent(width=8) }}
-{% endif %}
+{% set combined_annotations = web_annotations + annotations %}
+{% set seen = [] %}
+{% for annotation in combined_annotations %}
+    {% if annotation not in seen %}
+        {{ annotation | indent(width=8) }}
+        {% do seen.append(annotation) %}
+    {% endif %}
+{% endfor %}
     spec:
 {% if uwsgi_listen_queue_size is defined and uwsgi_listen_queue_size|int > 128 %}
       securityContext:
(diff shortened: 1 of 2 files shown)


roles/installer/templates/deployments/web.yaml.j2
```diff
@@ -44,11 +44,14 @@ spec:
   ] %}
         checksum-secret-{{ secret }}: "{{ lookup('ansible.builtin.vars', secret, default='')["resources"][0]["data"] | default('') | sha1 }}"
 {% endfor %}
-{% if web_annotations %}
-        {{ web_annotations | indent(width=8) }}
```
R34SGA, line 48: This case was only accepting one single annotation, In our case we need multiple lines

roles/installer/templates/deployments/web.yaml.j2
```diff
@@ -44,11 +44,14 @@ spec:
...
-{% elif annotations %}
-        {{ annotations | indent(width=8) }}
```
R34SGA, line 50: even worse that 'elif' case was completely overided by web_annotation...

roles/installer/templates/deployments/web.yaml.j2
```diff
@@ -44,11 +44,14 @@ spec:
...
-{% endif %}
+{% set combined_annotations = web_annotations + annotations %}
+{% set seen = [] %}
+{% for annotation in combined_annotations %}
```
R34SGA, line 49: very elegant way to factorize code with minimum impact

R34SGA (approved):
Nicely done @alexiseichst  !

kurokobo:
Could you please clarify your first issue, and provide actual AWX CR that can reproduce your issue? How you've tested this PR?
Also, is there any specific reason why you don't make any changes for `task_annotations`?

rooftopcellist (approved):
Thanks for the PR @alexiseichst 

My only note is that it might be nice to have a note in the docs here about how the annotations will now be merged:
* https://github.com/ansible/awx-operator/blob/c7c7171110c90fa65435294b6c063410b15fca04/docs/user-guide/advanced-configuration/assigning-awx-pods-to-specific-nodes.md?plain=1#L31

And also add entries for `task_annotations` and `web_annotations` in the doc.

Feel free to add this in a follow-up PR if you agree.

rooftopcellist (changes requested):
These changes should be done for the task.yaml.j2 file as well.

rooftopcellist:
@alexiseichst I think we should add these same changes to the task deployment before merging.  And CI is currently failing.  This PR needs more attention before it can be merged.

PR #1824, opened 2024-04-13
docs: add a tip to add extra settings through configmaps or secrets and improve an example for that

kurokobo:
##### SUMMARY

The page "[Custom volume and volume mount options](https://ansible.readthedocs.io/projects/awx-operator/en/latest/user-guide/advanced-configuration/custom-volume-and-volume-mount-options.html)" contains an example of mounting `custom.py` under `/etc/tower/conf.d/` that contains extra settings for AWX.

This is very handy way if users' `exta_settings` is too big, but there is less information about this way. I think the page for `exta_settings` should contain this alternative handy way.

This PR changes:

- Add a tip to provide alternative way to add extra settings by mounting configmaps or secrets of `*.py` files.
- Add `web_extra_volume_mounts` in the example in `custom-volume-and-volume-mount-options.md` to ensure `custom.py` is mounted on both task and web pods.
- Fix minor issues
  - Correct indentation for the code blocks in `custom-volume-and-volume-mount-options.md`.
  - Correct `warning` block in `custom-volume-and-volume-mount-options.md`.

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

**The main setting file `/etc/tower/settings.py` is also mounted on the migration job pods and the metrics utility job pods, and currently there is no way to add custom volumes to migration job pod and metrics utility job pods.**

**Mounting `custom.py` may cause inconsistencies between the setting files in the task/web pods and migration/metrics utility pods, but I do not expect that `custom.py` will provide any settings that would affect migrations or metrics utilities. Please let me know if I am wrong (in which case I should rather remove the `custom.py` from this examples).**

### DIFF
--- a/docs/user-guide/advanced-configuration/custom-volume-and-volume-mount-options.md
+++ b/docs/user-guide/advanced-configuration/custom-volume-and-volume-mount-options.md
@@ -26,59 +26,64 @@ metadata:
   namespace: <target namespace>
 data:
   ansible.cfg: |
-     [defaults]
-     remote_tmp = /tmp
-     [ssh_connection]
-     ssh_args = -C -o ControlMaster=auto -o ControlPersist=60s
+    [defaults]
+    remote_tmp = /tmp
+    [ssh_connection]
+    ssh_args = -C -o ControlMaster=auto -o ControlPersist=60s
   custom.py:  |
-      INSIGHTS_URL_BASE = "example.org"
-      AWX_CLEANUP_PATHS = True
+    INSIGHTS_URL_BASE = "example.org"
+    AWX_CLEANUP_PATHS = True
 ```
 Example spec file for volumes and volume mounts
 
 ```yaml
 ---
-    spec:
-    ...
-      extra_volumes: |
-        - name: ansible-cfg
-          configMap:
-            defaultMode: 420
-            items:
-              - key: ansible.cfg
-                path: ansible.cfg
-            name: <resourcename>-extra-config
-        - name: custom-py
-          configMap:
-            defaultMode: 420
-            items:
-              - key: custom.py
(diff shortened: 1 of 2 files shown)


kurokobo:
Extra settings:
image

Custom volume and volume mount options:
image

rooftopcellist (approved):

rooftopcellist:
@kurokobo Thank you for contributing these docs!  

As an aside, it would be nice to make this easier for users.  Maybe we could add a `settings_file_configmap` parameter that would automatically mount the contents of the configmap specified into the respective containers.  

Generally, we try not to add unnecessary new parameters, but this is a common enough use case and would say users the trouble of double quotes issues on values for the existing `extra_settings` param implementation.  

What do you think about that idea?

kurokobo:
@rooftopcellist 
Hi, thanks for reviewing and merging.

> As an aside, it would be nice to make this easier for users. Maybe we could add a `settings_file_configmap` parameter that would automatically mount the contents of the configmap specified into the respective containers.

Yes, this is exactly one of the things I was going to draft as an RFE. I already made quick investigation and it seems to that perhaps we can implement almost same things on Galaxy Operator and EDA Server Operator as well.
On both galaxy_ng and eda-server, there are no predefined configuration directory (like `/etc/tower/conf.d` for AWX), but both of them already have django-split-settings that can include multiple files into main settings file. So I didn't test at all, but I think,

- For Galaxy Operator. we may be able to add `include` to load custom settings files into `settings.py` that generated by Galaxy Operator
- For EDA Server Operator, we may be able to define `EDA_SETTINGS_FILE` with a python file that generated by EDA Server Operator that contains `include`

For now, let me make an RFE for AWX Operator.

PR #1800, opened 2024-03-28
Add Bundle CA Cert volumeMount to migrations job pod

rooftopcellist:
##### SUMMARY
Fixes: https://github.com/ansible/awx-operator/issues/1782

The bundle ca cert needed to be mounted into the k8s migration job pod for ssl support.  

##### ISSUE TYPE

 - Bug, Docs Fix or other nominal change

PR #1846, opened 2024-04-30
Fix custom CA certificates for task/web/migration

### LINKED ISSUE #1782: AWX database migration job fail when using custom CA certificate

### Bug Summary

During the upgrade to awx 24.0.0, a new pod is created for migration. This pod doesnt contain the custom ca certificate and fails when using external postgres with verify enabled.

### AWX Operator version

2.13.1

### AWX version

24.0.0

### Kubernetes platform

kubernetes

1.27.11

### Modifications

no

### Steps to reproduce

Update from awx 23.8.1 to 24.0.0
Add custom ca certs with external postgres. 

### Expected results

migration pod run with custom ca certs added and migration finishes in task pods.

### Actual results

no custom ca certs added and pod failing with django.db.utils.OperationalError: connection failed: certificate verify failed

### Additional information

  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/psycopg/connection.py", line 728, in connect
    raise ex.with_traceback(None)
django.db.utils.OperationalError: connection failed: certificate verify failed

### Operator Logs

_No response_

JoelKle:
##### SUMMARY
This PR fixes the usage of custom CA certificates in the migration job and improves the existing implementation in the task & web deployment.
During the upgrade to awx 24.0.0, a new pod is created for migration. This pod doesnt contain the custom ca certificate and fails when using external postgres with verify enabled.

PR fixes: https://github.com/ansible/awx-operator/issues/1782
Follow up / Improves: https://github.com/ansible/awx-operator/pull/1800

##### ISSUE TYPE

##### ADDITIONAL INFORMATION
This PR is based on the great work of @YassineFadhlaoui in https://github.com/ansible/awx-operator/issues/1782#issuecomment-2072044985 and @akkaba23 in https://github.com/ansible/awx-operator/pull/1800#issuecomment-2031485996

The following has changed:
- Added a new init container `init-bundle-ca-trust` to the `task` + `web` deployment and to the `migration` job
  - the init container runs the `update-ca-trust extract` if `bundle_ca_crt` is set.
  - I've added a new init container for the following reasons:
    - this init container need to run before all other init containers
    - only this init container requires `runAsUser: 0`

- Removed the `update-ca-trust` command from the init container `init-receptor` because it will run once in the new init container `init-bundle-ca-trust`

- Removed the mounting of the volume `{{ ansible_operator_meta.name }}-bundle-cacert` from containers that really do not need it

- Added the whole `bundle_ca_crt` logic to the `migration` job

I've successfully tested that change during my upgrade from awx-operator v2.12.1 to v2.15.0

### DIFF
--- a/roles/installer/templates/deployments/task.yaml.j2
+++ b/roles/installer/templates/deployments/task.yaml.j2
@@ -74,6 +74,28 @@ spec:
       priorityClassName: '{{ control_plane_priority_class }}'
 {% endif %}
       initContainers:
+{% if bundle_ca_crt  %}
+        - name: init-bundle-ca-trust
+          image: '{{ _init_container_image }}'
+          imagePullPolicy: '{{ image_pull_policy }}'
+          resources: {{ init_container_resource_requirements }}
+          securityContext:
+            runAsUser: 0
+            runAsGroup: 0
+          command:
+            - /bin/sh
+            - -c
+            - |
+              mkdir -p /etc/pki/ca-trust/extracted/{java,pem,openssl,edk2}
+              update-ca-trust extract
+          volumeMounts:
+            - name: "ca-trust-extracted"
+              mountPath: "/etc/pki/ca-trust/extracted"
+            - name: "{{ ansible_operator_meta.name }}-bundle-cacert"
+              mountPath: /etc/pki/ca-trust/source/anchors/bundle-ca.crt
+              subPath: bundle-ca.crt
+              readOnly: true
+{% endif %}
         - name: init-database
           image: '{{ _image }}'
           imagePullPolicy: '{{ image_pull_policy }}'
@@ -96,6 +118,10 @@ spec:
               subPath: settings.py
               readOnly: true
             {{ lookup("template", "common/volume_mounts/extra_settings_files.yaml.j2")  | indent(width=12) | trim }}
+{% if bundle_ca_crt  %}
+            - name: "ca-trust-extracted"
+              mountPath: "/etc/pki/ca-trust/extracted"
+{% endif %}
 {% if development_mode | bool %}
             - name: awx-devel
               mountPath: "/awx_devel"
@@ -123,10 +149,6 @@ spec:
                 outcert=/etc/receptor/tls/receptor.crt \
                 notafter=$(date --iso-8601=seconds --utc --date "10 years") \
                 verify=yes
-{% if bundle_ca_crt  %}
-              mkdir -p /etc/pki/ca-trust/extracted/{java,pem,openssl,edk2}
-              update-ca-trust
-{% endif %}
 {% if init_container_extra_commands %}
               {{ init_container_extra_commands | indent(width=14) }}
 {% endif %}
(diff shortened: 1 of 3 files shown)


akkaba23:
Thank you for the nice work done on this PR @JoelKle .

JoelKle:
@rooftopcellist @fosterseth Could you have a look on this PR? Thx :)

roles/installer/templates/deployments/web.yaml.j2
```diff
@@ -80,31 +80,43 @@ spec:
...
-{% if bundle_ca_crt or projects_persistence|bool or init_container_extra_commands %}
-        - name: init
+{% if bundle_ca_crt  %}
+        - name: init-bundle-ca-trust
           image: '{{ _init_container_image }}'
           imagePullPolicy: '{{ image_pull_policy }}'
           resources: {{ init_container_resource_requirements }}
+          securityContext:
+            runAsUser: 0
+            runAsGroup: 0
```
rooftopcellist: @JoelKle This will break on Openshift.  Can we template this in only if is_k8s=true?

roles/installer/templates/deployments/task.yaml.j2
```diff
@@ -74,6 +74,28 @@ spec:
...
 {% endif %}
       initContainers:
+{% if bundle_ca_crt  %}
+        - name: init-bundle-ca-trust
+          image: '{{ _init_container_image }}'
+          imagePullPolicy: '{{ image_pull_policy }}'
+          resources: {{ init_container_resource_requirements }}
+          securityContext:
+            runAsUser: 0
+            runAsGroup: 0
```
rooftopcellist: @JoelKle This will break on Openshift. Can we template this in only if is_k8s=true?

roles/installer/templates/deployments/web.yaml.j2
rooftopcellist: Did `update-ca-trust extract` require more permissions than just `update-ca-trust`?  I am trying to understand why the securityContext bit was added.

JoelKle: I removed the whole `securityContext` block from this PR. Now the init-container behaves as before.
@rooftopcellist Is this fine for you?

TL;DR
Withouth the `runAsUser: 0` parameter the init-container runs the update-ca-trust command as `uid=1000(awx) gid=0(root)`. That means the files in the folder `/etc/pki/ca-trust/extracted` are owned by `awx:root`
With the `runAsUser: 0` parameter all files are owned by `root:root` which is more accurate.

JoelKle: The `update-ca-trust extract` behaves the same as withouth the `extract`. However with the `extract` the command may print more warnings, ...
Check the manpage of update-ca-trust extract > "COMMANDS"

JoelKle:
@rooftopcellist Are there any other blocker for you?

rooftopcellist:
@JoelKle thanks for following up, and sorry for the delay here.  Could you please rebase to resolve the conflicts?  We'll see if CI passes on the run kicked off by pushing the rebased branch.  

I just checked the CI failure, and it looks like it was flake, so once rebased I think we will be good.

JoelKle:
@rooftopcellist Rebase done. Look like the CI run needs your approval?

rooftopcellist (approved):

PR #1847, opened 2024-04-30
Add postgresql option target_session_attrs

### LINKED ISSUE #1792: Support for external PostgreSQL cluster client-side High Availability

### Feature Summary

Add ability to deploy an AWX instance with an external PostgreSQL cluster using client-side High Availability mode.

According to the [PostgreSQL documentation](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-TARGET-SESSION-ATTRS), when using client-side High Availability, you must change the `target_session_attrs` default value `any` to `read-write` in order to select the primary postgreSQL server.

To allow such behavior, I suggest the following :

- Adjusting the Jinja template [credentials.py.j2](https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/settings/credentials.py.j2) to allow target_session_attrs option ;
- Add `awx_postgres_target_session_attrs` var in [database_configuration.yml](https://github.com/ansible/awx-operator/blob/devel/roles/installer/tasks/database_configuration.yml) ;
- Update [database-configuration.md](https://github.com/ansible/awx-operator/blob/devel/docs/user-guide/database-configuration.md) documentation.

larsux:
##### SUMMARY
Added support for the postgresql option 'target_session_attrs'. This is required if you have a postgreSQL database cluster and want to ensure that AWX only connects to the primary database node with write permissions.

This change is all the more necessary as this option cannot be passed in `custom.py` as the order in which the configuration files are loaded is not stable. It is therefor not possible to guarantee, that the `DATABASES` dictionary is already defined or will be overwritten later.

PR fixes #1792 

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
The following has been added:
* add `target_session_attrs` to the database credentials template in the ansible installer role
* add `target_session_attrs` to the postgresql config template in the helm chart

I've successfully tested that change with `molecule/kind` and in our own K8s Cluster with postgreSQL-Cluster database backend.

### DIFF
--- a/roles/installer/templates/settings/credentials.py.j2
+++ b/roles/installer/templates/settings/credentials.py.j2
@@ -11,6 +11,9 @@ DATABASES = {
 {% if awx_postgres_sslmode in ['verify-ca', 'verify-full'] %}
                      'sslrootcert': '{{ ca_trust_bundle }}',
 {% endif %}
+{% if awx_postgres_target_session_attrs %}
+                     'target_session_attrs': '{{ awx_postgres_target_session_attrs }}',
+{% endif %}
         },
     }
 }
@@ -25,6 +28,9 @@ LISTENER_DATABASES = {
             'keepalives_count': {{ postgres_keepalives_count }},
 {% else %}
             'keepalives': 0,
+{% endif %}
+{% if awx_postgres_target_session_attrs %}
+            'target_session_attrs': '{{ awx_postgres_target_session_attrs }}',
 {% endif %}
         },
     }
--- a/.helm/starter/templates/postgres-config.yaml
+++ b/.helm/starter/templates/postgres-config.yaml
@@ -12,6 +12,7 @@ stringData:
   username: {{ .username }}
   password: {{ .password }}
   sslmode: {{ .sslmode }}
+  target_session_attrs: {{ .target_session_attrs | default "any" }}
   type: {{ .type }}
 type: Opaque
 {{- end }}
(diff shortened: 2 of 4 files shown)


roles/installer/templates/settings/credentials.py.j2
```diff
@@ -10,6 +10,9 @@ DATABASES = {
         'OPTIONS': { 'sslmode': '{{ awx_postgres_sslmode }}',
 {% if awx_postgres_sslmode in ['verify-ca', 'verify-full'] %}
                      'sslrootcert': '{{ ca_trust_bundle }}',
+{% endif %}
+{% if awx_postgres_target_session_attrs %}
```
fosterseth, line 14: we probably want to check if defined, and also not empty string

roles/installer/templates/settings/credentials.py.j2
```diff
@@ -25,6 +28,9 @@ LISTENER_DATABASES = {
             'keepalives_count': {{ postgres_keepalives_count }},
 {% else %}
             'keepalives': 0,
+{% endif %}
+{% if awx_postgres_target_session_attrs %}
```
fosterseth, line 32: same here

roles/installer/templates/settings/credentials.py.j2
fosterseth: nvm this is not needed

rooftopcellist (approved):

fosterseth (approved):

rooftopcellist:
Thanks for the contribution @larsux

PR #1853, opened 2024-05-10
node assignment configs are attached on all k8s workloads

ranvit:
##### SUMMARY
The db migration pod and the backup process pod have failed to schedule on a k8s cluster where every node has a taint. So this PR adds the task_* related node-assignment params and the global node-assignment params as fallback to every k8s workload that didn't have node-assignment configuration available on them.

This fixes issues like #1774

##### ISSUE TYPE

##### ADDITIONAL INFORMATION
I PR'd a similar fix with https://github.com/ansible/awx-operator/pull/1804, only targeting the migration pod. Realised I need somethign similar on the backup pod.

The task_* params seem well suited for most of the k8s workloads except for the mesh-ingress deployment, so I had that only use the global params

### DIFF
--- a/roles/backup/templates/management-pod.yml.j2
+++ b/roles/backup/templates/management-pod.yml.j2
@@ -20,9 +20,33 @@ spec:
     resources:
       {{ backup_resource_requirements | to_nice_yaml(indent=2) | indent(width=6, first=False) }}
 {%- endif %}
-{% if db_management_pod_node_selector %}
+{% if task_node_selector %}
   nodeSelector:
-        {{ db_management_pod_node_selector | indent(width=8) }}
+    {{ task_node_selector | indent(width=4) }}
+{% elif node_selector %}
+  nodeSelector:
+    {{ node_selector | indent(width=4) }}
+{% endif %}
+{% if task_topology_spread_constraints %}
+  topologySpreadConstraints:
+    {{ task_topology_spread_constraints | indent(width=4) }}
+{% elif topology_spread_constraints %}
+  topologySpreadConstraints:
+    {{ topology_spread_constraints | indent(width=4) }}
+{% endif %}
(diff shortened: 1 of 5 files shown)


roles/backup/templates/management-pod.yml.j2
```diff
@@ -20,9 +20,33 @@ spec:
     resources:
       {{ backup_resource_requirements | to_nice_yaml(indent=2) | indent(width=6, first=False) }}
 {%- endif %}
-{% if db_management_pod_node_selector %}
+{% if task_node_selector %}
```
rooftopcellist, line 23: @ranvit The `task_node_selector` parameter only exists in the installer role (AWX CR), not in the back role (AWXBackup CR).  

So db_management_pod_node_selector is correct.  

If you need topologySpreadConstraints, tolerations, and affinity as well, those params would need to be added to the AWXBackup CR and should be prefixed with `db_management_pod_*`

rooftopcellist: This PR also needs a rebase now to resolve conflicts.

djyasin:
Hello @ranvit, please let us know if you are still interested in merging this PR. If so, please be sure to address the comments left by @rooftopcellist and resolve any merge conflicts. 

Thank you for your time!

victorfernpach:
Hello!, 
We are in the same situation that @ranvit , we need a way to run the xxxx-db-management pods generated by AWXBackup object in a cluster where all the nodes have taints. Its possible to merge this PR? Thank you in advance.

ranvit:
@victorfernpach sorry but I'm not available to iterate on this PR. maybe you can address the comments on this PR and close it out?

victorfernpach:
I will create a new PR to try to address the comments left by @rooftopcellist.

PR #1854, opened 2024-05-10
Assign ownership of backup pvc to postgres image's uid

### LINKED ISSUE #1830: AWXBackup: mkdir: cannot create directory '/backups/tower-openshift-backup-2024-04-17-141257': Permission denied

### Bug Summary

Upgraded awx operator from 1.1.4 to 2.13.1 and started to get issues when trying to take backups.
Here`s an example of the AWXBackup I have:

kind: AWXBackup
metadata:
  name: awx-demo
  namespace: awx-test
spec:
  deployment_name: awx-demo
  backup_pvc: 'backup-pvc'
  no_log: false
  
Once applied, operator tries to create a folder for the backup on the db-management pod. However, its getting the issue permission denied

[backup : Set backup directory name] **************************************\r\ntask path: /opt/ansible/roles/backup/tasks/postgres.yml:55\nok: [localhost] => {\"ansible_facts\": {\"backup_dir\": \"/backups/tower-openshift-backup-2024-04-17-141257\"}, \"changed\": false}\n\r\nTASK [backup : Create directory for backup] ************************************\r\ntask path: /opt/ansible/roles/backup/tasks/postgres.yml:59\nansible.cfg.\nfatal: [localhost]: FAILED! => {\"changed\": true, \"rc\": 1, \"return_code\": 1, \"stderr\": \"mkdir: cannot create directory '/backups/tower-openshift-backup-2024-04-17-141257': Permission denied\\n\", \"stderr_lines\": [\"mkdir: cannot create directory '/backups/tower-openshift-backup-2024-04-17-141257': Permission denied\"]

### AWX Operator version

2.13.1

### AWX version

24.0.0

### Kubernetes platform

kubernetes

microk8s v1.28.8

### Modifications

no

### Steps to reproduce

Fresh installation and trying to create a backup using AWXBackup CR.

### Expected results

Take the backup successfully 

### Actual results

Failed backup

### Additional information

_No response_

### Operator Logs

_No response_

ranvit:
##### SUMMARY
Backups are failing because the postgres image runs as uid 26 that doesn't have perms to the backup PVC. This fixes #1830 

##### ISSUE TYPE

##### ADDITIONAL INFORMATION
The other option is to have the db-management pod run as root (uid 0), but I dont think thats the preferred solution.

### DIFF
--- a/roles/backup/templates/management-pod.yml.j2
+++ b/roles/backup/templates/management-pod.yml.j2
@@ -7,6 +7,15 @@ metadata:
   labels:
     {{ lookup("template", "../common/templates/labels/common.yaml.j2") | indent(width=4) | trim }}
 spec:
+  initContainers:
+    - name: init-pvc-chown
+      image: busybox
+      # _postgres_image runs as uid 26
+      command: ["sh", "-c", "chown -R :26 /backups && chmod -R 770 /backups"]
+      volumeMounts:
+        - name: {{ ansible_operator_meta.name }}-backup
+          mountPath: /backups
+          readOnly: false
   containers:
   - name: {{ ansible_operator_meta.name }}-db-management
     image: "{{ _postgres_image }}"


kurokobo:
@ranvit 
Hi, thanks for working on this!

Note that AFAIK changing owner and permission for PVC is NOT required for all users, but only required for specific users e.g. on hostPath or longhorn on k8s. So adding init container with root privilege _by default_ may cause side effects for those who are on the k8s or OpenShift where the issue does not exist.

There is already similar implementation for PVC for PSQL to change owner and permission, so I think it would be better to follow existing implementation instead of going different way. How do you think? The exising implementation is having a flag to enable init container to invoke specific commands. Refer to: https://github.com/ansible/awx-operator/pull/1805

fosterseth:
As @kurokobo mentioned, we should make this init container optional, following the same pattern we did in #1805

also we should use postgres image instead of busybox, try to use same variable names as other postgres init implementation, etc..

Raptus1:
Jumping on this because of a migration test using AWXBackup and AWXRestore for the first time migrating from pg 13 to 15 from Operator 2.11.0 to 2.18.0 and AWX from 23.7.0 to 24.5.0.

Migration itself is not working (too big for a leap?) so once the migration is done from 13 to 15, I end up with a empty AWX deploy hence the effort on the backup/restore through the operator, but maybe this is the exact cause of the problem that causing the migration to not go well:


 TASK [Get AWX object definition from pvc] ******************************** 
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to copy file from Pod: cat: /backups/tower-openshift-backup-2024-06-24-204806/awx_object: Permission denied\n"}

https://github.com/ansible/awx-operator/blob/devel/roles/restore/tasks/import_vars.yml#L11

Running the deploy from another AWX instance, so not so easy to just add the workaround mentioned above. At least I'm not familiar with how I can change the permissions needed.

dangoncalves:
Hello @ranvit : are you still working on this PR?

PR #1857, opened 2024-05-15
Create _metrics_utility_secret for metrics utility

dhageman:
##### SUMMARY
The template for the metrics utility cronjobs references  the variable _metrics_utility_secret.  This variable is missing from the list of variables constructed for the utility.

##### ISSUE TYPE


rooftopcellist (approved):

rooftopcellist (approved):
Without this, the `_metrics_utility_secret` param would not be set from  `metrics_utility_secret`. 

It will fail if the metrics_utility_secret or default name for that secret does not exist.  We could potentially add a preflight check for that in a follow-up PR. But not required.

PR #1858, opened 2024-05-16
Update k8s_exec and k8s_cp to include container name

fluzzykitten:
##### SUMMARY
We need to specify a container in environments that use sidecar injection, like in the case of istio service mesh. If the container is not specified, and a side car is injected so there are multiple containers running in the pod, this task will fail because a container was not specified in a pod with multiple containers.

##### ISSUE TYPE
 Bug preventing backups in environments that use Istio service mesh or other automatic side-car injection

##### ADDITIONAL INFORMATION
Container that we create is the same name as the pod, so we just need to tell the backup to use the container we created, to differentiate it from any sidecars containers that might've been injected

### DIFF
--- a/roles/backup/tasks/awx-cro.yml
+++ b/roles/backup/tasks/awx-cro.yml
@@ -31,5 +31,6 @@
   k8s_cp:
     namespace: "{{ backup_pvc_namespace }}"
     pod: "{{ ansible_operator_meta.name }}-db-management"
+    container: "{{ ansible_operator_meta.name }}-db-management"
     remote_path: "{{ backup_dir  }}/awx_object"
     content: "{{ awx_spec | to_yaml }}"
--- a/roles/backup/tasks/delete_backup.yml
+++ b/roles/backup/tasks/delete_backup.yml
@@ -3,5 +3,6 @@
   k8s_exec:
     namespace: "{{ backup_pvc_namespace }}"
     pod: "{{ ansible_operator_meta.name }}-db-management"
+    container: "{{ ansible_operator_meta.name }}-db-management"
     command: >-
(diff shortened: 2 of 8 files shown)


rooftopcellist:
Good point, thanks for the PR @fluzzykitten !

I checked that the container name is indeed correct:
* https://github.com/ansible/awx-operator/blob/devel/roles/backup/templates/management-pod.yml.j2#L11

However, I expect you will also need to specify the container here as well right @fluzzykitten ?
* https://github.com/ansible/awx-operator/blob/e98c913f864ad26c2a81cdba77b7ffdf97c13282/roles/backup/tasks/awx-cro.yml#L33
* https://github.com/ansible/awx-operator/blob/e98c913f864ad26c2a81cdba77b7ffdf97c13282/roles/backup/tasks/postgres.yml#L128


Actually, pretty much all of these places:
* https://github.com/search?q=repo%3Aansible%2Fawx-operator+pod%3A+%22%7B%7B+ansible_operator_meta.name+%7D%7D-db-management%22&type=code

rooftopcellist:
While looking at this PR, I noticed this typo:
* https://github.com/ansible/awx-operator/pull/1859

fluzzykitten:
@rooftopcellist Shoot! You're right, I only pushed one of the files. My bad. MR updated with the other instances of the `k8s_exec` and `k8s_cp` modules that didn't set a container, and would be affected by sidecar container injection.

rooftopcellist:
@fluzzykitten thanks for the quick update!  I just tested out a backup and restore and didn't run into any issues.  Thanks for the contribution!

rooftopcellist (approved):

PR #1875, opened 2024-05-27
fix: Make basic install without operator works

chicuongha:
##### SUMMARY

Following the roles/installer/README.md, AWX installation without operator not working.
Operator using the same role, we should fixed it to make it works without operator.
Most errors are undefined variables, or tasks related to the AWX component updates.

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

  - Update role name for README.md
  - Avoid the this_awx['resources'][0] is undefined in database_configuration.yml
  - Add update_status variable to include or not the update_status.yml
  - metrics_utility_enabled exists in CRD but not as variable

With the fix, I can deploy AWX in k8s with ansible-playbook command.

### DIFF
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -443,6 +443,9 @@ ldap_password_secret: ''
 # Secret to lookup that provides the custom CA trusted bundle
 bundle_cacert_secret: ''
 
+# Set false for basic install without operator
+update_status: true
+
 # Whether secrets should be garbage collected
 # on teardown
 #
--- a/docs/troubleshooting/debugging.md
+++ b/docs/troubleshooting/debugging.md
@@ -96,6 +96,8 @@ Create a vars file:
 ansible_operator_meta:
   name: awx
   namespace: awx
+set_self_labels: false
+update_status: false
 service_type: nodeport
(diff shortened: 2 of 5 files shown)


roles/installer/defaults/main.yml
```diff
@@ -443,6 +443,9 @@ ldap_password_secret: ''
 # Secret to lookup that provides the custom CA trusted bundle
 bundle_cacert_secret: ''
 
+# Set false for basic install without operator
+update_status: true
```
rooftopcellist, line 447: It would be nice to add and example of how to use this to the debugging docs that cover how to use this flow.  
* https://github.com/ansible/awx-operator/blob/devel/docs/troubleshooting/debugging.md#iterating-on-the-installer-without-deploying-the-operator

Could you please update that @chicuongha ?  Otherwise this change looks great, thanks!

chicuongha: @rooftopcellist  documentation file updated

rooftopcellist (approved):

rooftopcellist:
I just resolved the conflicts and approved the PR.  It should merge shortly.  Thanks for the change!

PR #1878, opened 2024-05-28
Ability to add custom labels to the DB migration pods

sbilque:
##### SUMMARY
This PR allows to add the labels defined in the `additional_labels` parameter to the DB migration pods.

Useful when labels are used to control pod creation based on label rules.

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
Currently, labels are added at the DB migration job level, but not in the migration pod associated with that job.
The creation of this pod may be blocked if, for example, there is a control system based on labels.

This change simply reproduces what exists in the definition of task and web pods.


rooftopcellist (approved):
Thanks for the contribution!

PR #1889, opened 2024-06-04
installer: db-migrate: add support for resource requirements

sinuscosinustan:
##### SUMMARY

Namespaces with an active ResourceQuota can't spawn a migration job pod, because there's no option available to set a resource request.

##### ISSUE TYPE
 - New or Enhanced Feature

### DIFF
--- a/roles/installer/templates/jobs/migration.yaml.j2
+++ b/roles/installer/templates/jobs/migration.yaml.j2
@@ -40,6 +40,7 @@ spec:
             - awx-manage
             - migrate
             - --noinput
+          resources: {{ migration_job_resource_requirements }}
           volumeMounts:
             - name: {{ ansible_operator_meta.name }}-application-credentials
               mountPath: "/etc/tower/conf.d/credentials.py"
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -1462,6 +1462,32 @@ spec:
                         type: string
                     type: object
                 type: object
+              migration_job_resource_requirements:
+                description: Resource requirements for the django migration job container
+                properties:
+                  requests:
+                    properties:
+                      cpu:
+                        type: string
+                      memory:
+                        type: string
+                      storage:
+                        type: string
+                      ephemeral-storage:
+                        type: string
+                    type: object
+                  limits:
+                    properties:
+                      cpu:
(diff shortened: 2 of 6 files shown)


roles/installer/templates/jobs/migration.yaml.j2
```diff
@@ -16,6 +20,7 @@ spec:
             - awx-manage
             - migrate
             - --noinput
+          resources: {{ migration_job_resource_requirements }}
```
fosterseth, line 43: for users who are upgrading and might not have `migration_job_resource_requirements` defined, can we fallback to using `task_resource_requirements`?

sinuscosinustan: Is a fallback required when there's a default for `migration_job_resource_requirements`? Also, there's no definition at all for this job so it could be counted as "new feature".

roles/installer/templates/jobs/migration.yaml.j2
```diff
@@ -8,6 +8,10 @@ metadata:
     {{ lookup("template", "../common/templates/labels/version.yaml.j2") | indent(width=4) | trim }}
 spec:
   template:
+    metadata:
+      labels:
+        {{ lookup("template", "../common/templates/labels/common.yaml.j2")  | indent(width=8) | trim }}
```
sbilque: The labels are included since PR #1878

sinuscosinustan: rebased and changed commit message

rooftopcellist:
@tomsiewert sorry for the late review on this.  This is a well-formed PR, and would work.  But I think it may be better to use `task_resource_requirements` here so that we don't add new parameters to the CRD unnecessarily.  This way if we change the implementation down the line (to initcontainer for migrations or otherwise), we wouldn't have unused parameters in the CRD.

I can't think of a use case where a user would need the k8s migration job to have different resource requirements then the task pod.

rooftopcellist:
Here is an alternate approach:
* https://github.com/ansible/awx-operator/pull/1912

I am going to merge that for now so that there is a way for people with ResourceQuotas to set limits, but will leave this PR open for a few days in case you want to make a case for this approach over the other.

djyasin:
We are going to go ahead and close this PR. Please feel free to open an issue with any concerns. Thank you for your time!

PR #1891, opened 2024-06-04
Enable readiness probe for task pod in CI

TheRealHaoLiu:
##### SUMMARY
Avoid race condition where job launch before task container is ready

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

```

```


AlanCoding (approved):
If it passes checks then :shipit:

chrismeyersfsu (approved):

PR #1906, opened 2024-06-24
Split metrics utility cronjobs for crc and local report

aknochow:
##### SUMMARY

This separates the console-specific metrics utility into its own tasks and templates as it's not always used together.

##### ISSUE TYPE

### DIFF
--- a/roles/installer/templates/cronjobs/metrics-utility-gather.yaml.j2
+++ b/roles/installer/templates/cronjobs/metrics-utility-gather.yaml.j2
@@ -52,9 +52,11 @@ spec:
                 name: {{ metrics_utility_secret }}
 {% endif %}
             volumeMounts:
+{% if _metrics_utility_ship_target == "directory" %}
             - name: {{ ansible_operator_meta.name }}-metrics-utility
               mountPath: /metrics-utility
               readOnly: false
+{% endif %}
             - name: "{{ ansible_operator_meta.name }}-application-credentials"
               mountPath: "/etc/tower/conf.d/credentials.py"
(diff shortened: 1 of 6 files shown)


roles/installer/templates/cronjobs/metrics-utility-gather.yaml.j2
```diff
@@ -52,9 +52,11 @@ spec:
                 name: {{ metrics_utility_secret }}
 {% endif %}
             volumeMounts:
+{% if metrics_utility_ship_target != "crc" %}
```
rooftopcellist: Shouldn't this match the conditional below?

```suggestion
{% if metrics_utility_ship_target == "directory" %}
```

aknochow: No because this will be required when shipping the metrics from the other cronjob that uses this same template to targets other than directory (such as s3 buckets).

aknochow: And it may be needed for crc as well... waiting for confirmation still.

rooftopcellist: But shouldn't the volumeMount and the volume both only be added under the same conditions?

aknochow: I suppose until we add more shipping target options these can both be either == directory or != crc so I'll just do == directory for now until we have more targets to ship to.

rooftopcellist (approved):

PR #1914, opened 2024-07-05
chore: added aggregate labels to ClusterRoles with kustomize

myles-coleman:
##### SUMMARY
We want to add [user-facing](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles) roles to awx ClusterRole resources. This would allow our users to automatically view/manage these resources without an admin needing to craft custom role bindings. 

#### Cert-Manager Example
Another operator that does this is [cert-manager](https://github.com/cert-manager/cert-manager/blob/659f22bf7ed15ab37dc94f3b714818924e83de3c/deploy/charts/cert-manager/templates/rbac.yaml#L401-L420). Here is an [issue](https://github.com/cert-manager/cert-manager/issues/6239) from cert-manager where they handle a similar problem. They only create the aggregate ClusterRoles if the global.rbac.aggregateClusterRoles value is set to true in their [values.yaml](https://github.com/cert-manager/cert-manager/blob/659f22bf7ed15ab37dc94f3b714818924e83de3c/deploy/charts/cert-manager/values.yaml#L31-L35). 

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
This PR will certainly change current state behavior, possibly granting new permissions to people who update as these ClusterRoles previously had to be specifically installed. An alternative to this would be providing a kustomize patch file which can add these roles, so that people can opt-in instead of automatically being granted these permissions.


rooftopcellist:
cc @rcarrillocruz @TheRealHaoLiu What do you think about this?

PR #1917, opened 2024-07-11
Add ability to disable superuser creation on install

TheRealHaoLiu:
##### ISSUE TYPE

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -57,6 +57,10 @@ spec:
               hostname:  # deprecated
                 description: (Deprecated) The hostname of the instance
                 type: string
+              create_superuser:
+                description: If a superuser should be created
+                type: boolean
+                default: true
               admin_email:
                 description: The admin user email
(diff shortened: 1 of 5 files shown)


TheRealHaoLiu:
@AlanCoding is this PR needed?

config/crd/bases/awx.ansible.com_awxs.yaml
```diff
@@ -57,6 +57,10 @@ spec:
               hostname:  # deprecated
                 description: (Deprecated) The hostname of the instance
                 type: string
+              create_superuser:
```
rooftopcellist, line 60: Since the admin_user and admin_password_secret parameters have already been part of the operator for a while, could we be consistent with the naming of this param and use `create_admin_user` as the param name?

AlanCoding:
This is interesting, but what I've found more relevant recently is the ability to create the admin user _without a password_.

In fact, the JWT auth mechanics may require the admin user to exist anyway. But it doesn't need a usable password set. I would rather it not be, in many cases.

PR #1947, opened 2024-08-28
add postgres securityContext at pod level

dstraub:
##### SUMMARY
The description  for `postgres_security_context_settings`   '...will be set under the pod-level securityContext field' is wrong, this settings are injected in container-level securityContext.
For setting on pod-level the field `postgres_pod_security_context_settings` is added

##### ISSUE TYPE
 - Bug

##### ADDITIONAL INFORMATION
For some CSI-Drivers (e.g. Longhorn) it's required to modify the securityContext on pod-level.
(see  also https://longhorn.io/docs/1.7.0/nodes-and-volumes/volumes/pvc-ownership-and-permission).

The existing field `postgres_security_context_settings` is for container-level.
Maybe this field should renamed to postgres_container_security_context_settings.

```
    ...
        volumeMounts:
        - mountPath: /var/lib/pgsql/data
          name: postgres-15
          subPath: data
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext:
        fsGroup: 26
        fsGroupChangePolicy: Always
      terminationGracePeriodSeconds: 30
    ...

```


Foxlik:
This is much better approach than #1952 as there are different fields allowed in each. At last in our 1.26 clusters. And I think we need the same settings for every pods to actually make AWX work with `pod-security.kubernetes.io/enforce=restricted`.

PR #1982, opened 2024-10-24
adding redirect page

aknochow:
##### ISSUE TYPE


roles/installer/templates/deployments/task.yaml.j2
```diff
@@ -516,6 +521,14 @@ spec:
       terminationGracePeriodSeconds: {{ termination_grace_period_seconds }}
 {% endif %}
       volumes:
+{% if public_base_url is defined %}
+        - name: redirect-page
```
rooftopcellist: This should be configmapped into the web deployment, where nginx is running.

dsavineau (approved):

rooftopcellist (approved):

PR #1990, opened 2024-11-05
Point at awx devel collection

chrismeyersfsu:
##### SUMMARY
awx.awx collection on galaxy is ooooold at this point. Releases are paused, so point at awx collection in devel to get that new bleeding edge hotness.

AAP-34010

##### ISSUE TYPE

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/molecule/requirements.yml
+++ b/molecule/requirements.yml
@@ -4,4 +4,6 @@ collections:
   - name: kubernetes.core
   - name: operator_sdk.util
   - name: community.docker
-  - name: awx.awx
+  - name: https://github.com/ansible/awx.git#/awx_collection/
+    type: git
+    version: devel


molecule/requirements.yml
```diff
@@ -4,4 +4,6 @@ collections:
   - name: kubernetes.core
   - name: operator_sdk.util
   - name: community.docker
-  - name: awx.awx
+  - name: https://github.com/ansible/awx.git#/awx_collection/
```
pb82, line 7: is this URL correct? A local git clone with it fails.

pb82 (approved):

molecule/requirements.yml
chrismeyersfsu: Yep, ansible-galaxy collection install -r requirements.yml seems to be OK with it

Also, pulled it from their docs https://docs.ansible.com/ansible/latest/collections_guide/collections_installing.html#install-multiple-collections-with-a-requirements-file

```
collections:
  - name: https://github.com/organization/repo_name.git
    type: git
    version: devel
```

Issue #2013, opened 2025-02-07
Request to Parameterize the client_max_body_size Directive in the Nginx Configuration for the Web Pod


shellclear:
### Please confirm the following

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [x] I understand that AWX Operator is open source software provided for free and that I might not receive a timely response.

### Feature Summary

Currently, the client_max_body_size directive in the Nginx configuration, defined within the ConfigMap mounted in the web pod at /etc/nginx/nginx.conf, is hardcoded to 5M. This limitation can cause issues when uploading files larger than this size through the web server.

I propose making this directive configurable so that its value can be adjusted via a key in the AWX CR.

shellclear:
https://github.com/ansible/awx-operator/pull/1287

shellclear:
https://github.com/ansible/awx-operator/issues/1286

shellclear:
Merged against devel branch

PR #2014, opened 2025-02-07
Parameterization of the client_max_body_size directive in Nginx

shellclear:
##### SUMMARY
This commit aims to parameterize an Nginx configuration that is currently hardcoded to 5M, causing issues with file uploads.

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
With this change, it will be possible to set the value of the client_max_body_size directive during the creation of the AWX CR or modify an existing AWX instance running in the cluster.

The default value for the directive remains 5M.

```yaml
kind: AWX
metadata:
  name: example-awx1
  namespace: awx
spec:
  nginx_client_max_body_size: 10
```


shellclear:
https://github.com/ansible/awx-operator/issues/2013

fosterseth:
@rooftopcellist I believe this is configurable through the AAP installer, so maybe makes sense to allow it here too

config/crd/bases/awx.ansible.com_awxs.yaml
```diff
@@ -1736,6 +1736,9 @@ spec:
               nginx_worker_connections:
                 description: Set the number of connections per worker for nginx
                 type: integer
+              nginx_client_max_body_size:
+                description: Sets the maximum allowed size of the client request body in megabytes(defaults to 5M)
```
fosterseth:
```suggestion
                description: Sets the maximum allowed size of the client request body in megabytes (defaults to 5M)
```

config/manifests/bases/awx-operator.clusterserviceversion.yaml
```diff
@@ -584,6 +584,11 @@ spec:
         - urn:alm:descriptor:com.tectonic.ui:advanced
         - urn:alm:descriptor:com.tectonic.ui:number
         - urn:alm:descriptor:com.tectonic.ui:hidden
+      - displayName: NGINX Maximum allowed size of the client request body in megabytes
```
fosterseth:
```suggestion
      - displayName: Set the maximum allowed size of the client request body in megabytes for nginx
```

rooftopcellist:
I agree this should be added, and I fully expect the diff here to work.  But I am torn because this same param is configurable in galaxy-operator, but there it is type string not an int that expects the "M" units on the end to be passed on the string.

I think what you have here in this PR is the better approach, but the inconsistency across ansible operators may be confusing.

We plan to create a new apiVersion for all of these operators to make a consistent parameter interface.  When we do, we should modify the galaxy-operator to use this approach (int and adding on the units in the template).

rooftopcellist (approved):

rooftopcellist:
@fosterseth If you agree with this, feel free to merge next week.

shellclear:
> I agree this should be added, and I fully expect the diff here to work. But I am torn because this same param is configurable in galaxy-operator, but there it is type string not an int that expects the "M" units on the end to be passed on the string.
> 
> I think what you have here in this PR is the better approach, but the inconsistency across ansible operators may be confusing.
> 
> We plan to create a new apiVersion for all of these operators to make a consistent parameter interface. When we do, we should modify the galaxy-operator to use this approach (int and adding on the units in the template).

for me, it needs to be an integer. I'll push a new PR to change the parameter in the galaxy-operator so that we can keep the operator aligned.

fosterseth (approved):

PR #2018, opened 2025-02-19
allow configuration of uwsgi timeout

kdelee:
This is important when you have proxies in front of the service that may have stricter timeouts, as you need at least as strict a timeout at uwsgi to get meaningful traceback and see source of problem.

##### SUMMARY

Allow configuration of uwsgi timeout (aka harakiri)

This is important when you have a proxy in front of the service
that may have stricter timeouts, as you need at least as strict a
timeout at uwsgi to get meaningful traceback and see source of problem.

##### ISSUE TYPE


kdelee:
closing in favor of https://github.com/ansible/awx-operator/pull/2056

PR #2025, opened 2025-02-28
Update migrate_schema to use check_migrations

aldato:
##### SUMMARY
We've seen the following errors in the operator logs (trimmed most of the output log since it's not required):

```
[installer : Create kubernetes job to perform the migration] **************\r\ntask path:

---

ResourceClaims:[]core.PodResourceClaim(nil)}}: field is immutable\\\",\\\"field\\\":\\\"spec.template\\\"}]},\\\"code\\\":422}\\\\n'\", \"reason\": \"Unprocessable Entity\", \"status\": 422}\n\r\n

\r\nlocalhost                  : ok=73   changed=0    unreachable=0    failed=1    skipped=76   rescued=0    ignored=0
```

If we check the output of the main conditional we can see that we've two unexpected results:

```
$ awx-manage showmigrations | grep -v '[X]' | grep '[ ]' | wc -l
2

$ awx-manage showmigrations | grep -v '[X]' | grep '[ ]'
 (no migrations)
 (no migrations)
```
They belong to:

```
dab_jwt_consumer
 (no migrations)
dab_rest_filters
 (no migrations)
```

The block interprets there are pending migrations when that's not the case and tries to edit the k8s job that already ran. If there are changes in the AutomationController CR it tries to change the k8s job spec.template, which is immutable.

##### ISSUE TYPE


roles/installer/tasks/migrate_schema.yml
```diff
@@ -6,7 +6,7 @@
     pod: "{{ awx_web_pod_name }}"
     container: "{{ ansible_operator_meta.name }}-web"
     command: >-
-      bash -c "awx-manage showmigrations | grep -v '[X]' | grep '[ ]' | wc -l"
+      bash -c "awx-manage check_migrations && echo $?"
```
chrismeyersfsu: The `&& echo $?` is good for bash command line debugging outside of ansible. But I think you want to remove it in this PR.

The `database_check.rc != 0` below is correct.

aldato: Yup, forgot to remove that and updated it just now. Thanks :)

ppmathis:
Thank you for the PR, @aldato! I stumbled across the same issue with AWX myself and decided to give this a test, but noticed that while the issue with the immutable Kubernetes job no longer happens, migrations are actually never being executed at all. Spawning an AWX operator with the modified `migrate_schema.yml` and then launching a new instance also reveals this, as it's forever stuck in the `init-database` task and migrations simply never run.

This led me towards checking out how `check_migrations` is defined, and well its naming is quite misleading; as can be seen [here](https://github.com/ansible/awx/blob/99be91e939a18305372729a718c983622b87212a/awx/main/management/commands/check_migrations.py#L4) in its command file, it's just an alias for `makemigrations` from Django.

So unfortunately the usage of `check_migrations` is wrong, as it will no longer check if there are any migrations that have not been applied to the system, but instead check if any Django models have changed and would require creating new migration files, which usually never happens in a release scenario, therefor never running the migrations.

In the end I reverted all changes to the `migrate_schema.yml` file and instead just used an updated command that excludes the specific extraneous lines caused by two Django apps with no migrations: `bash -c "awx-manage showmigrations | grep -v '(no migrations)' | grep -v '[X]' | grep '[ ]' | wc -l"`

Doing so worked well in my tests, meaning it both prevented the original issue but still applied migrations as needed. Maybe this PR could be updated to match this behavior?

aldato:
Thanks a lot for the suggestion and investigation @ppmathis! Just updated the PR with the changes that you proposed.

rooftopcellist (approved):

PR #2026, opened 2025-03-03
Add deployment type shortname for legacy API url

jamesmarshall24:
##### SUMMARY

Add deployment_type_shortname for legacy api url

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

n/a

### DIFF
--- a/roles/installer/templates/configmaps/redirect-page.configmap.html.j2
+++ b/roles/installer/templates/configmaps/redirect-page.configmap.html.j2
@@ -66,7 +66,7 @@ data:
         <p class="doc-note">
             The API endpoints for this platform service will temporarily remain available at the URL for this service.
             Please use the Ansible Automation Platform API endpoints corresponding to this component in the future.
-            These can be found at <a href="{{ public_base_url }}/api/{{ deployment_type }}" target="_blank">{{ public_base_url }}/api/{{ deployment_type }}</a>.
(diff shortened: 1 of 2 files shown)


rooftopcellist (approved):

roles/installer/templates/configmaps/redirect-page.configmap.html.j2
```diff
@@ -66,7 +66,7 @@ data:
         <p class="doc-note">
             The API endpoints for this platform service will temporarily remain available at the URL for this service.
             Please use the Ansible Automation Platform API endpoints corresponding to this component in the future.
-            These can be found at <a href="{{ public_base_url }}/api/{{ deployment_type }}" target="_blank">{{ public_base_url }}/api/{{ deployment_type }}</a>.
+            These can be found at <a href="{{ public_base_url }}/api/{{ deployment_type_shortname }}" target="_blank">{{ public_base_url }}/api/{{ deployment_type_shortname }}</a>.
```
rooftopcellist: @jamesmarshall24 I just tested this out and without the trailing slash, it's a 404.  That's technically an nginx conf bug, but can you add the slash on the end, which should make it work until we can fix the nginx.conf.

```suggestion
            These can be found at <a href="{{ public_base_url }}/api/{{ deployment_type_shortname }}" target="_blank">{{ public_base_url }}/api/{{ deployment_type_shortname }}/</a>.
```

rooftopcellist (approved):

PR #2043, opened 2025-05-09
fix: explicitly import ldap on config file

rochacbruno:

```
File "/etc/tower/conf.d/ldap.py", line 2, in <module> ldap.OPT_X_TLS_REQUIRE_CERT: True,
^^^^
NameError: name 'ldap' is not defined
```

### DIFF
--- a/roles/installer/templates/settings/ldap.py.j2
+++ b/roles/installer/templates/settings/ldap.py.j2
@@ -1,11 +1,13 @@
+{% if ldap_cacert_ca_crt %}
 import ldap
 
 AUTH_LDAP_GLOBAL_OPTIONS = {
-{% if ldap_cacert_ca_crt %}
     ldap.OPT_X_TLS_REQUIRE_CERT: True,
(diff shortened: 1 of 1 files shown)


rooftopcellist (approved):

roles/installer/templates/settings/ldap.py.j2
```diff
@@ -1,3 +1,5 @@
+import ldap
+
```
AlanCoding, line 3: I do think it would be nice to nest this in the if. This requires some refactoring.

```jinja2
{% if ldap_cacert_ca_crt %}
import ldap

AUTH_LDAP_GLOBAL_OPTIONS = {
    ldap.OPT_X_TLS_REQUIRE_CERT: True,
    ldap.OPT_X_TLS_CACERTFILE: "/etc/openldap/certs/ldap-ca.crt"
}
{% else %}
AUTH_LDAP_GLOBAL_OPTIONS = {}
{% endif %}
```

dsavineau (approved):

PR #2045, opened 2025-05-12
Add route_annotations feature to mesh ingress CRD

sharvarikhedkar:
##### SUMMARY
This PR adds ability to set custom annonations on AAP mesh ingress routes through AWX CR.
Jira: https://issues.redhat.com/browse/AAP-44139

##### ISSUE TYPE

 - New or Enhanced Feature

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxmeshingresses.yaml
+++ b/config/crd/bases/awx.ansible.com_awxmeshingresses.yaml
@@ -71,9 +71,7 @@ spec:
                 type: string
               route_annotations:
                 description: Annotations to add to the OpenShift Route
-                type: object
-                additionalProperties:
-                  type: string
+                type: string
               ingress_class_name:
                 description: The name of ingress class to use instead of the cluster default.
                 type: string
--- a/config/crd/bases/awx.ansible.com_awxmeshingresses.yaml
+++ b/config/crd/bases/awx.ansible.com_awxmeshingresses.yaml
@@ -69,6 +69,11 @@ spec:
               ingress_annotations:
                 description: Annotations to add to the Ingress Controller
                 type: string
+              route_annotations:
+                description: Annotations to add to the OpenShift Route
+                type: object
+                additionalProperties:
+                  type: string
               ingress_class_name:
                 description: The name of ingress class to use instead of the cluster default.
                 type: string
--- a/config/manifests/bases/awx-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/awx-operator.clusterserviceversion.yaml
@@ -50,6 +50,10 @@ spec:
         path: ingress_annotations
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:text
+      - displayName: Route Annotations
+        path: route_annotations
+        x-descriptors:
+          - urn:alm:descriptor:com.tectonic.ui:text
       - displayName: Ingress Class Name
         path: ingress_class_name
         x-descriptors:
--- a/config/manifests/bases/awx-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/awx-operator.clusterserviceversion.yaml
@@ -53,7 +53,9 @@ spec:
       - displayName: Route Annotations
         path: route_annotations
         x-descriptors:
-          - urn:alm:descriptor:com.tectonic.ui:text
+          - 'urn:alm:descriptor:com.tectonic.ui:advanced'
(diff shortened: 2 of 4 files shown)


config/crd/bases/awx.ansible.com_awxmeshingresses.yaml
```diff
@@ -69,6 +69,11 @@ spec:
               ingress_annotations:
                 description: Annotations to add to the Ingress Controller
                 type: string
+              route_annotations:
+                description: Annotations to add to the OpenShift Route
+                type: object
+                additionalProperties:
+                  type: string
```
rooftopcellist: @aknochow what do you think about this?  I like this notation better for adding annotations.  But it does not match the current format used by ingress_annotations.  

If we just had type: string like we currently have for ingress_annotations, it would look like this:
image

config/crd/bases/awx.ansible.com_awxmeshingresses.yaml
```diff
@@ -69,6 +69,11 @@ spec:
               ingress_annotations:
                 description: Annotations to add to the Ingress Controller
                 type: string
```
rooftopcellist, line 71: @sharvarikhedkar we need a similar entry in the CSV file

sharvarikhedkar: I'm not sure I follow, do you mean https://github.com/ansible/awx-operator/pull/2045/files#diff-2a4968792da0598da409c1e5bbc604542cdf7b2b52bbbf4179dd018fcf9baaf4R53-R57?

config/crd/bases/awx.ansible.com_awxmeshingresses.yaml
rooftopcellist: Alternatively, it would look like this if you used an object with key/value pairs:
image

```
                ingress_annotations:
                  description: Annotations to add to the Ingress Controller
                  type: object
                  properties:
                    key:
                      type: string
                    value:
                      type: string
```

Which, when configured, would look like this in yaml:

```
  ingress_annotations:
    key: kubernetes.haproxy
    value: 'true'
```

rooftopcellist: Where-as I think there is a cleaner way to do this long-term, all of the other operators currently use `type: string` for annotations and labels.  I think if we change this, we should change it for all of them at once and cut a new apiVersion.

So the CRD would change to this:
```suggestion
              route_annotations:
                description: Annotations to add to the OpenShift Route
                type: string
```

Here is an example of how it would be configured with `type: string`

```
---
spec:
  ...
  ingress_annotations: |
    environment: testing
    region: east-2a
```

config/manifests/bases/awx-operator.clusterserviceversion.yaml
```diff
@@ -50,6 +50,10 @@ spec:
         path: ingress_annotations
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:text
+      - displayName: Route Annotations
+        path: route_annotations
+        x-descriptors:
+          - urn:alm:descriptor:com.tectonic.ui:text
```
rooftopcellist: @sharvarikhedkar I think we should change this to only show if `ingress_type: Route` is selected.

```suggestion
      - displayName: Route Annotations
        path: route_annotations
        x-descriptors:
          - 'urn:alm:descriptor:com.tectonic.ui:advanced'
          - 'urn:alm:descriptor:com.tectonic.ui:text'
          - 'urn:alm:descriptor:com.tectonic.ui:fieldDependency:ingress_type:Route'
```

config/crd/bases/awx.ansible.com_awxmeshingresses.yaml
sharvarikhedkar: changed back to string.

config/manifests/bases/awx-operator.clusterserviceversion.yaml
sharvarikhedkar: updated.

dsavineau (approved):

config/crd/bases/awx.ansible.com_awxmeshingresses.yaml
rooftopcellist: I missed that on the first pass.  Looks like you have it already, thanks.  Resolved.

rooftopcellist (approved):

PR #2046, opened 2025-05-13
Make playbook compatible with new Ansible-core type system

AlanCoding:
##### SUMMARY
In recent release, ansible-core got pickier about types. CLI arguments of `-e` are always string, which kind of causes this problem, so ports from CLI need to be converted to ints.

##### ISSUE TYPE

### DIFF
--- a/ansible/instantiate-awx-deployment.yml
+++ b/ansible/instantiate-awx-deployment.yml
@@ -26,7 +26,7 @@
             image_version: "{{ image_version | default(omit) }}"
             development_mode: "{{ development_mode | default(omit) | bool }}"
             image_pull_policy: "{{ image_pull_policy | default(omit) }}"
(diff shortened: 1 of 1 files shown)


ansible/instantiate-awx-deployment.yml
```diff
@@ -26,7 +26,7 @@
             image_version: "{{ image_version | default(omit) }}"
             development_mode: "{{ development_mode | default(omit) | bool }}"
             image_pull_policy: "{{ image_pull_policy | default(omit) }}"
-            nodeport_port: "{{ nodeport_port | default(omit) }}"
+            nodeport_port: "{{ nodeport_port | default(omit) | int }}"
```
bcoca, line 29:
```suggestion
            nodeport_port: "{{ nodeport_port | ternary(nodeport_port | int, omit) }}"
```

PR #2056, opened 2025-07-01
set api timeout to match proxy timeout

rebeccahhh:
##### SUMMARY

Timeout before the openshift route times out
not timing out before undercuts usefulness of our log-traceback-middleware in django-ansible-base that logs a traceback from requests that get timed out -- because uwsgi or gunicorn has to send the timeout signal to the worker handling the request. Also leads to issues where requests that envoy has already timed out are filling up queues of the workers of the components.

Also, configure nginx to return a 503 if WSGI server doesn't respond.

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

This makes no changes to the CR, it just aligns existing timeouts for nginx and uwsgi to the known default timeout of 30 seconds, which is not currently configurable because we don't allow route annotations.

### DIFF
--- a/roles/installer/templates/configmaps/config.yaml.j2
+++ b/roles/installer/templates/configmaps/config.yaml.j2
@@ -187,7 +187,7 @@ data:
                 allow 127.0.0.1;
                 deny all;
             }
-
+ 
             location {{ (ingress_path + '/static').replace('//', '/') }} {
                 alias /var/lib/awx/public/static/;
             }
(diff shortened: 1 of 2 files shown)


rebeccahhh:
image
image
screenshots post change, everything functions still :)

roles/installer/templates/configmaps/config.yaml.j2
```diff
@@ -187,7 +187,7 @@ data:
                 allow 127.0.0.1;
                 deny all;
             }
-
+
```
kdelee: see i fyou can get rid of this whitespace diff

kdelee (commented):
just one nit on whitespace diff

rooftopcellist (approved):
Good to go after the whitespace is removed.

rebeccahhh:
Whitespace removed ✅

PR #2064, opened 2025-07-21
Run import_auth_config_to_gateway when public_url is defined

rcarrillocruz:
##### SUMMARY

Run import_auth_config in case the public_url is defined

##### ISSUE TYPE

 - New or Enhanced Feature

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/roles/installer/tasks/initialize_django.yml
+++ b/roles/installer/tasks/initialize_django.yml
@@ -56,6 +56,17 @@
       bash -c "awx-manage unregister_queue --queuename=tower"
   when: "'[tower capacity=' in legacy_queue.stdout"
 
+- name: Import auth config when public_url is defined and gating version is defined
+  kubernetes.core.k8s_exec:
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    pod: "{{ awx_web_pod_name }}"
+    container: "{{ ansible_operator_meta.name }}-web"
+    command: >-
+      bash -c "awx-manage import_auth_config_to_gateway"
+  when:
+    - gating_version | length
+    - public_base_url is defined
+
 - name: Check for specified default execution environment pull credentials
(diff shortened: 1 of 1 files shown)


aknochow (approved):

roles/installer/tasks/initialize_django.yml
```diff
@@ -56,6 +56,18 @@
...
+- name: Import auth config when public_url is defined and previous version is less than 4.6.0
+  kubernetes.core.k8s_exec:
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    pod: "{{ awx_web_pod_name }}"
+    container: "{{ ansible_operator_meta.name }}-web"
+    command: >-
+      bash -c "awx-manage import_auth_config_to_gateway"
+  when:
+    - previous_version is defined
+    - previous_version is version('4.6.0', '<')
```
dsavineau: Gating version is only defined for downstream so you shouldn't have to specify version here in upstream

```suggestion
    - gating_version | length
```

That's how we're already doing for specific upgrade workflow

https://github.com/ansible/awx-operator/blob/devel/roles/installer/tasks/main.yml#L30-L32

roles/installer/tasks/initialize_django.yml
```diff
@@ -56,6 +56,18 @@
       bash -c "awx-manage unregister_queue --queuename=tower"
   when: "'[tower capacity=' in legacy_queue.stdout"
 
+- name: Import auth config when public_url is defined and previous version is less than 4.6.0
```
dsavineau:
```suggestion
- name: Import auth config when public_url is defined and gating version is defined
```

dsavineau (changes requested):

dsavineau (approved):

PR #2076, opened 2025-08-27
Guard against missing version status on existing CR

rooftopcellist:
##### SUMMARY
Fixes bad conditional that was causing the reconciliation loop to fail if the version status is not set on the AWX CR, which can happen in some cases.

##### ISSUE TYPE


msilmser:
lgtm

PR #2081, opened 2025-09-15
Fix ca certs

aknochow:
##### SUMMARY
Fixing bundle ca cert in controller

##### ISSUE TYPE

### DIFF
--- a/roles/installer/tasks/load_bundle_cacert_secret.yml
+++ b/roles/installer/tasks/load_bundle_cacert_secret.yml
@@ -11,3 +11,4 @@
   set_fact:
     bundle_ca_crt: '{{ bundle_cacert["resources"][0]["data"]["bundle-ca.crt"] | b64decode }}'
   no_log: "{{ no_log }}"
+  when: '"bundle-ca.crt" in bundle_cacert["resources"][0]["data"]'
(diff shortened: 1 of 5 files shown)


roles/installer/tasks/load_bundle_cacert_secret.yml
```diff
@@ -11,4 +11,3 @@
   set_fact:
     bundle_ca_crt: '{{ bundle_cacert["resources"][0]["data"]["bundle-ca.crt"] | b64decode }}'
   no_log: "{{ no_log }}"
-  when: '"bundle-ca.crt" in bundle_cacert["resources"][0]["data"]'
```
dsavineau: I don't think we should remove that condition

Deployments rely on the `bundle-ca.crt` key within the secret data so if you remove that one and user don't define it then deployment creation will fail

roles/installer/templates/jobs/migration.yaml.j2
```diff
@@ -28,7 +28,7 @@ spec:
           volumeMounts:
             - name: "ca-trust-extracted"
               mountPath: "/etc/pki/ca-trust/extracted"
-            - name: "{{ ansible_operator_meta.name }}-bundle-cacert"
+            - name: "{{ bundle_cacert_secret }}"
```
rooftopcellist: Renaming this mount makes it less consistent with the other operators, eda-server-operator for example.  Do we want to rename this everywhere?

rooftopcellist (approved):

PR #2103, opened 2026-02-27
Fix AWXRestore multiple bugs

### LINKED ISSUE #2093: [BUG] AWXRestore fails: force_drop_db ignored, CREATE DATABASE lacks privileges, and pg_restore fails on partitioned indexes

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/awx-operator/issues) for duplicates.
- [x] I understand that the AWX Operator is open source software provided for free and that I might not receive a timely response.

### Bug Summary

> **Note to maintainers**: This issue describes three related bugs in the AWXRestore process. Feel free to split into separate issues if preferred, but they are interdependent and should be addressed together for a complete fix.

AWXRestore has three critical bugs that prevent successful database restoration:

1. **Bug-1**: `force_drop_db` parameter is always ignored due to Ansible vars precedence
2. **Bug-2**: `CREATE DATABASE` fails because managed PostgreSQL user lacks `CREATEDB` privilege
3. **Bug-3**: `pg_restore --clean` fails on partitioned table indexes due to dependency ordering

**Combined Impact**: AWXRestore is effectively non-functional for any AWX instance that has been running long enough to create partitioned event tables.

### AWX Operator version

2.19.1

### AWX version

24.6.1

### Kubernetes platform

kubernetes

1.33

### Modifications

no

### Steps to reproduce

1. Deploy AWX with managed PostgreSQL (default)
2. Let AWX run to create partitioned event tables (or run some jobs)
3. Create backup with AWXBackup CR
4. Apply AWXRestore CR:
   ```yaml
   kind: AWXRestore
   metadata:
     name: restore-test
     namespace: awx
   spec:
     deployment_name: awx-demo
     backup_name: awx-backup
     force_drop_db: true  # Ignored due to Bug-1
   ```
5. Restore fails with partition index error (Bug-3)

### Expected results

Restore should complete successfully when:
- `force_drop_db: false`: pg_restore cleans and restores to existing database
- `force_drop_db: true`: Database is dropped, recreated, and restored

### Actual results

- `force_drop_db` is always `false` (Bug-1)
- Even if fixed, CREATE DATABASE fails (Bug-2)
- Even if both fixed, pg_restore fails on partition indexes (Bug-3)

### Location
`roles/restore/vars/main.yml:19`

```yaml
force_drop_db: false
```

### Problem
In Ansible, variables defined in `vars/` have **higher precedence** than values from CR spec and `defaults/main.yml`.

This means that regardless of what value a user sets in their AWXRestore CR:

```yaml
kind: AWXRestore
spec:
  force_drop_db: true   # ← ALWAYS IGNORED!
```

The value will **always** be `false`.

### Evidence
When applying AWXRestore with `force_drop_db: true`, operator logs show **no DROP DATABASE attempt**. The restore proceeds directly to `pg_restore --clean` and fails on partition indexes.

### Suggested Fix
Remove `force_drop_db: false` from `vars/main.yml` and read explicitly from spec:

```yaml
# Add to roles/restore/tasks/postgres.yml after line 5
- name: Set force_drop_db from spec
  set_fact:
    force_drop_db: "{{ spec['force_drop_db'] | default(false) | bool }}"
```

---

### Location
`roles/restore/tasks/postgres.yml:97-105`

### Problem
The `CREATE DATABASE` command uses the AWX database user which does **not** have the `CREATEDB` privilege in managed PostgreSQL deployments.

```yaml
pg_create_db: >-
  echo 'CREATE DATABASE {{ awx_postgres_database }} ...' |
  PGPASSWORD='{{ awx_postgres_pass }}' psql -U {{ awx_postgres_user }} ...
```

### Manual Verification
```bash
# DROP works (user owns the database)
echo 'DROP DATABASE awx WITH (FORCE);' | psql -U awx -d postgres
# ✅ Success

# CREATE fails (user lacks CREATEDB privilege)
echo 'CREATE DATABASE awx WITH OWNER = awx;' | psql -U awx -d postgres
# ❌ ERROR: permission denied to create database
```

### Suggested Fix

**Option A**: Grant CREATEDB during PostgreSQL setup:
```sql
ALTER USER {{ awx_postgres_user }} CREATEDB;
```

**Option B**: Use postgres superuser for DROP/CREATE operations

**Option C**: Use schema-level cleanup instead:
```sql
DROP SCHEMA public CASCADE;
CREATE SCHEMA public;
GRANT ALL ON SCHEMA public TO {{ awx_postgres_user }};
```

---

### Location
`roles/restore/tasks/postgres.yml:75-83`

```yaml
pg_restore: >-
  pg_restore --clean --if-exists
  -U {{ awx_postgres_user }}
  -h {{ resolvable_db_host }}
  -d {{ awx_postgres_database }}
  -p {{ awx_postgres_port }}
```

### Problem
AWX uses PostgreSQL table partitioning for event tables (e.g., `main_jobevent`, `main_systemjobevent`). When creating an index on a partitioned parent table, PostgreSQL automatically creates "attached" indexes on all partitions.

During `pg_restore --clean`:
1. pg_restore attempts to DROP partition indexes in the order they appear in the dump
2. PostgreSQL blocks the DROP because partition indexes depend on the parent index
3. The restore fails with:

```
pg_restore: error: could not execute query: ERROR: cannot drop index
public.main_systemjobevent_20260112__system_job_id_job_created_uui_idx
because index public.main_system_system__e39825_idx requires it
HINT: You can drop index public.main_system_system__e39825_idx instead.
```

### Root Cause
The `--clean` flag in `pg_restore` generates DROP statements based on the dump content, but **does not respect the dependency order** required for partitioned table indexes.

### Suggested Fixes

**Option A** (Recommended): When `force_drop_db: true`, don't use `--clean` (database is already empty):
```yaml
  set_fact:
    pg_restore: >-
      pg_restore {{ '' if force_drop_db else '--clean --if-exists' }}
      -U {{ awx_postgres_user }}
      -h {{ resolvable_db_host }}
      -d {{ awx_postgres_database }}
      -p {{ awx_postgres_port }}
```

**Option B**: Use `--section` flags for controlled restore:
```bash
pg_restore --section=pre-data -d awx backup.dump
pg_restore --section=data -d awx backup.dump
pg_restore --section=post-data -d awx backup.dump
```

**Option C**: Pre-drop all indexes in correct order before pg_restore

---

## Impact Summary

| Bug | Severity | Impact |
|-----|----------|--------|
| 1 | High | `force_drop_db` feature completely broken |
| 2 | High | Cannot create database even if Bug-1 is fixed |
| 3 | High | Cannot restore to existing DB with partitioned tables |

---

## Related Files

| File | Line | Bug |
|------|------|-----|
| `roles/restore/vars/main.yml` | 19 | 1 - hardcoded `force_drop_db: false` |
| `roles/restore/tasks/postgres.yml` | 85-112 | 1 - block never executes |
| `roles/restore/tasks/postgres.yml` | 97-105 | 2 - uses user without CREATEDB |
| `roles/restore/tasks/postgres.yml` | 75-83 | 3 - `--clean` flag causes failure |
| `config/crd/bases/awx.ansible.com_awxrestores.yaml` | 129-132 | CRD defines field correctly |

---

## Workaround

Currently there is **no workaround** within the operator.

Manual workaround requires direct database access:
1. Scale down AWX deployments (`kubectl scale deployment awx-task awx-web --replicas=0`)
2. Connect to PostgreSQL as superuser
3. Manually drop and recreate database
4. Run `pg_restore` without `--clean` flag
5. Run AWX migrations (`awx-manage migrate`)
6. Scale up AWX deployments

### Operator Logs

--------------------------- Ansible Task StdOut -------------------------------

 TASK [Restore database dump to the new postgresql container] ******************************** 
fatal: [localhost]: FAILED! => {"changed": true, "rc": 1, "return_code": 1, "stderr": "pg_restore: error: could not execute query: ERROR:  cannot drop index public.main_systemjobevent_20260112__system_job_id_job_created_uui_idx because index public.main_system_system__e39825_idx requires it\nHINT:  You can drop index public.main_system_system__e39825_idx instead.\nCommand was: DROP INDEX IF EXISTS public.main_systemjobevent_20260112__system_job_id_job_created_uui_idx;\npg_restore: error: could not execute query: ERROR:  cannot drop index public.main_systemjobevent_20260112__system_job_id_job_created_uu_idx2 because index public.main_system_system__e39825_idx requires it\nHINT:  You can drop index public.main_system_system__e39825_idx instead.\nCommand was: DROP INDEX IF EXISTS 
...

lucas-benedito:
##### SUMMARY
- Move force_drop_db from vars/main.yml to defaults/main.yml so CR spec values are not overridden by Ansible variable precedence
- Grant CREATEDB priv to database user before DROP/CREATE and revoke it after restore, following the containerized-installer pattern
- Omit --clean --if-exists from pg_restore when force_drop_db is true since the database is freshly created and empty, avoiding partition index dependency errors

##### ISSUE TYPE

##### ADDITIONAL INFORMATION
Fixes: #2093

### DIFF
--- a/roles/restore/tasks/postgres.yml
+++ b/roles/restore/tasks/postgres.yml
@@ -83,12 +83,23 @@
 - name: Set pg_restore command
   set_fact:
     pg_restore: >-
-      pg_restore --clean --if-exists --no-owner --no-acl
+      pg_restore {{ force_drop_db | bool | ternary('', '--clean --if-exists') }} --no-owner --no-acl
       -U {{ awx_postgres_user }}
       -h {{ resolvable_db_host }}
       -d {{ awx_postgres_database }}
       -p {{ awx_postgres_port }}
   no_log: "{{ no_log }}"
+
+- name: Grant CREATEDB privilege to database user for force_drop_db
+  kubernetes.core.k8s_exec:
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    pod: "{{ postgres_pod_name }}"
+    container: postgres
+    command: >-
+      psql -c "ALTER USER {{ awx_postgres_user }} CREATEDB;"
+  when:
+    - force_drop_db | bool
+    - awx_postgres_type == 'managed'
 
 - name: Force drop and create database if force_drop_db is true
   block:
@@ -155,3 +166,14 @@
       "
   register: data_migration
   no_log: "{{ no_log }}"
+
+- name: Revoke CREATEDB privilege from database user
+  kubernetes.core.k8s_exec:
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    pod: "{{ postgres_pod_name }}"
+    container: postgres
+    command: >-
+      psql -c "ALTER USER {{ awx_postgres_user }} NOCREATEDB;"
+  when:
+    - force_drop_db | bool
+    - awx_postgres_type == 'managed'
--- a/roles/restore/defaults/main.yml
+++ b/roles/restore/defaults/main.yml
@@ -40,5 +40,8 @@ additional_labels: []
 # Maintain some of the recommended `app.kubernetes.io/*` labels on the resource (self)
 set_self_labels: true
 
+# If set to true, the restore process will drop and recreate the database schema before restoring
+force_drop_db: false
+
 spec_overrides: {}
 ...
--- a/roles/restore/vars/main.yml
+++ b/roles/restore/vars/main.yml
@@ -14,7 +14,4 @@ broadcast_websocket_secret: '{{ deployment_name }}-broadcast-websocket'
 postgres_configuration_secret: '{{ deployment_name }}-postgres-configuration'
 supported_pg_version: 15
 image_pull_policy: IfNotPresent
-
-# If set to true, the restore process will delete the existing database and create a new one
-force_drop_db: false
 pg_drop_create: ''


jamesmarshall24 (approved):

roles/restore/tasks/postgres.yml
```diff
@@ -83,13 +83,23 @@
 - name: Set pg_restore command
   set_fact:
     pg_restore: >-
-      pg_restore --clean --if-exists --no-owner --no-acl
+      pg_restore {{ '' if force_drop_db else '--clean --if-exists' }} --no-owner --no-acl
```
dsavineau:
```suggestion
      pg_restore {{ force_drop_db | bool | ternary('', '--clean --if-exists') }} --no-owner --no-acl
```

roles/restore/tasks/postgres.yml
```diff
@@ -83,13 +83,23 @@
...
   no_log: "{{ no_log }}"
 
+- name: Grant CREATEDB privilege to database user for force_drop_db
+  kubernetes.core.k8s_exec:
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    pod: "{{ postgres_pod_name }}"
+    container: postgres
+    command: >-
+      psql -c "ALTER USER {{ awx_postgres_user }} CREATEDB;"
+  no_log: "{{ no_log }}"
```
dsavineau:
```suggestion
```
no_log isn't needed here

roles/restore/tasks/postgres.yml
```diff
@@ -83,13 +83,23 @@
...
+  when: force_drop_db and awx_postgres_type == 'managed'
```
dsavineau:
```suggestion
  when:
    - force_drop_db | bool
    - awx_postgres_type == 'managed'
```

rooftopcellist (approved):

roles/restore/tasks/postgres.yml
```diff
@@ -155,3 +165,13 @@
...
+
+- name: Revoke CREATEDB privilege from database user
+  kubernetes.core.k8s_exec:
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    pod: "{{ postgres_pod_name }}"
+    container: postgres
+    command: >-
+      psql -c "ALTER USER {{ awx_postgres_user }} NOCREATEDB;"
+  no_log: "{{ no_log }}"
+  when: force_drop_db and awx_postgres_type == 'managed'
```
dsavineau: ditto

lucas-benedito:
@dsavineau Updated. Thanks for the suggestions

dsavineau (approved):

PR #2109, opened 2026-03-18
Add create_superuser option to skip admin user creation

TheRealHaoLiu:
## Summary

When deploying AWX as part of AAP where all authentication goes through the AAP Gateway, the component-level admin user is not needed. This PR adds a boolean `create_superuser` field (default: `true`) that allows operators to opt out of creating the admin superuser during installation.

## Changes

- `roles/installer/defaults/main.yml` — add `create_superuser: true` default
- `roles/installer/tasks/install.yml` — skip `admin_password_configuration` when `create_superuser=false`
- `roles/installer/tasks/initialize_django.yml` — wrap `createsuperuser` + `update_password` in a block gated by `create_superuser | bool`

## Behavior

When `create_superuser: false`:
- The `<name>-admin-password` K8s Secret is not created
- `awx-manage createsuperuser` is not run
- `awx-manage update_password` is not run

When `create_superuser: true` (default): no change to existing behavior.

## Context

In AAP 2.7, direct API access to platform components (Controller, Hub, EDA) is being removed. All authentication is centralized through the AAP Gateway. The component-level admin user was previously needed for installer post-install tasks (license upload, postinstall config), but those have since been updated to use gateway credentials. There is no remaining post-creation caller for the controller admin user in the installer or operator.

Relates-to: AAP-68980

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxs.yaml
+++ b/config/crd/bases/awx.ansible.com_awxs.yaml
@@ -1691,6 +1691,10 @@ spec:
                 description: Whether or not to preload data upon instance creation
                 default: true
                 type: boolean
+              create_superuser:
+                description: Whether to create the admin superuser during installation. Set to false when deploying as part of AAP where the gateway admin user is used instead.
+                default: true
+                type: boolean
               task_args:
                 type: array
                 items:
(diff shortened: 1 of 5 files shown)


AlanCoding (approved):
Seems to be failing on a line length violation

PR #2114, opened 2026-05-06
refactor: remove proxy env var CRD fields and CSV specDescriptors

jamesmarshall24:
##### SUMMARY

Removes `http_proxy`, `https_proxy`, and `no_proxy` from the CRD spec and CSV
specDescriptors. Proxy configuration is injected into the operator pod environment
and propagated to containers via the existing ConfigMap mechanism.

##### ISSUE TYPE


rooftopcellist (approved):
Reviewed the diff — clean removal of proxy CRD fields and CSV specDescriptors. The lookup('env', ...) defaults in defaults/main.yml correctly inherit from the operator pod environment (set by OLM from the cluster-wide proxy). Consistent with the same change across all AAP operators.

PR #2120, opened 2026-06-02
Allow skipping PostgreSQL database backup/restore

dsavineau:
##### SUMMARY
This allows one to skip the PostgreSQL database dump/restore during via the AWXBackup and AWXRestore CRDs on external database configuration.

##### ISSUE TYPE
 - New or Enhanced Feature

##### ADDITIONAL INFORMATION
A new parameter called `postgres_skip_data` is introduced (defaults to false for backward compatibility) but can be turn on to skip the pg_dump and pg_restore parts.

At restore time, if no database dump file is found in the backup then the `postgres_skip_data` parameter needs to be explicitely set to true.

### DIFF
--- a/config/crd/bases/awx.ansible.com_awxbackups.yaml
+++ b/config/crd/bases/awx.ansible.com_awxbackups.yaml
@@ -101,6 +101,10 @@ spec:
               postgres_image_version:
                 description: PostgreSQL container image version to use
                 type: string
+              postgres_skip_data:
+                description: Skip PostgreSQL database dump.
+                type: boolean
+                default: false
               precreate_partition_hours:
                 description: Number of hours worth of events table partitions to precreate before backup to avoid pg_dump locks.
                 type: integer
@@ -156,3 +160,6 @@ spec:
               backupClaim:
                 description: Backup persistent volume claim
                 type: string
+              backupPostgres:
+                description: Backup PostgreSQL data
+                type: boolean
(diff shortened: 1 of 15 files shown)


config/crd/bases/awx.ansible.com_awxbackups.yaml
```diff
@@ -101,6 +101,10 @@ spec:
               postgres_image_version:
                 description: PostgreSQL container image version to use
                 type: string
+              postgres_skip_data:
+                description: Skip PostgresSQL database dump.
```
rooftopcellist: Typo: `PostgresSQL` → `PostgreSQL` (single capital S). This appears in 4 places across the CRDs and defaults files. It's user-facing text in the API schema.

roles/backup/tasks/update_status.yml
```diff
@@ -10,4 +10,5 @@
     status:
       backupDirectory: "{{ backup_dir }}"
       backupClaim: "{{ backup_pvc }}"
+      backupPostgres: "{{ not postgres_skip_data | bool }}"
```
rooftopcellist: This status value won't reflect reality when `postgres_skip_data: true` on a managed database. The backup still runs (due to the `or awx_postgres_type == 'managed'` condition in `creation.yml`), but this will report `backupPostgres: false`.

Should mirror the actual execution condition:
```yaml
backupPostgres: "{{ not postgres_skip_data | bool or awx_postgres_type == 'managed' }}"
```

rooftopcellist (approved):

Issue #191, opened 2021-04-08
quay.io complains about unpatched security flaws in the awx-operator images


lemmy04:
one of the bigger features of the quay.io registry is the built in vulnerabilty scan which checks container images for unpatched flaws.

As of today, april 8 2021, it complains about 2 high, 25 medium, and 5 low vulnerabilities in the awx-operator image.

What can be done to fix this?

![grafik](https://user-images.githubusercontent.com/664127/113998486-66d9e780-9859-11eb-9584-5fa3d4006b36.png)

shanemcd:
These are present in the `quay.io/operator-framework/ansible-operator` base image. You may want to ping that team and see if they can address this.

lemmy04:
they are present in **old** versions of the base image... devel has no vulnerabilities, latest has one. Maybe the awx operator should be based on one of those?

PR #2132, opened 2026-08-13
feat: runtime NetworkPolicies for AWX operand workloads

kaiokmo:
- [x] New or Enhanced Feature

## Summary

Implements runtime NetworkPolicy creation for all AWX operator-managed workloads per OCPSTRAT-819 (OCP 5.0, Q4 2026).

- **default-deny**: baseline deny-all for pods with `managed-by` label (layered product pattern)
- **web**: ingress from OpenShift router + AAP components on port 8052; unrestricted egress for LDAP/SAML/webhooks
- **task**: egress-only; unrestricted for receptor mesh, SCM polling, execution environments
- **postgres**: ingress from AWX pods on port 5432; DNS-only egress (only when `managed_database`)
- **db-management**: egress to postgres + DNS for backup/restore/migration pods
- **mesh-ingress**: ingress from router on port 27199; unrestricted egress for receptor mesh peers

Also adds `managed-by` and `component` labels to the mesh-ingress deployment template so the default-deny catches receptor pods, following the same approach as [PR #191](https://github.com/ansible/awx-resource-operator/pull/191) which added labels to job templates.

## References

- Jira: [AAP-73974](https://redhat.atlassian.net/browse/AAP-73974)
- Parent initiative: [ANSTRAT-2322](https://redhat.atlassian.net/browse/ANSTRAT-2322)
- Reference implementation: gateway operator [MR !914](https://gitlab.cee.redhat.com/ansible/aap-gateway-operator/-/merge_requests/914)
- SDP: [ansible/handbook#1569](https://github.com/ansible/handbook/pull/1569)
- Peer PRs: [eda-server-operator#362](https://github.com/ansible/eda-server-operator/pull/362), [galaxy-operator#278](https://github.com/ansible/galaxy-operator/pull/278), [awx-resource-operator#191](https://github.com/ansible/awx-resource-operator/pull/191)

## Test plan

- [x] Molecule assertions: default-deny, web, task, postgres, db-management NPs verified
- [x] Build wrapper image from this PR
- [x] Deploy to OCP5 test cluster and verify NPs are created
- [ ] Run ATF regression suite via [sandbox pipeline](https://jenkins-csb-aap-main.dno.corp.redhat.com/job/AAPQA/job/Sandbox/job/ttuffin/job/ANSTRAT-2322-testing/)
- [ ] Validate pod-to-pod connectivity with NPs applied

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -70,6 +70,7 @@ rules:
       - networking.k8s.io
     resources:
       - ingresses
+      - networkpolicies
     verbs:
       - get
       - list
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -70,12 +70,23 @@ rules:
       - networking.k8s.io
     resources:
       - ingresses
-      - networkpolicies
     verbs:
       - get
       - list
       - create
       - delete
+      - patch
+      - update
+      - watch
+  - apiGroups:
+      - networking.k8s.io
+    resources:
+      - networkpolicies
+    verbs:
+      - create
+      - delete
+      - get
+      - list
       - patch
       - update
       - watch
--- a/roles/installer/templates/networking/db-management.networkpolicy.yaml.j2
+++ b/roles/installer/templates/networking/db-management.networkpolicy.yaml.j2
@@ -1,0 +1,47 @@
+# NetworkPolicy for backup, restore, and migration pods.
+# Matches all pods with the common AWX labels that are NOT specifically
+# selected by the web or task NPs (those pods already have their own policies
+# with unrestricted egress).
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: '{{ ansible_operator_meta.name }}-db-management'
+  namespace: '{{ ansible_operator_meta.namespace }}'
+  labels:
+    {{ lookup("template", "../common/templates/labels/common.yaml.j2") | indent(4) | trim }}
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
+      app.kubernetes.io/component: '{{ deployment_type }}'
+  policyTypes:
+    - Egress
+  egress:
+    # Allow DNS resolution via OpenShift DNS.
+    - to:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: openshift-dns
+      ports:
+        - protocol: TCP
+          port: 5353
+        - protocol: UDP
+          port: 5353
+{% if managed_database | bool %}
+    # Allow db-management pods to reach managed PostgreSQL.
+    - to:
+        - podSelector:
+            matchLabels:
+              app.kubernetes.io/component: database
+              app.kubernetes.io/instance: 'postgres-{{ supported_pg_version }}-{{ ansible_operator_meta.name }}'
+      ports:
+        - protocol: TCP
+          port: {{ awx_postgres_port | default('5432') }}
+{% else %}
+    # External database: allow on the configured port without destination
+    # restriction since the database host is user-configured.
+    - ports:
+        - protocol: TCP
+          port: {{ awx_postgres_port | default('5432') }}
+{% endif %}
--- a/roles/installer/templates/networking/db-management.networkpolicy.yaml.j2
+++ b/roles/installer/templates/networking/db-management.networkpolicy.yaml.j2
@@ -14,7 +14,7 @@ spec:
   podSelector:
     matchLabels:
       app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
-      app.kubernetes.io/component: '{{ deployment_type }}'
+      app.kubernetes.io/component: db-management
   policyTypes:
     - Egress
   egress:
--- a/roles/installer/templates/networking/db-management.networkpolicy.yaml.j2
+++ b/roles/installer/templates/networking/db-management.networkpolicy.yaml.j2
@@ -18,12 +18,14 @@ spec:
   policyTypes:
     - Egress
   egress:
-    # Allow DNS resolution via OpenShift DNS.
-    - to:
-        - namespaceSelector:
-            matchLabels:
-              kubernetes.io/metadata.name: openshift-dns
-      ports:
+    # Allow DNS resolution (port 53 for vanilla k8s / Kind, port 5353 for
+    # OpenShift DNS).  No destination restriction — DNS is a low-risk
+    # egress target.
+    - ports:
+        - protocol: TCP
+          port: 53
+        - protocol: UDP
+          port: 53
         - protocol: TCP
           port: 5353
         - protocol: UDP
--- a/roles/installer/templates/networking/default-deny.networkpolicy.yaml.j2
+++ b/roles/installer/templates/networking/default-deny.networkpolicy.yaml.j2
@@ -1,0 +1,24 @@
+# Default deny policy for AWX operand pods (layered product pattern).
+# Selects only pods managed by this operator instance, not all pods in the
+# namespace, since AWX / Controller is a layered product that may share a
+# namespace with other workloads.
+#
+# See: https://docs.google.com/document/d/1CDoGSRd-h8VT4PMrK_83Ro0YzYPjORbkxtfTJU1sN6Q
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: '{{ ansible_operator_meta.name }}-default-deny'
+  namespace: '{{ ansible_operator_meta.namespace }}'
+  labels:
+    {{ lookup("template", "../common/templates/labels/common.yaml.j2") | indent(4) | trim }}
+spec:
+  podSelector:
+    matchExpressions:
+      - key: app.kubernetes.io/managed-by
+        operator: In
+        values:
+          - '{{ deployment_type }}-operator'
+  policyTypes:
+    - Ingress
+    - Egress
--- a/roles/installer/templates/networking/default-deny.networkpolicy.yaml.j2
+++ b/roles/installer/templates/networking/default-deny.networkpolicy.yaml.j2
@@ -22,5 +22,3 @@ spec:
   policyTypes:
     - Ingress
     - Egress
-  ingress: []
-  egress: []
--- a/roles/installer/templates/networking/default-deny.networkpolicy.yaml.j2
+++ b/roles/installer/templates/networking/default-deny.networkpolicy.yaml.j2
@@ -22,3 +22,5 @@ spec:
   policyTypes:
     - Ingress
     - Egress
+  ingress: []
+  egress: []
--- a/roles/installer/templates/networking/postgres.networkpolicy.yaml.j2
+++ b/roles/installer/templates/networking/postgres.networkpolicy.yaml.j2
@@ -20,12 +20,13 @@ spec:
     - Ingress
     - Egress
   ingress:
-    # Allow connections from AWX operand pods (web, task, migration jobs,
-    # backup/restore) and the operator controller-manager.
+    # Allow connections from this AWX instance's operand pods (web, task,
+    # migration jobs, backup/restore) and the operator controller-manager.
     - from:
         - podSelector:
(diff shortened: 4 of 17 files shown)


TheRealHaoLiu:
### Code review

Found 2 issues:

1. The postgres NetworkPolicy's egress only allows DNS to `openshift-dns` (no port-5432 rule and no catch-all), but the PostgreSQL major-version upgrade path in [`roles/installer/tasks/upgrade_postgres.yml`](https://github.com/ansible/awx-operator/blob/70bca1beb0197f8402e3de06f33577a4df654e55/roles/installer/tasks/upgrade_postgres.yml#L98-L105) execs into the new postgres pod and runs `pg_dump -h <old-postgres-service> ... -p 5432`. The new pod matches this policy's `podSelector`, so that outbound connection is blocked and the documented dump/restore upgrade breaks.

https://github.com/ansible/awx-operator/blob/70bca1beb0197f8402e3de06f33577a4df654e55/roles/installer/templates/networking/postgres.networkpolicy.yaml.j2#L34-L45

2. The postgres NetworkPolicy ingress uses bare `podSelector`s (no `namespaceSelector`), which only match pods in the AWX CR's namespace. Backup/restore management pods run in `backup_pvc_namespace` ([`roles/backup/defaults/main.yml`](https://github.com/ansible/awx-operator/blob/70bca1beb0197f8402e3de06f33577a4df654e55/roles/backup/defaults/main.yml#L8-L10), documented as overridable to another namespace). With that override set, the backup/restore pod can no longer reach postgres — a regression, since this worked before NetworkPolicies existed.

https://github.com/ansible/awx-operator/blob/70bca1beb0197f8402e3de06f33577a4df654e55/roles/installer/templates/networking/postgres.networkpolicy.yaml.j2#L24-L34

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

kaiokmo:
@TheRealHaoLiu resolved. please check again

config/rbac/role.yaml
```diff
@@ -70,6 +70,7 @@ rules:
       - networking.k8s.io
```
rooftopcellist, line 70: **ANSTRAT-2322 [CRITICAL]: networking.k8s.io/networkpolicies has insufficient verbs**

The rule grants only `get` and `list`. Managing NetworkPolicies at runtime requires all of: `create, delete, get, list, patch, update, watch`. Without `create`, `patch`, and `delete`, every `kubernetes.core.k8s` reconcile call with `state: present apply: true` will return Forbidden.

This must remain a standalone rule — do not merge with the `ingresses` rule:
```yaml
- apiGroups:
    - networking.k8s.io
  resources:
    - networkpolicies
  verbs:
    - create
    - delete
    - get
    - list
    - patch
    - update
    - watch
```

roles/installer/templates/networking/default-deny.networkpolicy.yaml.j2
```diff
@@ -0,0 +1,24 @@
...
+  labels:
+    {{ lookup("template", "../common/templates/labels/common.yaml.j2") | indent(4) | trim }}
+spec:
+  podSelector:
+    matchExpressions:
+      - key: app.kubernetes.io/managed-by
+        operator: In
+        values:
+          - '{{ deployment_type }}-operator'
+  policyTypes:
```
rooftopcellist, line 22: **ANSTRAT-2322 [MAJOR]: Default-deny missing explicit `ingress: []` and `egress: []`**

Setting `policyTypes: [Ingress, Egress]` without the corresponding `ingress:` and `egress:` fields is implicit, not the required pattern. Add explicit empty arrays after the `policyTypes` block:
```yaml
  ingress: []
  egress: []
```

roles/installer/templates/networking/db-management.networkpolicy.yaml.j2
```diff
@@ -0,0 +1,47 @@
...
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: '{{ ansible_operator_meta.name }}-db-management'
+  namespace: '{{ ansible_operator_meta.namespace }}'
+  labels:
+    {{ lookup("template", "../common/templates/labels/common.yaml.j2") | indent(4) | trim }}
+spec:
+  podSelector:
```
rooftopcellist, line 14: **ANSTRAT-2322 [MAJOR]: db-management podSelector matches web and task pods, not just db-management pods**

The selector also matches `web` and `task` Deployment pods. In external DB mode, this grants those pods an unintended egress path to postgres.

Add a distinct label (e.g. `app.kubernetes.io/component: db-management`) to backup/restore/migration pods and narrow this selector to match only that label.

roles/mesh_ingress/tasks/creation.yml
```diff
@@ -79,6 +79,9 @@
     external_hostname: "{{ ingress.result.status.ingress[0].host }}"
   when: ingress_type | lower == 'route'
 
+- name: Apply NetworkPolicy for mesh-ingress
```
rooftopcellist: **ANSTRAT-2322 [MAJOR]: mesh-ingress NetworkPolicy task is ordered after Deployment creation**

The NetworkPolicy include_tasks runs after Deployment creation. During reconciliation, mesh-ingress pods start up under the default-deny until the allow policy is applied.

Move the NetworkPolicy include_tasks call to the top of creation.yml, before any Deployment tasks.

rooftopcellist (changes requested):
ANSTRAT-2322 NetworkPolicy compliance review: four findings — one critical (RBAC), three major. See inline comments.

kaiokmo:
I had to apply the same fix of removing the explicit ingress/egress empty arrays from default-deny NetworkPolicy, as observed in https://github.com/ansible/galaxy-operator/pull/278, to avoid an infinite reconciliation loop.

rooftopcellist (approved):
NetworkPolicy implementation meets all ANSTRAT-2322 requirements. The reconcile-loop concern with empty ingress/egress arrays is fully resolved — the molecule test correctly validates the Kubernetes normalization behavior. The namespace-selector approach for OCP router ingress is consistent with the fleet-wide pattern and the OCP-recommended implementation. Approved.

roles/installer/templates/networking/postgres.networkpolicy.yaml.j2
```diff
@@ -0,0 +1,71 @@
...
+  policyTypes:
+    - Ingress
+    - Egress
+  ingress:
+    # Allow connections from AWX operand pods (web, task, migration jobs,
+    # backup/restore) and the operator controller-manager.
+    - from:
+        - podSelector:
+            matchLabels:
+              app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
```
lucas-benedito, line 28: The FROM selector allows any AWX-managed pod to reach postgres, regardless of which AWX instance it belongs to. In a namespace with two AWX instances, the postgres NP for `example-awx-1` permits ingress from `example-awx-2`'s web and task pods.

Add instance scoping to both FROM entries here:

```yaml
        - podSelector:
            matchLabels:
              app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
              app.kubernetes.io/part-of: '{{ ansible_operator_meta.name }}'
```

Apply the same fix to the `control-plane: controller-manager` podSelector below (line 31).

roles/installer/templates/networking/web.networkpolicy.yaml.j2
```diff
@@ -0,0 +1,86 @@
...
+            matchLabels:
+              app.kubernetes.io/component: database
+              app.kubernetes.io/instance: 'postgres-{{ supported_pg_version }}-{{ ansible_operator_meta.name }}'
+      ports:
+        - protocol: TCP
+          port: {{ awx_postgres_port | default('5432') }}
+{% endif %}
+    # Unrestricted egress for LDAP/SAML/webhooks and other user-configured
+    # external services whose addresses cannot be enumerated.
+    - {}
```
lucas-benedito, line 42: This `- {}` rule supersedes the scoped postgres egress block above it. Kubernetes evaluates NP egress rules as a union — if any rule matches, the packet is allowed. `- {}` permits egress to every destination and port, so the `{% if managed_database | bool %}` postgres block above it (lines 73–80) has no runtime effect.

Remove the conditional postgres egress block. The catch-all already covers it, and keeping the scoped block implies it provides a restriction it doesn't actually enforce.

roles/installer/templates/networking/task.networkpolicy.yaml.j2
```diff
@@ -0,0 +1,42 @@
...
+            matchLabels:
+              app.kubernetes.io/component: database
+              app.kubernetes.io/instance: 'postgres-{{ supported_pg_version }}-{{ ansible_operator_meta.name }}'
+      ports:
+        - protocol: TCP
+          port: {{ awx_postgres_port | default('5432') }}
+{% endif %}
+    # Unrestricted egress for receptor mesh, SCM polling, execution
+    # environments, Kubernetes API, and other external services.
+    - {}
```
lucas-benedito, line 32: Same issue as `web.networkpolicy.yaml.j2`: this `- {}` supersedes the `{% if managed_database | bool %}` postgres egress block above it (lines 29–36). Remove the conditional block — the catch-all already permits this traffic.

rooftopcellist (approved):
All open findings addressed. Postgres NP now has instance scoping via part-of label on the operand pod selector. Dead conditional managed_database egress blocks removed from web and task NPs — catch-all egress with descriptive comments is the correct pattern. Approved.

roles/installer/templates/networking/default-deny.networkpolicy.yaml.j2
kaiokmo: Intentionally omitted — adding explicit `ingress: []` and `egress: []` causes an infinite reconciliation loop. The Kubernetes API server normalizes these empty arrays by stripping them from the stored object, so `apply: true` detects a perpetual diff and re-triggers reconciliation on every cycle.

`policyTypes: [Ingress, Egress]` without `ingress`/`egress` fields already enforces default-deny semantics. Same pattern as [galaxy-operator#278](https://github.com/ansible/galaxy-operator/pull/278). The molecule test uses `| default([])` guards to handle the absent fields.

config/rbac/role.yaml
kaiokmo: Already addressed — `config/rbac/role.yaml` has all 7 verbs (`create, delete, get, list, patch, update, watch`) for `networkpolicies`. This was present in the initial implementation.

roles/installer/templates/networking/db-management.networkpolicy.yaml.j2
kaiokmo: Fixed — the `db-management` NP now uses `app.kubernetes.io/component: db-management` in its `podSelector`, and the backup/restore management pod templates have been updated to carry that label.

roles/mesh_ingress/tasks/creation.yml
kaiokmo: Fixed — `include_tasks: networkpolicy.yml` is now called before the Deployment creation in `roles/mesh_ingress/tasks/creation.yml`.

roles/installer/templates/networking/postgres.networkpolicy.yaml.j2
kaiokmo: Fixed — added `app.kubernetes.io/part-of: '{{ ansible_operator_meta.name }}'` to both `from` podSelector entries, scoping ingress to pods belonging to this specific AWX instance.

roles/installer/templates/networking/web.networkpolicy.yaml.j2
kaiokmo: Fixed — removed the redundant conditional postgres egress block. The catch-all `- {}` already permits all egress.

roles/installer/templates/networking/task.networkpolicy.yaml.j2
kaiokmo: Fixed — same as web, removed the redundant conditional block.

rooftopcellist:
The read the docs and nox-sessions changes are unrelated to the changes here.

PR #2135, opened 2026-08-25
fix: preserve backup_dir on finalizer path for clean_backup_on_delete

jamesmarshall24:
##### SUMMARY
`clean_backup_on_delete: true` stopped removing the backup directory
after CR deletion. The postgres_skip_data refactor (#2120) moved
backup_dir/timestamp setup into init.yml, which finalizer.yml also
runs on CR deletion. set_fact overwrote the block-scoped backup_dir
passed in from CR status, so delete_backup.yml deleted a freshly
created empty directory instead of the real backup.

Guards the three affected tasks with `when: not finalizer_run | bool`,
matching the existing finalizer_run branching in main.yml.

##### ISSUE TYPE

##### ADDITIONAL INFORMATION
Repro: create an AWXBackup with clean_backup_on_delete: true, wait for
backup success, delete the AWXBackup. Before: backup directory remains
on the PVC. After: directory is removed.

### DIFF
--- a/roles/backup/tasks/init.yml
+++ b/roles/backup/tasks/init.yml
@@ -97,10 +97,12 @@
 - name: Determine the timestamp for the backup once for all nodes
   set_fact:
     now: '{{ lookup("pipe", "date +%F-%H%M%S") }}'
+  when: not finalizer_run | bool
 
 - name: Set backup directory name
   set_fact:
     backup_dir: "/backups/tower-openshift-backup-{{ now }}"
+  when: not finalizer_run | bool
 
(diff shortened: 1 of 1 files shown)


rooftopcellist (approved):
Traced the regression end to end — the postgres_skip_data refactor in #2120 moved backup_dir/timestamp setup into init.yml, which the finalizer path also runs on CR deletion, and the unconditional set_fact was clobbering the block-scoped backup_dir passed in from CR status. The when: not finalizer_run guard is the right fix and matches the existing pattern in main.yml. LGTM.

PR #2141, opened 2026-09-09
PoC: trigger reconcile on PG secret rotation via label-selector watch

rooftopcellist:
## Summary

Proof of concept for discussion — not a finished patch. Rotating the Controller/AWX PostgreSQL configuration secret externally (e.g. via Vault Secrets Operator) never triggers a reconcile today, so `credentials.py` can go stale indefinitely. Root cause: that secret is deliberately never given an `ownerReference` (`cleanup.yml` actively strips one if present, so it survives CR deletion), and ansible-operator's dependent-resource watch requires one to fire.

This PR adds an independent, label-selector-based watch on the secret instead of relying on ownership:

- `watches.yaml`: new `kind: Secret` watch entry with a `matchExpressions: Exists` selector, plus `watchAnnotationsChanges: true` on the primary `AWX` entry.
- `playbooks/pg_secret_rotated.yml` (new): reads the owning `AWX`'s name directly off the secret's label and patches an annotation on it to force a real reconcile.
- `roles/installer/tasks/database_configuration.yml`: unconditionally (labels-only merge, never touches secret data) applies the watch label to whatever secret is actually resolved, so this also covers an externally supplied `postgres_configuration_secret` (BYO database, or a secret populated by VSO before the operator ever runs) — not just the case where the operator generates the secret itself.
- `roles/installer/templates/secrets/postgres_secret.yaml.j2`: the label's value is the owning CR's own name rather than a fixed constant, so one static watch entry correctly covers multiple `AWX` instances in the same namespace.
- `roles/restore/templates/secrets.yml.j2`: explicitly nulls the label when restoring the PG secret. `restore/tasks/main.yml` applies the restored secret before `deploy_awx.yml`/`postgres.yml` run, and `install.yml`'s "wait for restore to complete" gate sits after `database.yml`'s postgres upgrade/migration tasks rather than before them — so on an in-place restore, letting the label survive would let restore's own write retrigger this watch and force a concurrent reconcile mid-restore. The normal post-restore reconcile reasserts the label once `restoreComplete` is set.

The secret still carries no `ownerReference` throughout and never enters Kubernetes' garbage-collection graph.

## Validation

Before writing this patch, validated the underlying ansible-operator mechanism empirically against a live OpenShift cluster with the pinned `ansible-operator` binary:
- A `selector`-filtered watch on a bare `kind: Secret` (no CRD) registers and runs cleanly.
- It correctly matches on label and fires on real data changes (confirmed via `resourceVersion` correlation), not just on a polling interval.
- The secret's `ownerReferences` stayed empty throughout.

## Open questions for reviewers

- Is `matchExpressions`/`selector` on a bare non-CRD `kind: Secret` watch entry an intentional, supported usage of ansible-operator, or is this leaning on undocumented behavior we shouldn't rely on?
- `install.yml`'s "wait for restore to complete" gate sits after `database.yml`'s DB-mutating tasks rather than before them — a pre-existing gap independent of this PR, but worth a second set of eyes given this PR's restore-window fix is built around working around it rather than fixing it directly.
- No tests included yet — wanted early feedback on the approach before investing in molecule coverage.

## Test plan

- [ ] Add molecule/integration coverage for the new watch + playbook
- [ ] Verify against a real Vault Secrets Operator-managed secret, not just a manually rotated one
- [ ] Confirm behavior with multiple `AWX` instances in one namespace

### DIFF
--- a/playbooks/pg_secret_rotated.yml
+++ b/playbooks/pg_secret_rotated.yml
@@ -1,0 +1,57 @@
+---
+# Triggered by the "kind: Secret" watch entry in watches.yaml whenever a
+# secret carrying the awx.ansible.com/owner-name label changes.
+# ansible_operator_meta here refers to the Secret itself (the primary
+# resource of this watch entry), not an AWX CR. The label's value names the
+# owning CR directly (stamped in postgres_secret.yaml.j2 and self-healed
+# every reconcile in database_configuration.yml, regardless of whether the
+# secret is operator-managed or externally supplied), so there's no need to
+# enumerate every AWX in the namespace or guess by naming convention - just
+# read it and nudge that CR into a real reconcile. The actual
+# credentials.py re-render already happens correctly in the installer
+# role, it just never got triggered.
+- hosts: localhost
+  gather_facts: false
+  collections:
+    - kubernetes.core
+  tasks:
+    - name: Read the triggering PostgreSQL configuration secret
+      k8s_info:
+        kind: Secret
+        name: "{{ ansible_operator_meta.name }}"
+        namespace: "{{ ansible_operator_meta.namespace }}"
+      register: _triggering_secret
+
+    - name: Determine the owning AWX from the secret's label
+      set_fact:
+        _owner_name: >-
+          {{ _triggering_secret.resources[0].metadata.labels['awx.ansible.com/owner-name']
+             | default('') }}
+
+    - name: Confirm the owning AWX still exists
+      k8s_info:
+        api_version: awx.ansible.com/v1beta1
+        kind: AWX
+        name: "{{ _owner_name }}"
+        namespace: "{{ ansible_operator_meta.namespace }}"
+      register: _owner
+      when: _owner_name | length > 0
+
+    - name: Force a reconcile of the owning AWX
+      k8s:
+        kind: AWX
+        api_version: awx.ansible.com/v1beta1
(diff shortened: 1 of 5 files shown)


playbooks/pg_secret_rotated.yml
```diff
@@ -0,0 +1,57 @@
...
+        namespace: "{{ ansible_operator_meta.namespace }}"
+        definition:
+          metadata:
+            annotations:
+              # Value only needs to change to trigger a reconcile (see
+              # watchAnnotationsChanges: true on the AWX watch entry) - the
+              # secret's own resourceVersion is a convenient,
+              # always-changing value to use.
+              awx.ansible.com/pg-secret-resource-version: >-
+                {{ _triggering_secret.resources[0].metadata.resourceVersion }}
```
jamesmarshall24, line 54: Would a change to resourceVersion trigger an update here? I think this runs into similar issues we've seen previously using resourceVersion

rooftopcellist:
I think that this change is too brittle and more complex than it is worth for the small utility.  If you are rotating the SECRET_KEY or postgres_password, after updating the value on the k8s secret for `secret_key_secret` and/or `postgres_configuration_secret`, you will need to do a rollout restart on the awx-task and awx-web containers for the change to take effect.  

Let's move forward with https://github.com/ansible/awx-operator/pull/2142 instead.

